PR #3, opened 2024-12-02
ci: setup basic build and test

kevineger:
Influenced from the following:

- https://code.visualstudio.com/api/working-with-extensions/continuous-integration#github-actions
- https://github.com/microsoft/vscode-jupyter/blob/main/.github/workflows/build-test.yml
- https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-nodejs

Further improvements to follow once we have the basic CI working, namely: packaging workflow data as artifacts, release check (once we have a release), caching npm deps, lint checking, format checking, checking for package-lock.json changes, etc. The VS Code [Jupyter extension](https://github.com/microsoft/vscode-jupyter/blob/main/.github/workflows/build-test.yml) is a pretty solid reference for the things we'll want to add.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,0 +1,41 @@
+# For PRs.
+
+name: Build and Test
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  vsCodeTests:
+    name: VS Code Integration Tests
+
+    strategy:
+      matrix:
+        # macos-latest and windows-latest can be added as we work towards a release.
+        # For now, there's no sense acruing the cost of the runners.
+        os: [ubuntu-latest]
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test.yml
```diff
@@ -0,0 +1,41 @@
+# For PRs.
+
+name: Build and Test
+
+on:
+  push:
+    branches:
+    - main
```
katlyn-edwards: not sure: below on pull, main is indented. Should this also be indented? Or maybe the lower one out-dented?

kevineger: Sorry, no YAML linting yet.

Installed and ran the [defacto VS Code YAML formatter](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml).

PR #6, opened 2024-12-04
feat: naive Jupyter provider

kevineger:
For now, the provider naively provides and resolves a static list of Colab servers. Additionally, the RP connection information is provided as a hardcoded configuration until we have assignment logic implemented. Both of these items are flagged with TODOs.

Other minor changes:

- Deleted the temporary `tmp.unit.test.ts` since we now have a _real_ unit test. As per https://github.com/googlecolab/colab-vscode/pull/4.
- Updated the `test:unittests` script to quote the glob pattern that's used. This was a bit of a fun one to figure out. In short, when left as a naked glob, the shell will expand it. on Linux `sh` will not enable globstar by default. This was causing the test files to not both be resolved (prior to deleting the `tmp` one). This is most effectively solved by quoting the glob which in turn relies on the glob tool used by the tool (`mocha`) to consistently expand the star. [This blog post](https://medium.com/@jakubsynowiec/you-should-always-quote-your-globs-in-npm-scripts-621887a2a784) covers this in detail.
- Bumped the VS Code version.
- Added a dependency on `chai` for assertions and `sinon` for spies, stubs and mocks. Across the extension ecosystem and TS/JS in general, these are the (current) most popular choices.

### DIFF
--- a/src/jupyter/provider.ts
+++ b/src/jupyter/provider.ts
@@ -1,0 +1,97 @@
+import {
+  Jupyter,
+  JupyterServer,
+  JupyterServerCollection,
+  JupyterServerConnectionInformation,
+  JupyterServerProvider,
+} from "@vscode/jupyter-extension";
+import { CancellationToken, Uri, ProviderResult } from "vscode";
+
+/**
+ * Registers the Colab Jupyter Server provider with the Jupyter Kernels API.
+ *
+ * @param jupyter Kernels API.
+ * @param config for connecting to the Resource Proxy.
+ */
+export function register(
+  jupyter: Jupyter,
+  config: RpConfig
+): JupyterServerCollection {
+  return jupyter.createJupyterServerCollection(
+    "colab",
+    "Colab",
+    new ColabJupyterServerProvider(config)
+  );
+}
+
+/**
+ * Configuration for the Resource Proxy connection.
+ */
+export interface RpConfig {
+  /**
+   * Base {@link Uri Uri} of the Jupyter Server behind the resource proxy.
+   */
+  readonly baseUri: Uri;
+
+  /**
+   * The resource proxy token attached as a header to Jupyter Server requests.
+   */
+  readonly token: string;
+}
+
+/**
+ * Colab Jupyter server provider.
+ *
+ * Provides a static list of Colab Jupyter servers and resolves the connection information using the provided config.
+ */
+export class ColabJupyterServerProvider implements JupyterServerProvider {
+  // TODO: Fetch available servers from the backend. Hardcoded for now.
+  private readonly idToServer = new Map<string, ColabJupyterServer>([
+    ["m", new ColabJupyterServer("m", "Colab CPU")],
+    ["gpu-t4", new ColabJupyterServer("gpu-t4", "Colab T4")],
+    ["gpu-l4", new ColabJupyterServer("gpu-l4", "Colab L4")],
+    ["gpu-a100", new ColabJupyterServer("gpu-a100", "Colab A100")],
+    ["tpu-v28", new ColabJupyterServer("tpu-v28", "Colab TPU v2-8")],
+    ["tpu-v5e1", new ColabJupyterServer("tpu-v5e1", "Colab TPU v5e-1")],
+  ]);
+
+  constructor(private readonly config: RpConfig) {}
+
+  /**
+   * Provides the list of {@link JupyterServer Jupyter Servers}.
+   */
+  provideJupyterServers(
+    _token: CancellationToken
+  ): ProviderResult<JupyterServer[]> {
+    return Array.from(this.idToServer.values());
+  }
+
+  /**
+   * Resolves the connection for the provided {@link JupyterServer Jupyter Server}.
+   */
+  resolveJupyterServer(
+    server: JupyterServer,
+    _token: CancellationToken
+  ): ProviderResult<JupyterServer> {
+    const colabServer = this.idToServer.get(server.id);
+    if (!colabServer) {
+      return;
+    }
+    colabServer.resolve(this.config);
+    return colabServer;
+  }
+}
+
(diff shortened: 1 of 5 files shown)


src/jupyter/provider.ts
```diff
@@ -0,0 +1,97 @@
...
+    new ColabJupyterServerProvider(config)
+  );
+}
+
+/**
+ * Configuration for the Resource Proxy connection.
+ */
+export interface RpConfig {
+  /**
+   * Base Uri of the Jupyter Server behind the resource proxy.
```
zelliott: nit: If referring to the type, then `Uri`, if the proper noun, then [URI](https://developer.mozilla.org/en-US/docs/Glossary/URI).

src/jupyter/provider.ts
```diff
@@ -0,0 +1,97 @@
...
+  readonly token: string;
+}
+
+/**
+ * Colab Jupyter server provider.
+ *
+ * Provides a static list of Colab Jupyter servers and resolves the connection information using the provided config.
+ */
+export class ColabJupyterServerProvider implements JupyterServerProvider {
+  // TODO: Fetch available servers from the backend. Static for now.
```
zelliott: nit: "Static" confused me because I expected to see a `static` modifier here. Should there be one? Otherwise, consider changing to "Readonly", or something similar.

src/jupyter/provider.ts
```diff
@@ -0,0 +1,97 @@
...
+  private readonly idToServer = new Map<string, ColabJupyterServer>([
+    ["m", new ColabJupyterServer("m", "Colab CPU")],
```
zelliott, line 50: What i18n requirements do we have for these end user messages, if any? These messages in particular aren't the best example because they're _potentially_ the same across all languages, but I'm still curious.

src/jupyter/provider.ts
```diff
@@ -0,0 +1,97 @@
...
+    ["tpu-v5e1", new ColabJupyterServer("tpu-v5e1", "Colab TPU v5e-1")],
+  ]);
+
+  constructor(private readonly config: RpConfig) {}
+
+  /**
+   * Provides the list of {@link JupyterServer Jupyter Servers}.
+   */
+  provideJupyterServers(
+    _token: CancellationToken
```
zelliott, line 64: Are these `_` prefixes silencing linter unused arg warnings? You need to include the args to peroperly implement the interface, and they may be used in later CLs?

src/jupyter/provider.unit.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  JupyterServerProvider,
+  JupyterServerCollection,
+} from "@vscode/jupyter-extension";
+import { assert, expect } from "chai";
+import sinon from "sinon";
+import { CancellationToken, Uri } from "vscode";
+import { ColabJupyterServerProvider, register, RpConfig } from "./provider";
+
+describe("register", () => {
+  it("should create a Jupyter server collection", () => {
```
zelliott: nit: While we're establishing conventions and early in writing unit tests, WDYT about omitting "_should_" from all of these test names and pluralizing the verbs?

_"should create a Jupyter server collection"_ -> _"creates a Jupyter server collection"_

_"should"_ in my experience just makes the test name longer and doesn't provide any value, but I am open to being convinced otherwise.

src/jupyter/provider.ts
kevineger: image

I'll the TS Doc variant (`{@link Uri Uri}`) which has some nice features in most editors (see screenshot).

src/jupyter/provider.ts
kevineger: ~~Static~~ Hardcoded.

src/jupyter/provider.ts
kevineger: Initially we aren't targeting i18n support, but I'll send you a link to our internal design doc which talks about how we plan to add it.

src/jupyter/provider.ts
kevineger: Future _PRs_ 😉, potentially. But yes, it's to satisfy the `@typescript-eslint/no-unused-vars` rule I configured in [eslint.config.mjs](https://github.com/googlecolab/colab-vscode/blob/77e1edb57926f0ee283c0233013c417982880f65/eslint.config.mjs#L28). Here I must implement the interface.

src/jupyter/provider.unit.test.ts
kevineger: No strong feels either way, happy to remove and snap to what we do internally. Done.

kevineger:
> > Added a dependency on `chai` for assertions and `sinon` for spies, stubs and mocks. Across the extension ecosystem and TS/JS in general, these are the (current) most popular choices.
> 
> How much more popular than jasmine? I've never used chai or sinon and I'm happy to learn (it looks basically the same as jasmine), but there is some value in trying to adhere to g3 tools/frameworks where palatable to make it easy to contribute to this repo and context-switch between g3 and this repo. Will leave up to you.

VS Code integration tests lean on Mocha, so I want to use Mocha in the unit tests as well for consistency. According to `npm` installs, [Mocha](https://www.npmjs.com/search?q=mocha) is much more popular than [Jasmine](https://www.npmjs.com/search?q=jasmine). For what it's worth, the APIs are somewhat similar. [Chai](https://www.npmjs.com/package/chai) is the defacto assertion library for `node` apps and [Sinon](https://www.npmjs.com/search?q=sinon) as a standalone for framework-agnostic test spies, stubs and mocks is more popular than Jamine itself.

Another data point is the `vscode-jupyter` codebase, the dependent extension, which elected the [same test libraries](https://github.com/microsoft/vscode-jupyter/blob/main/package.json).

I'll start with these libraries and am happy to re-assess and migrate if we'd like in the future.

PR #7, opened 2024-12-07
feat: register Jupyter provider on activation

kevineger:
[RP hardcoded demo.webm](https://github.com/user-attachments/assets/abf3ab44-2320-4969-b154-57e844d4eea8)

For now, the URL and token are hardcoded as settings until we have auth and the assignment flow implemented. This is temporary.

Added `ms-toolsai.jupyter` to the extension dependencies. Added the necessary install step in the test runner.

Run tests with the `"insiders"` build of VS Code so it can be run alongside VS Code while developing ([VS Code tip](https://code.visualstudio.com/api/working-with-extensions/testing-extension#using-insiders-version-for-extension-development)). This is consistent with what I see other extensions doing ([example](https://github.com/microsoft/vscode-jupyter-hub/blob/main/src/test/runTest.ts#L18)).

Unfortunately there's a few more formatting changes than I'd like. Reminds me I need to setup Prettier and enforce that with automation sooner than later. I'll try to get to that shortly.

Other minor changes:

- Moved `register` to be a static function on `ColabJupyterServerProvider`.
- Removes the specification of `outFiles` for the _Extension Integration Tests_ launch configuration so VS Code can map breakpoints to the non-test `src/` files.
- Specifies `commonjs` in `package.json` module compatibility. Seems to be what most extensions set.
- Switched from Mocha's TDD -> BDD since that's what the team's more familiar with internally (with Jasmine).

### DIFF
--- a/src/test/extension.vscode.test.ts
+++ b/src/test/extension.vscode.test.ts
@@ -1,25 +1,24 @@
-import * as assert from 'assert';
-import * as vscode from 'vscode';
+import * as assert from "assert";
+import * as vscode from "vscode";
 
-suite('Extension', () => {
-	test('should be present', () => {
-		assert.ok(vscode.extensions.getExtension('google.colab'));
-	});
+describe("Extension", () => {
+  it("should be present", () => {
+    assert.ok(vscode.extensions.getExtension("google.colab"));
+  });
 
-	test('should activate', async () => {
-		const extension = vscode.extensions.getExtension('google.colab');
+  it("should activate", async () => {
+    await setConfig("resoruceProxyBaseUrl", "foo");
+    await setConfig("resoruceProxyToken", "bar");
+    const extension = vscode.extensions.getExtension("google.colab");
 
-		await extension?.activate();
+    await extension?.activate();
 
-		assert.strictEqual(extension?.isActive, true);
-	});
-
-	test('should register the helloWorld command', async () => {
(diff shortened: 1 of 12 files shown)


src/test/extension.vscode.test.ts
```diff
@@ -1,25 +1,24 @@
...
-	});
+describe("Extension", () => {
+  it("should be present", () => {
+    assert.ok(vscode.extensions.getExtension("google.colab"));
+  });
 
-	test('should activate', async () => {
-		const extension = vscode.extensions.getExtension('google.colab');
+  it("should activate", async () => {
+    await setConfig("resoruceProxyBaseUrl", "foo");
```
katlyn-edwards, line 10: idk: worth testing if these aren't provided that the extension throws?

kevineger: Spent a bit longer trying to make this work than I care to admit... but as far as I can tell there isn't a trivial way to test multiple activation paths. At least not given my current structure.

The VS Code `extensions` API doesn't provide any way to deactivate an extension, so the activation from the happy-path test means the second `activate()` is a no-op.

[Feature request](https://github.com/microsoft/vscode/issues/96090) where others would like the same.

As far as I can tell, to test this I'd have to have a standalone suite (in other words - opens a new VS Code window). Seems like overkill for a transient config path which will go away. Omitting for now.

PR #11, opened 2025-01-23
test: establish unit test pattern with VS Code

kevineger:
Unit testing a VS Code extension is super tricky! I spent a bit longer on this than I care to admit, but I think I've found a pattern which makes sense. I'll do my best to explain that here and link out to the resources which helped influence the approach.

The crux of the problem is `vscode` is not a normal node dependency. Rather, it's the _engine_ that runs the application. This means that we can't just import `vscode` into our tests and expect it to work. It's only available when VS Code is running, or the integration test runner is driving the tests. I like unit tests, a lot, so I wanted to find a way to test the extension without having to run the entire application.

With a few sub-option branches, there are basically two options for achieving this. You can either (1) wrap the `vscode` API or (2) mock it. If you're really curious what these (and the various sub-options) look like, you can checkout the resources I found and used to shape my approach. I tried all of them (including swapping to Jest) 😫.

- https://gitlab.com/gitlab-org/gitlab-vscode-extension/-/snippets/3710659
- https://devblogs.microsoft.com/ise/testing-vscode-extensions-with-typescript/#option-2:-use-jest-manual-mocks
- https://app.studyraid.com/en/read/8400/231878/mocking-vs-code-api
- https://www.richardkotze.com/coding/unit-test-mock-vs-code-extension-api-jest

By injecting the entire `vscode` module, it not only makes test setup more clear and consistent, but also encourages a single point of injection for the `vscode` environment (engine). While this PR only adds the stubs needed to swap the Jupyter extension integration test to a unit test, subsequent PRs add more tests and further build out the `vscodeStub`.


.vscode/launch.json
```diff
@@ -1,6 +1,14 @@
 {
   "version": "0.2.0",
   "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug Unit Tests",
```
zelliott: nit: Consider putting the `name` at the top of this configuration to align with the structures of the other two configurations, and make it easy to scan the list of configurations.

Also, what is this configuration used for? 😛 Does this make it possible to manually run unit tests via a launch option in the VSCode UI? Sorry, it's been a while since I've seen all of this.

src/jupyter/jupyter-extension.ts
```diff
@@ -1,11 +1,11 @@
 import { Jupyter } from "@vscode/jupyter-extension";
-import { extensions } from "vscode";
+import vscode from "vscode";
 
 /**
  * Get the exported API from the Jupyter extension.
  */
-export async function getJupyterApi(): Promise<Jupyter> {
-  const ext = extensions.getExtension<Jupyter>("ms-toolsai.jupyter");
+export async function getJupyterApi(vs: typeof vscode): Promise<Jupyter> {
```
zelliott, line 7: nit: It seems like there is a [convention](https://source.corp.google.com/search?q=%22code:%20typeof%20vscode%22%20file:third_party&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental) to call this `code`. I don't feel super strongly one way or the other, and convention aside actually kind of like `vs`. Just thought I would flag this for you.

src/jupyter/jupyter-extension.unit.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    beforeEach(() => {
+      activateStub = sinon.stub();
+    });
+
+    function getJupyterExtension(
+      status: ExtensionStatus,
+    ): Partial<vscode.Extension<Jupyter>> {
+      return {
+        isActive: status === ExtensionStatus.Active ? true : false,
```
zelliott: `isActive: status === ExtensionStatus.Active,`

src/jupyter/jupyter-extension.unit.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        exports: {
+          kernels: {
+            getKernel: sinon.stub(),
+          },
+          createJupyterServerCollection: sinon.stub(),
+        },
+      };
+    }
+
+    it("should throw an error if the Jupyter extension is not installed", async () => {
```
zelliott: nit: https://github.com/googlecolab/colab-vscode/pull/6#discussion_r1870328920

src/jupyter/jupyter-extension.unit.test.ts
```diff
@@ -0,0 +1,66 @@
...
+enum ExtensionStatus {
+  Active,
+  Inactive,
+}
+
+describe("Jupyter Extension", () => {
+  describe("getJupyterApi", () => {
+    let activateStub: SinonStub<[], Thenable<Jupyter>>;
+
+    beforeEach(() => {
```
zelliott, line 18: Do we need to restore sinon's sandbox after each test (https://sinonjs.org/releases/v19/general-setup/)? Maybe we're already doing this somewhere, I couldn't find it. I see it was in the old test module, maybe it was dropped accidentally.

src/jupyter/jupyter-extension.unit.test.ts
```diff
@@ -0,0 +1,66 @@
...
+          createJupyterServerCollection: sinon.stub(),
+        },
+      };
+    }
+
+    it("should throw an error if the Jupyter extension is not installed", async () => {
+      getExtensionStub.returns(undefined);
+
+      await expect(
+        getJupyterApi(vscodeStub as typeof vscode),
```
zelliott: I don't know how I feel about all of this `as typeof vscode` casting. I get that it's generally good practice to cast as closely to where you actually need to, but every single test will need to do this, and it begins to harm test readability.  I don't think there's ever an instance where you _don't_ want to immediately cast it (is there?) so perhaps that's a sign that the cast should be upstreamed to the `helpers/vscode.ts` module.

src/test/unit_test_setup.ts
zelliott, line 1: `unit-test-setup.ts`?

src/test/helpers/vscode.ts
```diff
@@ -0,0 +1,17 @@
...
+const getExtensionStub: SinonStub<
+  [extensionId: string],
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  vscode.Extension<any> | undefined
+> = sinon.stub();
+
+const vscodeStub: Pick<typeof vscode, "extensions"> = {
+  extensions: {
+    getExtension: getExtensionStub,
+  } as Partial<typeof vscode.extensions> as typeof vscode.extensions,
```
zelliott: `as please as give as me as the as type as i as want`

If easy, remind me again why this chained type casting is needed? I'm new to TS. If not easy, no worries I will dig into it myself.

.vscode/launch.json
kevineger: Done.

> Also, what is this configuration used for? 😛 Does this make it possible to manually run unit tests via a launch option in the VSCode UI?

Yep! 
image

src/jupyter/jupyter-extension.ts
kevineger: That (internal to Google) convention is just what the _Cloud Code_ extension opted for. Going to stick with `vs` if it's all the same.

src/jupyter/jupyter-extension.unit.test.ts
kevineger: Derp.

src/jupyter/jupyter-extension.unit.test.ts
kevineger: Done.

src/jupyter/jupyter-extension.unit.test.ts
kevineger: _Need to_, nope - tests are passing. Should we? Yes, it's better practice. Done.

Why it's working without - `activateStub` is reassigned `beforeEach` and the `getExtensionStub` for the `vscodeStub` is only ever setup for its `return` (calls aren't asserted).

src/jupyter/jupyter-extension.unit.test.ts
kevineger: Fair enough. I'll move to the cast up.

src/test/unit_test_setup.ts
kevineger: Woops, done.

src/test/helpers/vscode.ts
kevineger: Consider the following interface:

```ts
interface T {
  a: string;
  b: string;
}
```

- `const t: T = { a: "a" };` - _Property 'b' is missing in type '{ a: string; }' but required in type 'T'_.
- `const t: T = { a: "a" } as Pick<T, "a">;` - _Property 'b' is missing in type 'Pick<T, "a">' but required in type 'T'_.
- `const t: T = { a: "a" } as Pick<T, "a"> as T;` - ✅

I like doing `Pick` over `Partial` as it allows me in tests to deal with exactly what's defined instead of `maybeExists?.foo`.

PR #12, opened 2025-01-23
feat: add auth redirect handling

kevineger:
This is the first change in a chain of changes for adding Google OAuth2 authentication to the extension.

To support creating sessions (OAuth2 tokens), the extension needs to support listening to URI redirects.

The implementation of this logic was influenced by the GitHub authentication extension. A good starting point when looking at their logic are the [flows](https://github.com/microsoft/vscode/blob/ab172453fd2fe51a2bcf556c64542ee45675b3e8/extensions/github-authentication/src/flows.ts#L77).

### DIFF
--- a/src/auth/redirect.ts
+++ b/src/auth/redirect.ts
@@ -50,7 +50,6 @@ export class RedirectUriCodeProvider
         waitForExchangeTimeout(),
       ]);
     } catch (err: unknown) {
-      this.inFlightPromises.delete(nonce);
       if (err instanceof Error) {
         reject(err);
       } else {


src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+  constructor(private readonly vs: typeof vscode) {
+    this.redirectUriEmitter = new vs.EventEmitter<vscode.Uri>();
+    this.disposable = this.redirectUriEmitter;
+  }
+
+  /**
+   * Disposes the provider and cleans up resources.
+   */
+  dispose() {
+    this.disposable.dispose();
```
zelliott: Why not just `this.redirectUriEmitter.dispose()` and get rid of the `disposable` member entirely? I'm not sure this indirection provides any value, it obfuscates what's actually being disposed.

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  /**
+   * Handles the given URI by firing the redirect URI event.
+   *
+   * @param uri - The URI to handle.
+   */
+  public handleUri(uri: vscode.Uri): void {
```
zelliott: Omit `public` (it's the default)? Same with `waitForCode`. Follows g3 practices (go/tsjs-style#visibility).

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
+import vscode from "vscode";
+
+const EXCHANGE_TIMEOUT_MS = 60_000;
+
+/**
+ * Provides authentication codes.
+ */
+export interface CodeProvider {
+  waitForCode(
+    scopes: string,
```
zelliott: Optional: Consider making this a single options object param so parameters are always named. For example, [here](https://github.com/googlecolab/colab-vscode/pull/12/files#diff-4bc0acd73ad2b765cf38640753ffd51a2f035b4216661d8635593a465bbb2134R33) it's not immediately obvious what `"123"` is. This also makes this method easier to use in case any of these parameters become optional.

```ts
export interface WaitForCodeOptions {
  ...
}

waitForCode(options: WaitForCodeOptions) {

}
```

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+  ): { promise: Promise<U>; cancel: vscode.EventEmitter<void> } {
+    const cancel = new this.vs.EventEmitter<void>();
+    let subscription: vscode.Disposable | undefined;
+
+    const promise = new Promise<U>((resolve, reject) => {
+      subscription = event((value: T) => {
+        try {
+          adapter(value, resolve, reject);
+        } catch (e: unknown) {
+          const error = e as Error;
```
zelliott: How do you know `e` is an `Error`? You could also do something like:

```
if (e instanceof Error) {
  reject(e);
} else {
  reject(new Error('<some message>'));
}
```

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  /**
+   * Waits for an authorization code for the given scopes and nonce.
+   *
+   * @param scopes - The scopes for which the code is requested.
+   * @param nonce - A unique string to correlate the request and response.
+   * @param token - A cancellation token to cancel the request.
+   * @returns A promise that resolves to the authorization code.
+   * @throws An error if the request times out or is cancelled.
```
zelliott: nit: There are other things that can cause this method to throw (e.g. missing code or nonce), consider updating this comment to clarify that timeouts and cancellations are two examples of how this method can throw, but are not exhaustive.

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+      cancel.event(() => {
+        if (subscription) subscription.dispose();
+        reject(new Error("Cancelled"));
+      });
+    });
+
+    return {
+      promise: promise.finally(() => {
+        subscription?.dispose();
+        return undefined;
```
zelliott: nit: I _think_ you should be able to just do `return;`

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+        try {
+          adapter(value, resolve, reject);
+        } catch (e: unknown) {
+          const error = e as Error;
+          reject(error);
+        }
+      });
+
+      cancel.event(() => {
+        if (subscription) subscription.dispose();
```
zelliott: This `if (subscription)` check is just needed to make TS happy, right? In practice, `subscription` is guaranteed to be defined, because its assignment is run before this cancel event callback handler is created.

If that's all correct, it would be nice to use some open source version of go/guards-and-assertions in cases like these. That is, something like `assertExists(subscription).dispose();` - little utilities to help out in cases where TS is not smart enough. But not blocking.

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+      }
+
+      resolve(code);
+    };
+
+  /**
+   * Creates a promise that rejects if the cancellation token is triggered.
+   */
+  private cancellation(token: vscode.CancellationToken): Promise<string> {
+    const cancelEmitter = new this.vs.EventEmitter<void>();
```
zelliott: What actually listens to `cancelEmitter`? Can it just be removed?

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Waits for requests redirected to the extension to provide authentication codes.
+ */
+export class RedirectUriCodeProvider
+  implements CodeProvider, vscode.UriHandler, vscode.Disposable
+{
+  private readonly pendingNonces = new Map<string, string[]>();
+  private readonly codeExchangePromises = new Map<
+    string,
+    { promise: Promise<string>; cancel: vscode.EventEmitter<void> }
```
zelliott: This type is also used in the return type of `waitForCode`, you could create a common interface (or type alias) like:

```
interface PromiseWithCancel<T> {
  promise: Promise<T>;
  cancel: vscode.EventEmitter<void>;
}
```

and then use it in both places.

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    try {
+      return await Promise.race([
+        codeExchange.promise,
+        exchangeTimeout(),
+        this.cancellation(token),
+      ]);
+    } finally {
+      this.pendingNonces.delete(scopes);
+      codeExchange.cancel.fire();
```
zelliott: So, this is to abruptly cancel the pending `codeExchange.promise` in the case of an external cancellation or a timeout? It's a little odd/unfortunate that it's done even if `codeExchange.promise` happily resolves (i.e. _"why are we always cancelling?_) but it technically doesn't hurt. I don't have an immediate idea for how to improve this off the top of my head without introducing some complexity elsewhere, so fine with this as is, I think.

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+      this.codeExchangePromises.delete(scopes);
+    }
+  }
+
+  private handleEvent: (scopes: string) => PromiseAdapter<vscode.Uri, string> =
+    (scopes) => (uri, resolve, reject) => {
+      const query = new URLSearchParams(uri.query);
+      const code = query.get("code");
+      const nonce = query.get("nonce");
+      if (!code) {
```
zelliott: Right now we will reject if `code` or `nonce` are empty strings. Just confirming that's what we want.

src/auth/redirect.unit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+  });
+
+  it("times out waiting for code exchange", async () => {
+    const gotCode = redirect.waitForCode(
+      DEFAULT_SCOPE,
+      "123",
+      cancellationTokenSource.token,
+    );
+
+    clock.tick(60_001);
```
zelliott: Do we want to follow go/unit-testing-practices?polyglot=java#constants-dos-and-donts?

> Access constants defined in production code (by increasing their visibility if necessary) instead of duplicating the same constants in tests.

If so, you could follow the `export const TEST_ONLY` pattern, or just export the constant.

src/auth/redirect.unit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+      redirects[0].scope,
+      redirects[0].nonce,
+      cancellationTokenSource.token,
+    );
+    const secondCode = redirect.waitForCode(
+      redirects[1].scope,
+      redirects[1].nonce,
+      cancellationTokenSource.token,
+    );
+    // Redirect the second before the first.
```
zelliott, line 114: Why is it important to this test that the second is redirected before the first?

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
+import vscode from "vscode";
+
+const EXCHANGE_TIMEOUT_MS = 60_000;
+
+/**
+ * Provides authentication codes.
+ */
+export interface CodeProvider {
```
zelliott, line 8: Do we expect other implementers of this interface?

Update: I'm looking at https://github.com/googlecolab/colab-vscode/pull/13, I see that this interface is used. Perhaps to prevent clients from knowing about `handleUri`? Is `handleUri` needed really only for unit testing purposes?

src/test/helpers/vscode.ts
zelliott, line 1: I heard you love writing unit tests, do you envision eventually writing unit tests for some of this rapidly growing test double?

src/test/helpers/vscode.ts
```diff
@@ -1,17 +1,236 @@
...
+
+  dispose(): void {
+    if (!this.disposed) {
+      this._token.dispose();
+      this.disposed = true;
+    }
+  }
+}
+
+class DisposableStub implements vscode.Disposable {
```
zelliott: Should this be exported? A client of this module won't be able to properly use `vscodeStub.Disposable` otherwise, because its type isn't exported.

src/test/helpers/vscode.ts
```diff
@@ -1,17 +1,236 @@
...
+      scheme: this.scheme,
+      authority: this.authority,
+      path: this.path,
+      query: this.query,
+      fragment: this.fragment,
+    });
+  }
+}
+
+class TestEventEmitter<T> implements vscode.EventEmitter<T> {
```
zelliott: How should this class behave if the same listener is passed in twice? Right now, `dispose` will only dispose of the first listener. Should `listeners` be a `Set` instead of an array?

src/test/helpers/vscode.ts
```diff
@@ -1,17 +1,236 @@
...
+  authority: string;
+  path: string;
+  query: string;
+  fragment: string;
+}
+
+/**
+ * An approximate test double for vscode.Uri.
+ */
+class TestUri implements vscode.Uri {
```
zelliott: Thoughts on placing each test double in their own file?

src/auth/redirect.unit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+import * as sinon from "sinon";
+import vscode from "vscode";
+import {
+  TestCancellationTokenSource,
+  TestUri,
+  vscodeStub,
+} from "../test/helpers/vscode";
+import { RedirectUriCodeProvider } from "./redirect";
+
+const DEFAULT_SCOPE = "profile email";
```
zelliott: Should this be `DEFAULT_SCOPES`? As I understand, there are two scopes here. If not, then should the first argument to `waitForCode` be `scope`, not `scopes`?

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+  ): Promise<string>;
+}
+
+/**
+ * Waits for requests redirected to the extension to provide authentication codes.
+ */
+export class RedirectUriCodeProvider
+  implements CodeProvider, vscode.UriHandler, vscode.Disposable
+{
+  private readonly pendingNonces = new Map<string, string[]>();
```
zelliott: Can this be `new Map<string, Set<string>>();`? I don't _think_ it changes this class's behavior and makes it marginally more efficient if `waitForCode` is called over and over with the same nonce(s).

src/auth/redirect.ts
```diff
@@ -0,0 +1,188 @@
...
+  /**
+   * Waits for an authorization code for the given scopes and nonce.
+   *
+   * @param scopes - The scopes for which the code is requested.
+   * @param nonce - A unique string to correlate the request and response.
+   * @param token - A cancellation token to cancel the request.
+   * @returns A promise that resolves to the authorization code.
+   * @throws An error if the request times out or is cancelled.
+   */
+  public async waitForCode(
```
zelliott: Suppose `waitForCode` is called twice for the same scopes, with a delay in between:

```
const promise1 = waitForCode('profile', ...);
// 30s delay
const promise2 = waitForCode('profile email', ...);
```

Suppose a full 60s goes by without the first promise resolving. `promise1` will reject (timeout), and the `finally` clause on line 86 runs which cancels the backing `codeExchange` promise. This causes `promise2` to immediately be canceled (right?) - because `promise1` and `promise2` are both the same underlying promise. Is this WAI? Or does this behave differently?

src/auth/redirect.unit.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+describe("RedirectUriCodeProvider", () => {
```
zelliott, line 7: Could we add tests for multiple `waitForCode` calls w/ the same scope? I'm not quite sure what its expected behavior should be, and I'm not sure if it's working correctly (see my comment above starting with _"Suppose `waitForCode` is called twice for the same scopes..."_).

src/auth/redirect.ts
kevineger: I liked the idea of the `GoogleAuthProvider` having a simple `CodeProvider` injected, since it doesn't really care if it was obtained through a redirect or other means.

`handleUri` is implementing the [`UriHandler`](https://code.visualstudio.com/api/references/vscode-api#UriHandler) interface.

I don't have plans for other implementations currently, so if you feel the thin interface is a YAGNI smell, let me know.

src/auth/redirect.ts
kevineger: Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/auth/redirect.ts
kevineger: Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/auth/redirect.ts
kevineger: Done. I'll look into linting this across the repo.

src/auth/redirect.ts
kevineger: The only time that'd be better is the rare case where someone tries to sign in twice before completing the first sign in:

```ts
      const acceptedNonces = this.pendingNonces.get(scopes) ?? [];
      if (!acceptedNonces.includes(nonce)) {
        // Can happen if a sign in is triggered with one set of scopes while
        // there's already an existing sign in with a different set of scopes
        // in flight.
        //
        // Opt to return and wait for the next event, since it's likely we're
        // waiting on the user to continue the auth flow.
        //
        // Currently, a static set of scopes is used so this shouldn't be hit.
        return;
      }
```

---

Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/auth/redirect.ts
kevineger: Good point.

Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/auth/redirect.ts
kevineger: Going to remove entirely since technically in the main cases I care about (timeout/cancel) the promise is rejected, it doesn't throw.

src/auth/redirect.ts
kevineger: > Is this WAI? Or does this behave differently?

Your interpretation is correct. I was largely mirroring the behaviour I saw other OAuth flows doing ([GitHub example](https://github.com/microsoft/vscode/blob/main/extensions/github-authentication/src/github.ts#L42-L63)).

But thanks for bringing it up, upon re-consideration it feels like a bug. I think it'd make more sense for the timer to be tied to each attempt as opposed to attempts grouped by _scopes_.

Added a test case, `"resolves a code while another request times out"`, to specifically cover this case.

---

Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/auth/redirect.ts
kevineger: I think this should now be addressed after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`. Please re-open if you still feel it's off.

src/auth/redirect.ts
kevineger: `handleUri` will `throw`, yep.

src/auth/redirect.ts
kevineger: Sorry, artifact of me trying some other unrelated stuff, forgot to revert this. Thanks for the catch.

src/auth/redirect.ts
kevineger: Done.

src/auth/redirect.ts
kevineger: Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/auth/redirect.ts
kevineger: Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/auth/redirect.unit.test.ts
kevineger: Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/auth/redirect.unit.test.ts
kevineger: Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/auth/redirect.unit.test.ts
kevineger: Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/auth/redirect.unit.test.ts
kevineger: Behaviour wise, it's expected that out-of-order redirects still ultimately resolve the expect promise for a code.

src/test/helpers/vscode.ts
kevineger: Hah... Fair... If it's alright by you, I'll defer that a bit until I'm 100% sure I like this pattern after living with it for a while.

I also think I'd _integration_ (not unit) test the double(s) against the real VS code implementation.

src/test/helpers/vscode.ts
kevineger: I think that's a great thought 😉

src/test/helpers/vscode.ts
kevineger: Good point.

Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/test/helpers/vscode.ts
kevineger: Resolving since this is no longer relevant after I simplified the implementation to simply map `nonce`-s to their `InFlightPromise`.

src/auth/redirect.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    inFlight.resolve(code);
+  }
+}
+
+/**
+ * Creates a new promise with manual resolve/reject handlers.
+ */
+function createPromiseHandlers<T>(): {
```
zelliott, line 90: This helper function is a great idea... it's such a great idea that [it already exists](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers). We can use that instead, and also probably remove the `InFlightPromise` interface and just use [`PromiseWithResolvers`](http://google3/third_party/javascript/node_modules/typescript/stable/lib/lib.es2024.promise.d.ts;l=19;rcl=720549931).

src/auth/redirect.ts
```diff
@@ -0,0 +1,127 @@
...
+    this.inFlightPromises.set(nonce, { promise, resolve, reject });
+
+    try {
+      return await Promise.race([
+        promise,
+        waitForCancellation(token),
+        waitForExchangeTimeout(),
+      ]);
+    } catch (err: unknown) {
+      this.inFlightPromises.delete(nonce);
```
zelliott: You can remove this line, the `finally` clause will run in the case of resolution or rejection (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally).

src/auth/redirect.ts
kevineger: Requires ES2024. Tried briefly to upgrade target but could not get the produced JS to target the newer ES. Likely missing something with how Node is packaging here. Let's try and follow-up on this outside of this change. Easy drop-in once that's sorted.

src/auth/redirect.ts
kevineger: MB, missed that. Thanks.

PR #13, opened 2025-01-23
feat: add auth provider for Google OAuth2

kevineger:
Implements the [AuthenticationProvider](https://code.visualstudio.com/api/references/vscode-api#AuthenticationProvider) interface.

The implementation of this logic was influenced by the GitHub authentication extension. A good starting point when looking at their logic are the [flows](https://github.com/microsoft/vscode/blob/ab172453fd2fe51a2bcf556c64542ee45675b3e8/extensions/github-authentication/src/flows.ts#L77).

### DIFF
--- a/src/auth/provider.ts
+++ b/src/auth/provider.ts
@@ -262,33 +262,76 @@ function parseAuthenticationSessions(
 }
 
 /**
- * Type guard to validate the object is an array of {@link vscode.AuthenticationSession}.
+ * Type guard to check if a value is a string.
  */
-function areAuthenticationSessions(
-  objs: unknown,
-): objs is vscode.AuthenticationSession[] {
-  if (!Array.isArray(objs)) {
+const isString = (value: unknown): value is string => {
+  return typeof value === "string";
+};
+
+/**
+ * Type guard to check if a value is an array of strings.
+ */
+const isStringArray = (value: unknown): value is readonly string[] => {
+  return (
+    Array.isArray(value) && value.every((item) => typeof item === "string")
+  );
+};
+
+/**
+ * Type guard to check if a value matches the AuthenticationSessionAccountInformation shape
+ */
+const isAuthSessionAccountInfo = (
+  value: unknown,
+): value is vscode.AuthenticationSessionAccountInformation => {
+  if (!value || typeof value !== "object") {
     return false;
   }
-  for (const obj of objs) {
-    if (typeof obj !== "object" || obj === null) {
-      return false;
-    }
-    const session = obj as vscode.AuthenticationSession;
-    if (
-      typeof session.id !== "string" ||
-      typeof session.accessToken !== "string" ||
-      typeof session.account !== "object" ||
-      typeof session.account.id !== "string" ||
-      typeof session.account.label !== "string" ||
-      !Array.isArray(session.scopes) ||
-      session.scopes.some((scope: unknown) => typeof scope !== "string")
-    ) {
-      return false;
-    }
+
+  /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access*/
+
+  const account = value as any;
+  return (
+    typeof account === "object" &&
+    account !== null &&
+    "id" in account &&
+    "label" in account &&
+    isString(account.id) &&
+    isString(account.label)
+  );
+};
+
+/**
+ * Type guard to check if a value matches the AuthenticationSession interface
+ */
+const isAuthenticationSession = (
+  value: unknown,
+): value is vscode.AuthenticationSession => {
+  if (!value || typeof value !== "object") {
+    return false;
   }
-  return true;
-}
+
+  /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access*/
+  const session = value as any;
+  return (
+    "id" in session &&
+    "accessToken" in session &&
+    "account" in session &&
+    "scopes" in session &&
+    isString(session.id) &&
+    isString(session.accessToken) &&
+    isAuthSessionAccountInfo(session.account) &&
+    isStringArray(session.scopes)
+  );
+};
+
+/**
+ * Type guard to check if a value is an array of {@link vscode.AuthenticationSession} objects.
+ */
+const areAuthenticationSessions = (
+  value: unknown,
+): value is vscode.AuthenticationSession[] => {
+  return Array.isArray(value) && value.every(isAuthenticationSession);
+};
 
 /**
  * User information queried for following a successful login.
@@ -302,10 +345,16 @@ interface UserInfo {
  * Type guard to validate the object is {@link UserInfo}.
  */
 function isUserInfo(obj: unknown): obj is UserInfo {
+  if (typeof obj !== "object" || obj === null) {
+    return false;
+  }
+
+  /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access*/
+  const userInfo = obj as any;
   return (
-    typeof obj === "object" &&
-    obj !== null &&
-    typeof (obj as UserInfo).name === "string" &&
-    typeof (obj as UserInfo).email === "string"
+    "name" in userInfo &&
+    "email" in userInfo &&
+    isString(userInfo.name) &&
+    isString(userInfo.email)
   );
 }


src/auth/provider.ts
```diff
@@ -0,0 +1,271 @@
+import { OAuth2Client, CodeChallengeMethod } from "google-auth-library";
+import fetch from "node-fetch";
+import { v4 as uuid } from "uuid";
+import vscode from "vscode";
+import { PackageInfo } from "../config/package_info";
+import { CodeProvider } from "./redirect";
+
+const PROVIDER_ID = "google";
+const PROVIDER_LABEL = "Google";
+const REQUIRED_SCOPES = ["profile", "email"];
```
zelliott: nit: You could do `const REQUIRED_SCOPES = ["profile", "email"] as const;` to make this a readonly array.

src/auth/provider.ts
```diff
@@ -0,0 +1,271 @@
...
+  public async getSessions(
+    _scopes: readonly string[] | undefined,
+    _options: vscode.AuthenticationProviderSessionOptions,
+  ): Promise<vscode.AuthenticationSession[]> {
+    const allSessions = await this.context.secrets.get(SESSIONS_KEY);
+    if (!allSessions) {
+      return [];
+    }
+
+    return JSON.parse(allSessions) as vscode.AuthenticationSession[];
```
zelliott: _Not blocking:_ `JSON.parse` returns an `unknown`. The [preference in g3](go/tsjs-style#any-unknown) for dealing with unknowns is to use type guards/assertions. There are some open source type guard libraries [listed here](https://www.reddit.com/r/typescript/comments/sjm8vu/how_to_type_guard_interfaces_beautifully/), perhaps something for us to consider if we foresee a lot of unsafe casting like this in the future.

src/auth/provider.ts
```diff
@@ -0,0 +1,271 @@
...
+
+      this.emitter.fire({
+        added: [session],
+        removed: [],
+        changed: [],
+      });
+
+      return session;
+    } catch (e) {
+      const error = e as Error;
```
zelliott: https://github.com/googlecolab/colab-vscode/pull/12#discussion_r1928998005

src/auth/provider.ts
```diff
@@ -0,0 +1,271 @@
...
+   * @param scopes - Scopes required for the session.
+   * @returns The created session.
+   * @throws An error if login fails.
+   */
+  public async createSession(
+    scopes: string[],
+  ): Promise<vscode.AuthenticationSession> {
+    try {
+      const scopeSet = new Set([...scopes, ...REQUIRED_SCOPES]);
+      const sortedScopes = Array.from(scopeSet).sort();
```
zelliott, line 108: Why sort?

src/auth/provider.ts
```diff
@@ -0,0 +1,271 @@
...
+    }
+  }
+
+  /**
+   * Removes a session by ID.
+   *
+   * @param sessionId - The session ID.
+   */
+  public async removeSession(sessionId: string): Promise<void> {
+    const allSessions = await this.context.secrets.get(SESSIONS_KEY);
```
zelliott: nit: `allSessions` -> `sessionsJson` (or some other name to indicate it's the same sessions as `sessions`, just serialized)

src/auth/provider.ts
```diff
@@ -0,0 +1,271 @@
...
+      throw e;
+    }
+  }
+
+  /**
+   * Removes a session by ID.
+   *
+   * @param sessionId - The session ID.
+   */
+  public async removeSession(sessionId: string): Promise<void> {
```
zelliott: Do we need a "transaction" of sorts here in the case of `removeSession` being called multiple times in parallel?

```
await Promise.all([
  removeSession(1),
  removeSession(2),
]);
```

Depending on the order in which `secrets.(get|store)` promises resolve, these two remove calls could overwrite one another.

---

Thinking more, this could actually apply to all sorts of parallel calls this class could make: e.g. parallel `createSession` and `removeSession` calls...

src/auth/provider.ts
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  /**
+   * Retrieves the authentication sessions that have been persisted.
+   *
+   * @param _scopes - Currently unused.
+   * @param _options - Currently unused.
+   * @returns An array of stored authentication sessions.
+   */
+  public async getSessions(
```
zelliott: Remove `public` (apply throughout).

src/auth/provider.ts
```diff
@@ -0,0 +1,271 @@
...
+          scope: scopes,
+          state: encodedCallbackUri,
+          prompt: "login",
+          code_challenge_method: CodeChallengeMethod.S256,
+          code_challenge: pkce.codeChallenge,
+        });
+
+        await this.vs.env.openExternal(this.vs.Uri.parse(authorizeUrl));
+
+        const code = await promisedCode;
```
zelliott, line 197: If one of the lines above throws, does `promisedCode` need to be cancelled?

src/auth/provider.ts
```diff
@@ -0,0 +1,271 @@
...
+      },
+    });
+    if (!response.ok) {
+      const errorText = await response.text();
+      throw new Error(
+        `Failed to fetch user info: ${response.statusText}. Response: ${errorText}`,
+      );
+    }
+    const json: unknown = await response.json();
+    if (!isUserInfo(json)) {
```
zelliott, line 244: Ah, I see you do have a type guard for this case, just not the instances above. Maybe you guard one and not the other because you "trust" one and not the other? Not sure.

src/auth/provider.ts
```diff
@@ -0,0 +1,271 @@
...
+      throw new Error("Invalid user info, got: " + JSON.stringify(json));
```
zelliott: Template string?

src/auth/provider.unit.test.ts
```diff
@@ -0,0 +1,329 @@
...
+
+      expect({ ...session, id: undefined }).to.deep.equal({
+        ...DEFAULT_SESSION,
+        id: undefined,
+      });
+    });
+  });
+
+  describe("removeSession", () => {
+    it("when no sessions exist", async () => {
```
zelliott: Will read as _"removeSession when no sessions exist"_ - doesn't explain _what_ the expected behavior is. Maybe _"removeSession (removes|succeeds) when no sessions exist"_? Apply throughout.

src/auth/provider.ts
kevineger: A read only tuple, sounds good.

src/auth/provider.unit.test.ts
kevineger: Fixed and improved others which lacked the same.

src/auth/provider.ts
kevineger: Done. All the same to me.

src/auth/provider.ts
kevineger: https://github.com/googlecolab/colab-vscode/pull/18 😉

src/auth/provider.ts
kevineger: Added a type guard and `parseAuthenticationSessions` helper since we parse in two places.

src/auth/provider.ts
kevineger: For now we (Colab) only need the `REQUIRED_SCOPES` which I'm forcing to exist on each new session. But a more robust auth provider (used beyond what we **currently** need) will need to use the `scopes` provided to `getSessions` to ensure the obtained sessions has the requested scopes. This sort makes that future expansion easier.

Example for where GitHub does this "tier" of scoping.

image

src/auth/provider.ts
kevineger: Done.

src/auth/provider.ts
kevineger: That's a good call-out.

The GitHub auth provider doesn't lock (e.g. [`removeSession`](https://github.com/microsoft/vscode/blob/beb5ab21b3768d4046c8990c854cbe0ee110e648/extensions/github-authentication/src/github.ts#L369-L401)). However, the interface / docs don't make any guarantees surrounding concurrency. Like the GitHub implementation, I assume that VS Code handles auth session management (create/remove) in a "single-threaded" manner.

---

If it's alright by you, I'll tackle this in a follow-up after the core functionality is in.

src/auth/provider.ts
kevineger: Done.

src/auth/provider.ts
kevineger: I think it's fine to let the timeout or explicit user cancellation cancel it (promise handling in `redirect.ts`).

src/auth/provider.ts
kevineger: Added type guard for `vscode.AuthenticationSession[]` as done here.

src/auth/provider.ts
```diff
@@ -0,0 +1,311 @@
...
+    if (typeof obj !== "object" || obj === null) {
+      return false;
+    }
+    const session = obj as vscode.AuthenticationSession;
+    if (
+      typeof session.id !== "string" ||
+      typeof session.accessToken !== "string" ||
+      typeof session.account !== "object" ||
+      typeof session.account.id !== "string" ||
+      typeof session.account.label !== "string" ||
```
zelliott: As implemented, this type guard may throw if `objs` is not an `AuthenticationSession[]`. For example, `session.account` may be `undefined`, and `undefined.id` will throw.

There may be more instances like this in this type guard, this is just the first one I saw. Fix this and do a quick pass for any additional issues?

kevineger: Done.

Following this change I may switch to [zod](https://www.npmjs.com/package/zod) or something.

PR #14, opened 2025-01-23
feat: add client for interacting with Colab

kevineger:


src/colab/client.ts
```diff
@@ -0,0 +1,169 @@
...
+export class ColabClient {
+  private readonly httpsAgent?: https.Agent;
+
+  constructor(
+    private readonly domain: URL,
+    private session: () => Promise<AuthenticationSession>,
+  ) {
+    // TODO: Temporary workaround to allow self-signed certificates
+    // in local development.
+    if (domain.host === "https://localhost:8888") {
```
wh1ter4bb1t-js: temporary or not this is not a guaranteed path, I'm fine if this is the default path; but I think this probably should be an env variable so people can pass the correct path.

src/colab/client.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+
+  private async getAssignment(
+    notebookHash: string,
+    variant: Variant,
+    accelerator?: Accelerator,
+  ): Promise<AssignmentToken | AssignedAssignment> {
+    const url = this.assignUrl(notebookHash, variant, accelerator);
+    const response = await this.issueRequest<
```
wh1ter4bb1t-js, line 98: considering your defaulting to a response with "assigned" type... have you considered error handling, a good try-catch for all those possible 400s and 500s

src/colab/client.ts
```diff
@@ -0,0 +1,169 @@
...
+    const requestHeaders = new fetch.Headers(headers);
+    requestHeaders.set("Accept", "application/json");
+    requestHeaders.set("Authorization", `Bearer ${authSession.accessToken}`);
+    const request = new Request(endpoint, {
+      method,
+      headers: requestHeaders,
+      agent: this.httpsAgent,
+    });
+    const response = await fetch(request);
+    if (!response.ok) {
```
wh1ter4bb1t-js, line 151: `ok` is only for the 200 range. redirects are not included which will cause issues if we ever change the path. Is that okay?

src/colab/client.unit.test.ts
```diff
@@ -0,0 +1,228 @@
+import { expect } from "chai";
+import { Response } from "node-fetch";
+import * as nodeFetch from "node-fetch";
```
wh1ter4bb1t-js, line 3: nit: a single import statement

src/colab/client.unit.test.ts
```diff
@@ -0,0 +1,228 @@
...
+        assignmentsCount: 3,
+        eligibleGpus: [Accelerator.T4],
+        ineligibleGpus: [Accelerator.A100, Accelerator.L4],
+        freeCcuQuotaInfo: {
+          remainingTokens: 4,
+          nextRefillTimestampSec: 5,
+        },
+      };
+      fetchStub
+        .withArgs(matchAuthorizedRequest("tun/m/ccu-info", "GET"))
```
wh1ter4bb1t-js, line 63: nit: maybe annoying but set as a constant in `client.ts` and use that there and here so that you don't need to change strings across files

src/colab/client.unit.test.ts
```diff
@@ -0,0 +1,228 @@
...
+        variant: Variant.DEFAULT,
+        machineShape: Shape.STANDARD,
+        runtimeProxyInfo: {
+          token: "mock-token",
+          tokenExpiresInSeconds: 42,
+          url: "https://mock-url.com",
+        },
+      };
+      fetchStub
+        .withArgs(matchAuthorizedRequest("tun/m/assign", "GET"))
```
wh1ter4bb1t-js, line 103: same as above

src/colab/client.unit.test.ts
```diff
@@ -0,0 +1,228 @@
...
+    it("without existing assignment", async () => {
+      const mockGetResponse: GetAssignmentResponse = {
+        acc: Accelerator.A100,
+        nbh: NOTEBOOK_HASH,
+        p: false,
+        token: "mock-xsrf-token",
+        variant: Variant.DEFAULT,
+      };
+      fetchStub
+        .withArgs(matchAuthorizedRequest("tun/m/assign", "GET"))
```
wh1ter4bb1t-js, line 124: same as above

src/colab/client.unit.test.ts
```diff
@@ -0,0 +1,228 @@
...
+        variant: Variant.DEFAULT,
+        machineShape: Shape.STANDARD,
+        runtimeProxyInfo: {
+          token: "mock-token",
+          tokenExpiresInSeconds: 42,
+          url: "https://mock-url.com",
+        },
+      };
+      fetchStub
+        .withArgs(matchAuthorizedRequest("tun/m/assign", "POST"))
```
wh1ter4bb1t-js, line 145: same as above

src/colab/client.unit.test.ts
```diff
@@ -0,0 +1,228 @@
...
+      assignmentsCount: 3,
+      eligibleGpus: [Accelerator.T4],
+      ineligibleGpus: [Accelerator.A100, Accelerator.L4],
+      freeCcuQuotaInfo: {
+        remainingTokens: 4,
+        nextRefillTimestampSec: 5,
+      },
+    };
+    fetchStub
+      .withArgs(matchAuthorizedRequest("tun/m/ccu-info", "GET"))
```
wh1ter4bb1t-js, line 188: same as above

src/colab/client.unit.test.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    const result = await client.ccuInfo();
+
+    expect(result).to.deep.equal(mockResponse);
+    sinon.assert.calledOnce(fetchStub);
+  });
+
+  it("handles invalid JSON responses", async () => {
+    fetchStub
+      .withArgs(matchAuthorizedRequest("tun/m/ccu-info", "GET"))
```
wh1ter4bb1t-js, line 198: same as above

src/colab/client.ts
```diff
@@ -0,0 +1,169 @@
...
+} from "./api";
+
+const XSSI_PREFIX = ")]}'\n";
+const XSRF_HEADER_KEY = "X-Goog-Colab-Token";
+const CCU_INFO_ENDPOINT = "/tun/m/ccu-info?authuser=0";
+const ASSIGN_ENDPOINT = "/tun/m/assign?authuser=0";
+
+// To discriminate the type of GET assignment responses.
+interface AssignmentToken extends GetAssignmentResponse {
+  type: "to_assign";
```
wh1ter4bb1t-js: creating an assignment type enum and using those values instead will likely be safer, string literals are only guaranteed to break at runtime, we probably want to know whats up at compile time.

src/colab/api.ts
```diff
@@ -0,0 +1,128 @@
+/**
+ * @fileoverview API types for interacting with Colab's backend.
+ *
+ * As mentioned throughout several of the relevant fields, a lot of the name choices
+ * are due to historical reasons and are not ideal.
+ */
+
+export interface FreeCCUQuotaInfo {
+  // Number of tokens remaining in the "USAGE-mCCUs" quota group (remaining
+  // free usage allowance in milli-CCUs).
```
zelliott: Ideally comments like these should be JSDoc comments so that they appear in intellisense tooltip hints and stuff. Doesn't apply to _all_ of the comments in this file, will defer to your judgment.

src/colab/api.ts
```diff
@@ -0,0 +1,128 @@
...
+  acc: Accelerator;
+  // The notebook ID hash. Same as the `nbh` query parameter from the request.
+  nbh: string;
+  // Whether or not Recaptcha should prompt.
+  p: boolean;
+  // XSRF token to be provided when posting an assignment.
+  token: string;
+  /**
+   * The string representation of the pool {@link Variant}.
+   * TODO: should the enum just be string values? Check with the return from POST assign.
```
zelliott: The biggest nit ever: This TODO might be something I'd put in a non-JSDoc comment above/below the JSDoc comment because people using `GetAssignmentResponse` don't care about it.

```
/**
 * The string representation...
 */
// TODO: ...
variant: Variant;
```

src/colab/client.ts
```diff
@@ -0,0 +1,169 @@
...
+    notebookHash: string,
+    variant: Variant,
+    accelerator?: Accelerator,
+  ): Promise<Assignment> {
+    const getAssignment = await this.getAssignment(
+      notebookHash,
+      variant,
+      accelerator,
+    );
+    switch (getAssignment.type) {
```
zelliott: Add exhaustive check?

src/colab/client.ts
```diff
@@ -0,0 +1,169 @@
...
+const ASSIGN_ENDPOINT = "/tun/m/assign?authuser=0";
+
+// To discriminate the type of GET assignment responses.
+interface AssignmentToken extends GetAssignmentResponse {
+  type: "to_assign";
+}
+
+// To discriminate the type of GET assignment responses.
+interface AssignedAssignment extends Assignment {
+  type: "assigned";
```
zelliott: nit: You'll notice `type` is colored funky because it's a keyword. Sometimes I see people use `kind` instead in TS. Another option is `assignmentType`. But also feel free to leave as-is of course.

src/colab/client.ts
```diff
@@ -0,0 +1,169 @@
...
+    variant: Variant,
+    accelerator?: Accelerator,
+  ): Promise<Assignment> {
+    const url = this.assignUrl(notebookHash, variant, accelerator);
+    return this.issueRequest<Assignment>(url, "POST", [
+      [XSRF_HEADER_KEY, xsrfToken],
+    ]);
+  }
+
+  private assignUrl(
```
zelliott: Rename to `(get|build)AssignUrl`? On initial read, I thought `assign` was the verb (i.e. this method _assigns_ a URL), but that's not correct.

src/colab/client.ts
```diff
@@ -0,0 +1,169 @@
...
+      agent: this.httpsAgent,
+    });
+    const response = await fetch(request);
+    if (!response.ok) {
+      throw new Error(
+        `Failed to ${method} ${endpoint.toString()}: ${response.statusText}`,
+      );
+    }
+    const body = await response.text();
+    return JSON.parse(stripXssiPrefix(body)) as T;
```
zelliott, line 158: Unsafe cast, this one's trickier though because of the generic. This method could be updated to accept a custom type guard (e.g. maybe `guard: (obj: unknown): unknown is T`), or it could just return `Promise<unknown>` and the caller can narrow the type itself.

src/colab/client.unit.test.ts
```diff
@@ -0,0 +1,228 @@
...
+    } as AuthenticationSession);
+    client = new ColabClient(new URL(DOMAIN), sessionStub);
+  });
+
+  afterEach(() => {
+    fetchStub.restore();
+  });
+
+  describe("getCCUInfo", () => {
+    it("success", async () => {
```
zelliott: nit: https://github.com/googlecolab/colab-vscode/pull/13#discussion_r1932495943 (apply throughout this test file)

src/colab/client.unit.test.ts
```diff
@@ -0,0 +1,228 @@
...
+      };
+      fetchStub
+        .withArgs(matchAuthorizedRequest("tun/m/ccu-info", "GET"))
+        .resolves(
+          new Response(withXSSI(JSON.stringify(mockResponse)), { status: 200 }),
+        );
+
+      const result = await client.ccuInfo();
+
+      expect(result).to.deep.equal(mockResponse);
```
zelliott: nit: We're using chai... right? We could use their [async matchers](https://www.chaijs.com/plugins/chai-as-promised/) throughout this file, instead of awaiting _then_ expecting.

src/colab/client.ts
```diff
@@ -0,0 +1,169 @@
...
+   * @param variant - The machine variant to assign.
+   * @param accelerator - The accelerator to assign.
+   * @returns The assignment which is assigned to the user.
+   */
+  async assign(
+    notebookHash: string,
+    variant: Variant,
+    accelerator?: Accelerator,
+  ): Promise<Assignment> {
+    const getAssignment = await this.getAssignment(
```
zelliott: I would've expected this variable to be called `assignment`, or `assignmentResult`, or `assignmentInfo`... basically to be a noun.

src/colab/client.unit.test.ts
```diff
@@ -0,0 +1,228 @@
...
+        id: "mock-account-id",
+        label: "mock-account-label",
+      },
+      scopes: ["foo"],
+    } as AuthenticationSession);
+    client = new ColabClient(new URL(DOMAIN), sessionStub);
+  });
+
+  afterEach(() => {
+    fetchStub.restore();
```
zelliott: Restore/reset sinon?

src/colab/api.ts
kevineger: Good point!

src/colab/api.ts
kevineger: Good point. But I'll do you one better and remove it since I confirmed.

src/colab/client.ts
kevineger: I think using a literal type as done here is pretty starndard when [discriminating unions](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions).

If code tries to discriminate the union using a different literal, that'll blow up at compile with `Type '"does_not_exist"' is not comparable to type '"to_assign" | "assigned"'`.

src/colab/client.ts
kevineger: Good point. Done.

src/colab/client.ts
kevineger: I'll just use `hostname` to skip the port. This is temporary hardcoding while we're early in development.

src/colab/client.ts
kevineger: Sorry, artifact of a refactor. Fixed.

src/colab/client.ts
kevineger: I love exhaustive checking, but it's not needed here.

The method returns a `Promise<Assignment>`, so if you're missing a `case`, compiler will complain with _Function lacks ending return statement and return type does not include 'undefined'_.

src/colab/client.ts
kevineger: Method returns `Promise<AssignmentToken | AssignedAssignment>` so any errors thrown from the fetch are fine to bubble up. Not sure I follow. Please re-open if I'm overlooking something!

src/colab/client.ts
kevineger: Been working in Go too much (which would omit the `get` here). Fixed.

src/colab/client.ts
kevineger: Would prefer to set `{ redirect: "follow" }` if we cared to support redirects. For now I think it's fine to omit. We're in the midst of developing a new assignments API so a lot of this is going to change. For now, simplicity is 👑.

src/colab/client.ts
kevineger: API surface is kinda large to write all the explicit type guards 😫. I'll defer type guarding and instead add `zod` validation (or something functionally equivalent) in a standalone PR.

src/colab/client.unit.test.ts
kevineger: `import * as nodeFetch, { Response } from "node-fetch";` is not valid syntax, if that's what you mean?

src/colab/client.unit.test.ts
kevineger: `fetchStub` is our only stub, but I can just restore all of `sinon` if preferred.

src/colab/client.unit.test.ts
kevineger: Done.

src/colab/client.unit.test.ts
kevineger: I'd prefer not to export strings just for test.

src/colab/client.unit.test.ts
kevineger: Sure, but I'll still need to `await` the _as-promised_ assertion in the cases where I'm asserting on the `fetchStub` after.

src/colab/client.unit.test.ts
kevineger: Same answer / centralize discussion in root comment.

src/colab/client.unit.test.ts
kevineger: Same answer / centralize discussion in root comment.

src/colab/client.unit.test.ts
kevineger: Same answer / centralize discussion in root comment.

src/colab/client.unit.test.ts
kevineger: Same answer / centralize discussion in root comment.

src/colab/client.unit.test.ts
kevineger: Same answer / centralize discussion in root comment.

PR #15, opened 2025-01-23
feat: real Jupyter server provider

kevineger:
[Demo of the e2e working locally](https://github.com/user-attachments/assets/cf93bff2-f0a7-4341-a49c-436964f46bad).

### DIFF
--- a/src/jupyter/provider.ts
+++ b/src/jupyter/provider.ts
@@ -66,6 +66,7 @@ export class ColabJupyterServerProvider
     server: JupyterServer,
     _token: CancellationToken,
   ): ProviderResult<JupyterServer> {
+    // TODO: Derive NBH.
     const nbh = "booooooooooooooooooooooooooooooooooooooooooo";
 
     const colabServer = SERVERS.get(server.id);
--- a/src/jupyter/provider.unit.test.ts
+++ b/src/jupyter/provider.unit.test.ts
@@ -151,7 +151,6 @@ describe("ColabJupyterServerProvider", () => {
     it("rejects for server assignments without connection information", async () => {
       const server = SERVERS.get("gpu-a100");
       assert.isDefined(server);
-      // TODO: Derive NBH.
       const nbh = "booooooooooooooooooooooooooooooooooooooooooo";
       const assignment: Assignment = {
         accelerator: Accelerator.A100,


src/extension.ts
```diff
@@ -1,32 +1,43 @@
...
+import { GoogleAuthProvider } from "./auth/provider";
+import { RedirectUriCodeProvider } from "./auth/redirect";
+import { ColabClient } from "./colab/client";
 import { getJupyterApi } from "./jupyter/jupyter-extension";
-import { ColabJupyterServerProvider, RpConfig } from "./jupyter/provider";
+import { ColabJupyterServerProvider } from "./jupyter/provider";
+
+const CLIENT_ID =
+  "1014160490159-8bdmhbrghjfch5sb8ltuofo1mk1totmr.apps.googleusercontent.com";
+const CLIENT_NOT_SO_SECRET = "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";
```
wh1ter4bb1t-js, line 11: These are 5ever in the checksum, are you sure you don't want to opt for something like env files

kevineger: If the client (this extension) is installed on the user's device, there isn't much of a choice other than to embed the client secret into the code. E.g. `gcloud` and [Cloud Code](https://marketplace.visualstudio.com/items?itemName=GoogleCloudTools.cloudcode) do this.

We're at least taking the extra step of using PKCE in the auth provider.

---

Loosely related - we'll be rolling these before any release (or making this repo public) anyways; in the off-chance we go with a different OAuth flow.

kevineger:
> I haven't looked at this too much but I assume you are ripping out the RpConfig stuff. Should we keep that around until we have a way to connect to a specific VM of our choice? essentially if I build a vm with specific changes I want to connect to _that_ VM. I would not be able to do that with this change.

Discussed this IRL.

src/extension.ts
```diff
@@ -1,32 +1,43 @@
...
+  const serverProvider = new ColabJupyterServerProvider(
+    vscode,
+    jupyter,
+    colabClient,
+  );
 
-  return { baseUri: vscode.Uri.parse(baseUrl), token };
+  context.subscriptions.push(disposeUriHandler);
+  context.subscriptions.push(authProvider);
+  context.subscriptions.push(serverProvider);
```
zelliott: `.push(every, thing, at, once)` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push#syntax)

src/jupyter/provider.ts
```diff
@@ -1,66 +1,62 @@
...
+      return Array.from(SERVERS.values()).filter((server) => {
+        if (server.variant !== Variant.GPU) {
+          return true;
+        }
+        // Check both to make introducing new accelerators safer.
+        const eligibleGpu =
+          server.accelerator && eligibleGpus.has(server.accelerator);
+        const ineligibleGpu =
+          server.accelerator && ineligibleGpus.has(server.accelerator);
+        // TODO: Provide a ⚠️ warning for the servers which are inneligible for the user.
```
zelliott: ineligible*

src/jupyter/provider.unit.test.ts
```diff
@@ -1,105 +1,213 @@
...
+
+      sinon.assert.calledOnce(registrationDisposable.dispose);
+    });
   });
 
-  it("provides Jupyter servers", async () => {
-    const providedServers = await serverProvider.provideJupyterServers(
-      {} as CancellationToken,
-    );
+  describe("provideJupyterServers", () => {
```
zelliott, line 71: Mixing " and ' throughout, maybe there's a lint rule we can install?

src/extension.ts
kevineger: All the same to me! Done.

src/jupyter/provider.ts
kevineger: Time to setup a spell check lint!

src/jupyter/provider.unit.test.ts
kevineger: Will do in a standalone PR! (And leave these intentionally mucked up so there's a satisfying diff to look at 🤡.)

PR #17, opened 2025-01-28
feat: Adds custom fetch logic

wh1ter4bb1t-js:
Adds custom fetch logic to add proxy tokens to headers of all fetchs.
Because the original implementation in the Jupyter extension uses the
the 'node-fetch', I have opted to do the same as well since that will
likely have the least amount of edge cases in the future.

Because this requires us to rely on the jupyter extensions proposed API,
the package version has been downgraded to the latest version with the
"proposed" tag.

### DIFF
--- a/src/jupyter/provider.unit.test.ts
+++ b/src/jupyter/provider.unit.test.ts
@@ -250,7 +250,7 @@ describe("ColabJupyterServerProvider", () => {
       {
         headers: new Headers({
           "X-Colab-Runtime-Proxy-Token": assignment.runtimeProxyInfo.token,
-        })
+        }),
       },
     );
   });


package.json
```diff
@@ -71,7 +72,9 @@
     "typescript-eslint": "^8.15.0"
   },
   "dependencies": {
-    "@vscode/jupyter-extension": "^1.1.1",
-    "glob": "^11.0.0"
+    "@vscode/jupyter-extension": "1.0.93",
+    "glob": "^11.0.0",
+    "node-fetch": "^2.6.1",
+    "vsce": "^2.15.0"
```
kevineger: Not needed.

src/jupyter/jupyter_extension.ts
```diff
@@ -1,5 +1,5 @@
-import { Jupyter } from "@vscode/jupyter-extension";
-import { extensions } from "vscode";
+import {Jupyter} from "@vscode/jupyter-extension";
+import {extensions} from "vscode";
```
kevineger: I suspect this is / will fail `lint`.

src/jupyter/provider.ts
```diff
@@ -92,6 +98,23 @@ class ColabJupyterServer implements JupyterServer {
...
       baseUrl: config.baseUri,
       headers: { "X-Colab-Runtime-Proxy-Token": config.token },
-    };
+
+      // Overwrite the fetch method so that we can add our own custom headers to all requests made by the Jupyter extension.
+      fetch: async (info: nodeFetch.RequestInfo, init?: nodeFetch.RequestInit) => {
+        if (!init) {
+          init = {}
+        }
+        init.headers = {...init.headers, "X-Colab-Runtime-Proxy-Token": config.token}
```
kevineger: Could we be consistent between here and `client.ts`, which extends headers like:

```ts
// ...
const requestHeaders = new fetch.Headers(headers);
requestHeaders.set("Accept", "application/json");
// ...
```

src/jupyter/provider.ts
```diff
@@ -92,6 +98,23 @@ class ColabJupyterServer implements JupyterServer {
...
+
+        if (typeof info !== 'string' && !("href" in info)) {
+          info = new nodeFetch.Request(info.url, info)
+        }
+
+        // We use the method that Jupyter extension defaults to so as to not have unexpected results.
+        // https://github.com/microsoft/vscode-jupyter/blob/06c8ae1a5b9620043c59c08e0c2debb33847e5e6/src/kernels/jupyter/session/jupyterRequestCreator.node.ts#L142
+        // We have to disable all type rules because the response of fetch requests are always 'unknown' and we cannot account for all the call sites in the Jupyter extension.
```
kevineger: Where are we doing that?

src/jupyter/provider.ts
```diff
@@ -92,6 +98,23 @@ class ColabJupyterServer implements JupyterServer {
...
+      // Overwrite the fetch method so that we can add our own custom headers to all requests made by the Jupyter extension.
+      fetch: async (info: nodeFetch.RequestInfo, init?: nodeFetch.RequestInit) => {
+        if (!init) {
+          init = {}
+        }
+        init.headers = {...init.headers, "X-Colab-Runtime-Proxy-Token": config.token}
+
+        if (typeof info !== 'string' && !("href" in info)) {
+          info = new nodeFetch.Request(info.url, info)
+        }
```
kevineger: Sorry I'm not totally picking up on the logic here (brain is fried after a long day 💩)... why do we use our `init` in some cases but not all? Also why specifically when `typeof info !== 'string' && !("href" in info)`?

src/jupyter/provider.unit.test.ts
```diff
@@ -87,11 +88,13 @@ describe("ColabJupyterServerProvider", () => {
 
   expectedServers.forEach((server) => {
     it(`resolves the '${server.id}' Jupyter server`, async () => {
+      const fakeFetch = () =>  Promise.resolve();
```
kevineger: Could we be consistent in how we're stubbing out `fetch`? [Example](https://github.com/googlecolab/colab-vscode/blob/58bcd3d176a02e86c6406700f1c7e74b29fe2ab3/src/colab/client.unit.test.ts#L32).

src/jupyter/provider.unit.test.ts
```diff
@@ -100,7 +103,37 @@ describe("ColabJupyterServerProvider", () => {
...
+  it(`uses correct values from custom fetch`, async () => {
+    const fetchUrl = 'http://1234.com';
+    let fetchInfo: nodeFetch.RequestInfo = '';
+    let fetchInit: nodeFetch.RequestInit = {};
+
+    sinon.replace(nodeFetch, "default", ((info: nodeFetch.RequestInfo, init: nodeFetch.RequestInit) => {
+      // eslint-disable-next-line
+      fetchInfo = info;
+      // eslint-disable-next-line
+      fetchInit = init;
```
kevineger: These es lint disable stanzas aren't doing anything.

src/jupyter/provider.unit.test.ts
```diff
@@ -100,7 +103,37 @@ describe("ColabJupyterServerProvider", () => {
...
+
+
+    const expectedHeaders = new Headers();
+    expectedHeaders.append("X-Colab-Runtime-Proxy-Token", config.token)
+
+    const resolvedServer = await serverProvider.resolveJupyterServer(expectedServers[0], {} as CancellationToken)
+
+    if (resolvedServer?.connectionInformation?.fetch) {
+      // eslint-disable-next-line
+      await resolvedServer.connectionInformation.fetch(fetchUrl);
```
kevineger: Instead of disabling the lint, could we instead do an assertion that `connectionInformation` is defined?

src/jupyter/provider.ts
wh1ter4bb1t-js: dead comment. Deleted.

src/jupyter/provider.ts
wh1ter4bb1t-js: - init: I'm just taking the init object and passing the new headers through it and passing that off to the fetch method. This could likely be done in either init or info, but this is how the Jupyter extension does it, which is why I went this way.

- conditional: there are three types that fall into a the `RequestInfo` type:
1. string
2. Object with href property
3. Request Object

which...
- info: node-fetch requires a full path url, but depending on how you get the parsed url from a Request object it may not always be the full path. This was throwing errors. To fix that, I create a Request pass all the old info as an init and use the `url` getter method to construct the full path.

src/jupyter/provider.unit.test.ts
wh1ter4bb1t-js: huh, they were yelling at me pretty hard when I first wrote this. Done.

src/jupyter/jupyter-extension.unit.test.ts
```diff
@@ -32,6 +32,7 @@ describe("Jupyter Extension", () => {
         exports: {
           kernels: {
             getKernel: sinon.stub(),
+            onDidStart: sinon.stub(),
```
kevineger, line 35: Are we using this?

src/jupyter/provider.ts
```diff
@@ -90,6 +91,24 @@ export class ColabJupyterServerProvider
           connectionInformation: {
             baseUrl: this.vs.Uri.parse(url),
             headers: { "X-Colab-Runtime-Proxy-Token": token },
```
kevineger: Could we define a shared constant for this header since it's used twice (statically in the `connectionInformation.headers` and now in the `fetch` implementation)?

src/jupyter/provider.ts
```diff
@@ -3,6 +3,7 @@ import {
   JupyterServer,
   JupyterServerProvider,
 } from "@vscode/jupyter-extension";
+import * as nodeFetch from "node-fetch";
```
kevineger: Nit: use a named import over N.S. import.

Sorry I don't have the linter configured for this (yet), but internal guidance: [go/tsjs-style#namespace-versus-named-imports](http://go/tsjs-style#namespace-versus-named-imports). It's more idiomatic, avoid an unnecessary wrap and since `node-fetch` has this default export we can directly import it without the default lookup.

`import fetch from "node-fetch";` as done [here](https://github.com/googlecolab/colab-vscode/blob/8ee92bbaccd74fca52c1f06c9f3e6151a2e69891/src/colab/client.ts#L2).

kevineger: Apply to `src/jupyter/provider.unit.test.ts` as well.

src/jupyter/provider.unit.test.ts
```diff
@@ -192,6 +194,9 @@ describe("ColabJupyterServerProvider", () => {
         cancellationToken,
       );
 
+      assert.isDefined(resolvedServer?.connectionInformation?.fetch);
+      sinon.replace(resolvedServer.connectionInformation, "fetch", fetchStub);
+
```
kevineger: Nit: delete `\n`. Test body should be split by arrange-act-assert.

src/jupyter/provider.unit.test.ts
```diff
@@ -201,11 +206,63 @@ describe("ColabJupyterServerProvider", () => {
...
+          fetch: fetchStub.resolves(),
         },
         variant: server.variant,
         accelerator: server.accelerator,
       };
       expect(resolvedServer).to.deep.equal(expectedResolvedServer);
     });
   });
+
+  it(`uses correct values from custom fetch`, async () => {
```
kevineger: Unnecessary string template. Use `"regular string"`.

src/jupyter/provider.unit.test.ts
```diff
@@ -201,11 +206,63 @@ describe("ColabJupyterServerProvider", () => {
...
+
+    let fetchInfo: nodeFetch.RequestInfo = "";
+    let fetchInit: nodeFetch.RequestInit = {};
+    sinon.replace(nodeFetch, "default", ((
+      info: nodeFetch.RequestInfo,
+      init: nodeFetch.RequestInit,
+    ) => {
+      fetchInfo = info;
+      fetchInit = init;
+    }) as typeof nodeFetch.default);
```
kevineger: Can we setup the sinon stub without the cast? I think if you type the stub as done in `src/colab/client.unit.test.ts` you won't need to cast.

src/jupyter/provider.unit.test.ts
```diff
@@ -201,11 +206,63 @@ describe("ColabJupyterServerProvider", () => {
...
+      machineShape: Shape.STANDARD,
+      runtimeProxyInfo: {
+        token: "mock-token",
+        tokenExpiresInSeconds: 42,
+        url: "https://mock-url.com",
+      },
+    };
+    colabClientStub.assign.withArgs(nbh, server.variant).resolves(assignment);
+
+    let fetchInfo: nodeFetch.RequestInfo = "";
```
kevineger: Rather than swap in this fake which sets these variables, you can `sinon.assert.calledOnceWithMatch` or `sinon.assert.calledOnceWithExactly`.

https://sinonjs.org/releases/latest/assertions for more info. Should also be a few examples throughout other unit tests.

src/jupyter/provider.unit.test.ts
kevineger: Also apply to new test starting on L218.

src/jupyter/provider.ts
kevineger: I'm still missing why we need to modify `info` at all if we're just passing it along to default `fetch`.

> This was throwing errors.

What errors? From the Jupyter extension?

As we discussed a bit offline, I suspect what might be the issue here is that the Jupyter extension is using a different version of `node-fetch` than we are and that's what's causing the error. That identifies that the Jupyter extension's choice in the proposed API to allow overriding `fetch` is a leaky abstraction. They aren't running tests nor validating all possible versions of `fetch` (nor should they...). It'd be nice if they could instead offer a hook to modify/init the headers. Let's do some investigation into the errors and if that assumption is right or not.

wh1ter4bb1t-js: This does seem to be an issue with the version, it's and interesting thing where when I saw the issue using older _and_ newer versions than the one pinned in Jupyter extension. I don't know if maybe I just had a messed up workspace since the Jupyter extension `package.json` indicates newer than. I pinned the version to the base version that Jupyter extension is using for the time being, and removed this.

src/jupyter/jupyter-extension.unit.test.ts
wh1ter4bb1t-js: This is part of the [proposed API](https://github.com/microsoft/vscode-jupyter/blob/9f67a495511328861b8e6196c5b4b498e317f4c0/src/api.proposed.kernelStartHook.d.ts#L11), so we don't use it, but the compiler does get angry at us for not defining it.

src/jupyter/provider.ts
wh1ter4bb1t-js: Done

src/jupyter/provider.unit.test.ts
wh1ter4bb1t-js: Done.

src/jupyter/provider.unit.test.ts
wh1ter4bb1t-js: Done.

src/jupyter/provider.unit.test.ts
wh1ter4bb1t-js: Done.

src/jupyter/provider.ts
```diff
@@ -3,11 +3,16 @@ import {
...
+import fetch, { Headers } from "node-fetch";
+import vscode, { CancellationToken, ProviderResult } from "vscode";
+import { CCUInfo, Variant } from "../colab/api";
+import { ColabClient } from "../colab/client";
+import { SERVERS } from "./servers";
+
+/**
+ * Header key for the runtime proxy token.
+ */
+export const COLAB_RUNTIME_PROXY_TOKEN_HEADER = "X-Colab-Runtime-Proxy-Token";
```
kevineger: Nit: personally not a fan of exporting just for the sake of tests. Here I'd lean towards duplicating the value in test instead but fine either way.

package.json
```diff
@@ -78,10 +78,10 @@
     "typescript-eslint": "^8.15.0"
   },
   "dependencies": {
-    "@vscode/jupyter-extension": "^1.1.1",
+    "@vscode/jupyter-extension": "1.0.93",
     "glob": "^11.0.0",
     "google-auth-library": "^9.15.0",
-    "node-fetch": "^2.7.0",
+    "node-fetch": "2.6.7",
```
kevineger: Any ideas for how we could/should keep this in sync with the Jupyter extension version?

Would you consider [filing an issue](https://github.com/microsoft/vscode-jupyter/issues/new?template=1_bug_report.md) identifying why the API choice to support `fetch` overriding is problematic as versions skew? Curious what the maintainer says or if an alternate approach may be developed.

src/jupyter/provider.unit.test.ts
```diff
@@ -186,26 +191,74 @@ describe("ColabJupyterServerProvider", () => {
...
         },
         variant: server.variant,
         accelerator: server.accelerator,
       };
+
+      const resolvedServer = await serverProvider.resolveJupyterServer(
+        server,
+        cancellationToken,
+      );
+      assert.isDefined(resolvedServer?.connectionInformation?.fetch);
```
kevineger, line 211: Nit: as previously mentioned, structure test bodies by arrange-act-assert, separating them by newlines. In other words, here, put a newline between L212 and L213 and remove the newline on L215. As what's been correctly done in the added test case starting on L220.

```ts
it("test case", async () => {
  // Arrange block
  // \n
  // Act block
  // \n
  // Assert block
});
```

src/jupyter/provider.unit.test.ts
```diff
@@ -186,26 +191,74 @@ describe("ColabJupyterServerProvider", () => {
...
+      sinon.replace(resolvedServer.connectionInformation, "fetch", fetchStub);
+
       expect(resolvedServer).to.deep.equal(expectedResolvedServer);
     });
   });
+
+  it("uses correct values from custom fetch", async () => {
```
kevineger: Nit: go/unit-testing-practices?polyglot=typescript#naming - _Test names should summarize the behavior being tested and its expected outcome._

Recommendation: _specifies the Colab runtime proxy token header on fetch requests_

src/jupyter/provider.unit.test.ts
```diff
@@ -186,26 +191,74 @@ describe("ColabJupyterServerProvider", () => {
...
+    colabClientStub.assign.withArgs(nbh, server.variant).resolves(assignment);
+    assert.isDefined(assignment.runtimeProxyInfo);
+    const expectedHeaders = new Headers();
+    expectedHeaders.append(
+      COLAB_RUNTIME_PROXY_TOKEN_HEADER,
+      assignment.runtimeProxyInfo.token,
+    );
+    const expectedInit: fetch.RequestInit = {
+      headers: expectedHeaders,
+    };
```
kevineger: Nit: for readability, consider removing this and simply defining the thin object in the assert below.

```ts
sinon.assert.calledOnceWithExactly(
  fetchStub,
  assignment.runtimeProxyInfo.url,
  {
    headers: {
      "X-Colab-Runtime-Proxy-Token": assignment.runtimeProxyInfo.token,
    },
  },
);
```

src/jupyter/provider.unit.test.ts
wh1ter4bb1t-js: Done.

src/jupyter/provider.unit.test.ts
wh1ter4bb1t-js: since we use `Header()` which returns a [header class](https://fetch.spec.whatwg.org/#headers-class) I did not make this exact change but I did inline it.

src/jupyter/provider.ts
wh1ter4bb1t-js: I prefer an export since it reduces the area required to be changed; but my feelings are not strong enough for me to argue for them. Done.

package.json
wh1ter4bb1t-js: This is a tricky one, I was starting to have issues again with this and had to look back at it. I think I have been out of the npm world too long to understand the nuances of the hat. Looking back at it, it may seem like common sense that there would be issues if we are importing for backwards compatibility to `2.7.0` vs importing for backwards compatibility to `2.6.7` i.e. I think `^2.7.0` may install a different version of `gaxios`. 

Either way I think I would need more time in the environment to create a complete reproducible bug that I could share and even then I think when it comes to adding code that is expected to override the behavior of other packages the onus to make sure the package versions match is on us.

long way of saying I just did have a correct understanding of how `^` and `~` worked

wh1ter4bb1t-js: Scratch everything I said... This seems to just be an issue with `node-fetch`... and one that is made even more confusing that it works randomly on some `npm installs` and not others. I've commented on https://github.com/node-fetch/node-fetch/discussions/1598 to see if this should be filed as an issue, and filed a new [PR](https://github.com/googlecolab/colab-vscode/pull/27) with the workaround.

PR #25, opened 2025-02-20
fix: Add eslint rule for line length

wh1ter4bb1t-js:
Adds an eslint rule for max length of lines. Exceptions to the default,
tried to maintain as true to standard typescript style guide
https://google.github.io/styleguide/tsguide.html.

### DIFF
--- a/src/auth/provider.ts
+++ b/src/auth/provider.ts
@@ -278,7 +278,7 @@ const isStringArray = (value: unknown): value is readonly string[] => {
 };
 
 /**
- * Type guard to check if a value matches the 
+ * Type guard to check if a value matches the
  * AuthenticationSessionAccountInformation shape
  */
 const isAuthSessionAccountInfo = (
--- a/src/colab/api.ts
+++ b/src/colab/api.ts
@@ -1,7 +1,7 @@
 /**
  * @fileoverview API types for interacting with Colab's backend.
  *
- * As mentioned throughout several of the relevant fields, a lot of the name 
+ * As mentioned throughout several of the relevant fields, a lot of the name
  * choices are due to historical reasons and are not ideal.
  */
 


package.json
```diff
@@ -55,6 +55,7 @@
   "devDependencies": {
     "@cspell/eslint-plugin": "^8.17.3",
     "@eslint/js": "^9.15.0",
+    "@stylistic/eslint-plugin-ts": "^4.0.1",
```
kevineger, line 58: Is it worth taking on this dep? I believe we can configure what we want with `max-len` ([docs](https://eslint.org/docs/latest/rules/max-len)).

kevineger: Woops! (As you pointed out over IM) That rule is deprecated in favour of this plugin.

image

eslint.config.mjs
```diff
@@ -27,6 +29,16 @@ export default tseslint.config(
...
         },
       ],
+      "@/max-len": [
+        "error",
+        {
+          ignoreTrailingComments: true,
+          ignoreStrings: true,
+          ignoreTemplateLiterals: true,
+          ignoreUrls: true,
+          ignorePattern: "<.*>",
```
kevineger, line 39: Sorry, why this `ignorePattern`?

.prettierrc
```diff
@@ -1 +1,3 @@
-{}
+{
+  "printWidth": 80
+}
```
kevineger, line 3: To ensure that lint `--fix` is working, consider doing a `git reset` for all `.ts` files and re-running.

src/colab/client.unit.test.ts
```diff
@@ -1,18 +1,18 @@
...
   Accelerator,
-  CCUInfo,
   Assignment,
+  CCUInfo,
+  GetAssignmentResponse,
   Shape,
   SubscriptionState,
   SubscriptionTier,
   Variant,
-  GetAssignmentResponse,
```
kevineger: Was adding the `@stylistic/ts` plugin what highlighted this ordering issue and auto-fixed it? Or was it manual?

I'm all for proper ordering, I'd just prefer automation look after it.

src/jupyter/provider.ts
```diff
@@ -52,15 +52,17 @@ export class ColabJupyterServerProvider
...
 
         return eligibleGpu && !ineligibleGpu;
       });
     });
   }
 
   /**
-   * Resolves the connection for the provided {@link JupyterServer Jupyter Server}.
+   * Resolves the connection for the provided
+   * {@link JupyterServer Jupyter Server}.
```
kevineger, line 71: It's be real nice if Prettier would auto-wrap comments. Here I'd expect the `{@link ...}` tag to be broken up across lines.

eslint.config.mjs
wh1ter4bb1t-js: linter and and formatter argue about what to do when you have long type to the generic.

ie. do you want
```
function something<ThisIsALongTypeNameIDontKnowWhatToSay>()
```
or something like: 
```
function something<
  ThisIsALongTypeNameIDontKnowWhatToSay>()
```

The first choice was the only one that didn't make me cringe

src/colab/client.unit.test.ts
wh1ter4bb1t-js: This was done automatically

.prettierrc
wh1ter4bb1t-js: Done.

src/jupyter/provider.ts
wh1ter4bb1t-js: would be nice, sounds like a follow up to figure a way to do that

PR #26, opened 2025-02-20
feat: Use babel as compiler

wh1ter4bb1t-js:
Adds Babel as the Javascript compiler. There are a lot of benefits to
using babel over standard `tsc` this will allow us to more easily
integrate with other build tools like webpack and automatic polyfilling.

We can also use babel to create custom transformations to better
enforce code guidelines.

Babel transpilation is not as good as Typescripts own compiler meaning
it does not catch Type errors as easily. We can get around this by
running `tsc` in parallel when building so that those errors are still
surfaced without blocking.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
   },
   "scripts": {
     "clean": "rm -rf out/*",
+    "compile": "tsc -p ./",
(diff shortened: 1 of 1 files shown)


kevineger:
https://github.com/googlecolab/colab-vscode/pull/80

PR #27, opened 2025-02-22
Fix: add logic to get around node-fetch symbol bug

wh1ter4bb1t-js:
Node fetch has an interesting bug where symbols do not always match
requests. This seems to be a known issue but not prevalant enough to
warrant an issue, and there is a known workaround in [their
discussions](https://github.com/node-fetch/node-fetch/discussions/1598).

The Jupyter extension seems to get around throwing this typeError by
strategically type asserting to `any` and using their own `RequestInfo`
type.

Using a similar type strategy will likely make it more difficult for us
to debug issues in the future, since we will need to follow the type
pipeline from jupyter extension to the Colab extension back to the
jupyter extension and finally into node-fetch....

On the other hand using the workaround is ugly, but any issues that
occur with the fetch function we can still relatively easily debug
within our own extension.

I don't really know why sometimes running a fresh npm install pinned to
certain versions worked especially since if I left it for a day or so it
would no longer work, this lead to a lot of confusion/frustration on my
end.

Needless to say this fix will allow the fetch function to work properly
on all the `node-fetch` package versions that the Jupyter extension has
marked as a dependency.

### DIFF
--- a/src/jupyter/provider.unit.test.ts
+++ b/src/jupyter/provider.unit.test.ts
@@ -298,7 +298,9 @@ describe("ColabJupyterServerProvider", () => {
         url: "https://mock-url.com",
       },
     };
-    colabClientStub.assign.withArgs(sinon.match(isUUID), server.variant).resolves(assignment);
+    colabClientStub.assign
+      .withArgs(sinon.match(isUUID), server.variant)
+      .resolves(assignment);
     assert.isDefined(assignment.runtimeProxyInfo);
 
     const resolvedServer = await serverProvider.resolveJupyterServer(


kevineger:
I've never hit this issue myself, I don't think we need this change?

Locally, make sure you're running `npm ci` instead of `npm install`.

src/jupyter/provider.ts
```diff
@@ -96,17 +96,24 @@ export class ColabJupyterServerProvider
...
               requestHeaders.append(COLAB_RUNTIME_PROXY_TOKEN_HEADER, token);
               init.headers = requestHeaders;
 
+              // Create a new request with the correct symbols, so that
+              // node-fetch will parse it correctly.
+              // This is a known issue with node-fetch, it does not seem like it
+              // is considered enough of a problem to create an issue on. And
+              // this is the expected workaround:
+              // https://github.com/node-fetch/node-fetch/discussions/1598
+              if (typeof info !== "string" && !("href" in info)) {
```
kevineger: Consider a type-guard: `isRequest(info)`.

Add function at the bottom:

```ts
// A workaround to a known issue with node-fetch.
// https://github.com/node-fetch/node-fetch/discussions/1598
//
// Create a new request with the correct symbols, so that node-fetch will parse
// it correctly.
//
// JupyterHub's similar issue:
// https://github.com/microsoft/vscode-jupyter-hub/issues/53
function isRequest(info: RequestInfo): info is Request {
  return typeof info !== "string" && !("href" in info);
}
```

src/jupyter/provider.unit.test.ts
```diff
@@ -253,5 +253,48 @@ describe("ColabJupyterServerProvider", () => {
...
+    assert.isDefined(assignment.runtimeProxyInfo);
+
+    const resolvedServer = await serverProvider.resolveJupyterServer(
+      server,
+      cancellationToken,
+    );
+
+    assert.isDefined(resolvedServer?.connectionInformation?.fetch);
+    await resolvedServer.connectionInformation.fetch(
+      new Request(assignment.runtimeProxyInfo.url, { method: "GET" }),
```
kevineger: Sorry if I'm missing something but how are we verifying that it's using the new request?

Should the test assertion be that the `fetchStub` was not called with this object (equality vs. deep equality)?

Should there be tests for when `info` is a `string` or `URLLike`?

wh1ter4bb1t-js: We test strings in the test above this one, 

it might be hard to tell, but we create a fetch Request object and pass that through the fetch method here, and then below we assert that it was called with a node-fetch `Request` object. we could test on the object, but then the test is not testing to ensure that `nodefetch.default` is being called with the `nodefetch.Request`, just it's not being called with this Request object.

src/jupyter/provider.ts
wh1ter4bb1t-js: Done.

src/jupyter/provider.unit.test.ts
kevineger: I see the string test above now, thanks.

But still, shouldn't we be checking/asserting that the `fetchStub` was `calledOnceWithExactly` a (node) `Request` that is deep equal to what you have, but not equal (refer to the same object)?

src/jupyter/provider.ts
```diff
@@ -128,10 +128,32 @@ export class ColabJupyterServerProvider
...
+
+// A workaround to a known issue with node-fetch.
+// https://github.com/node-fetch/node-fetch/discussions/1598
+//
+// Create a new node-fetch request with the correct symbols, so that
+// node-fetch will parse it correctly.
+//
+// Parsed urls turn into [Request objects] here:
+// https://github.com/node-fetch/node-fetch/blob/8b3320d2a7c07bce4afc6b2bf6c3bbddda85b01f/src/request.js#L52
+function isRequest(info: RequestInfo): info is Request {
```
kevineger: Nit: can we de-dup the rationale?

image

wh1ter4bb1t-js: Done.

PR #28, opened 2025-02-22
Chore: Add client agent headers to fetch requests

wh1ter4bb1t-js:
Add the Colab client agent headers to headers being sent with the custom
fetch method. This will allow Colab to identify when VS Code is the
client.

### DIFF
--- a/src/jupyter/provider.ts
+++ b/src/jupyter/provider.ts
@@ -20,7 +20,7 @@ const COLAB_RUNTIME_PROXY_TOKEN_HEADER = "X-Colab-Runtime-Proxy-Token";
 const COLAB_CLIENT_AGENT_HEADER = "X-Colab-Client-Agent";
 
 /**
- * Const variable for the client agent value from VS Code.
+ * The client agent value for requests originating from VS Code.
  */
 const VSCODE_CLIENT_AGENT = "vscode";
 


src/jupyter/provider.unit.test.ts
```diff
@@ -214,7 +214,7 @@ describe("ColabJupyterServerProvider", () => {
     });
   });
 
-  it("specifies the Colab runtime proxy token header on fetch requests", async () => {
+  it("specifies the Colab runtime proxy token and client agent header on fetch requests", async () => {
```
kevineger: Nit: _"specifies the Colab headers on fetch requests"_

src/jupyter/provider.ts
```diff
@@ -10,10 +10,22 @@ import { ColabClient } from "../colab/client";
...
+ * COLAB_CLIENT_AGENT_HEADER is a header key for the Colab client agent.
+ */
+const COLAB_CLIENT_AGENT_HEADER = "X-Colab-Client-Agent";
+
+/**
+ * CLIENT_AGENT is an enum of possible client agents that the Colab client agent header will take.
+ */
+enum CLIENT_AGENT {
+  VSCODE = "vscode",
+}
```
kevineger: We aren't (and likely never will) set anything but `"vscode"`, no need for this enum. Switch to a constant?

src/jupyter/provider.ts
```diff
@@ -10,10 +10,22 @@ import { ColabClient } from "../colab/client";
...
 /**
- * Header key for the runtime proxy token.
+ * COLAB_RUNTIME_PROXY_TOKEN_HEADER is a header key for the Colab runtime proxy token.
  */
 const COLAB_RUNTIME_PROXY_TOKEN_HEADER = "X-Colab-Runtime-Proxy-Token";
 
+/**
+ * COLAB_CLIENT_AGENT_HEADER is a header key for the Colab client agent.
+ */
+const COLAB_CLIENT_AGENT_HEADER = "X-Colab-Client-Agent";
```
kevineger:
```suggestion
/**
 * A header key for the Colab runtime proxy token.
 */
const COLAB_RUNTIME_PROXY_TOKEN_HEADER = "X-Colab-Runtime-Proxy-Token";

/**
 * A header key for the Colab client agent.
 */
const COLAB_CLIENT_AGENT_HEADER = "X-Colab-Client-Agent";
```

src/jupyter/provider.unit.test.ts
wh1ter4bb1t-js: Done.

src/jupyter/provider.ts
wh1ter4bb1t-js: Done.

src/jupyter/provider.ts
wh1ter4bb1t-js: Done.

src/jupyter/provider.ts
```diff
@@ -10,10 +10,20 @@ import { ColabClient } from "../colab/client";
...
 const COLAB_RUNTIME_PROXY_TOKEN_HEADER = "X-Colab-Runtime-Proxy-Token";
 
+/**
+ * A header key for the Colab client agent.
+ */
+const COLAB_CLIENT_AGENT_HEADER = "X-Colab-Client-Agent";
+
+/**
+ * Const variable for the client agent value from VS Code.
+ */
```
kevineger: Nit: `Const variable` is implied by the fact it's a `const` variable 😉

```suggestion
/**
 * The client agent value for requests originating from VS Code.
 */
```

PR #31, opened 2025-02-25
chore: improve how the VS Code stub helper is created and used

kevineger:
Sinon stubs shouldn't be re-used like we currently are. This change
makes it more ergnomic to create the VS Code stub and interact with it.

Stubs can now be used directly on the `VsCodeStub`, e.g. `vsCodeStub.authentication.registerAuthenticationProvider`. This requires the addition of `.asVsCode()` which I feel is a welcomed tradeoff.

- Leveraged `sinon.SinonStubbedMember` instead of explicitly typing out the `SinonStub`.
- Simplifying how `Disposable` is stubbed, I removed the uneccessary `Disposable.from` in `src/auth/provider.ts` and `src/jupyter/provider.ts`.
- Lift the extension publisher into a constant to avoid retyping it
  throughout the project.
- Standardize on calling `sinon.restore()`

### DIFF
--- a/src/auth/provider.unit.test.ts
+++ b/src/auth/provider.unit.test.ts
@@ -157,9 +157,9 @@ describe("GoogleAuthProvider", () => {
           }),
           sinon.match.any,
         )
-        .callsFake((_, task) => {
-          return task({ report: sinon.stub() }, cancellationStub);
-        });
+        .callsFake((_, task) =>
+          task({ report: sinon.stub() }, cancellationStub),
+        );
       redirectUriHandlerStub.waitForCode.throws(new Error("Barf"));
 
       await expect(authProvider.createSession(REQUIRED_SCOPES)).to.be.rejected;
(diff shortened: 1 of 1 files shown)


src/auth/provider.unit.test.ts
```diff
@@ -151,35 +148,24 @@ describe("GoogleAuthProvider", () => {
...
-                increment?: number;
-              }>,
-              token: vscode.CancellationToken,
-            ) => Thenable<string>,
-          ) => {
-            return task({ report: sinon.stub() }, cancellationStub);
-          },
-        );
+        .callsFake((_, task) => {
+          return task({ report: sinon.stub() }, cancellationStub);
```
zelliott: nit: Consider shorthand if it doesn't format horribly:

```
(_, task) => task(...);
```

Apply throughout the CL.

kevineger: Done.

PR #32, opened 2025-02-25
feat: add support for storing the Colab Jupyter servers

kevineger:
On extension re-activation, VS Code reload, etc., we need to be able to populate the list of previously connected servers. While we could query for assignments every time, it's unecessary and slow. This storage class abstracts the complexity which will be leveraged by the re-worked Jupyter server provider in subsequent PRs.

Here I'm leveraging a new dep, `zod`. It's great! I plan to move existing parsing logic to it, instead of the type guards which are nasty to write and maintain.

### DIFF
--- a/src/jupyter/storage.ts
+++ b/src/jupyter/storage.ts
@@ -97,7 +97,8 @@ export class ServerStorage {
    * Remove an assigned server.
    *
    * @param serverId The ID of the server to remove.
-   * @returns true if a server was stored and has been removed, or false if the server does not exist.
+   * @returns true if a server was stored and has been removed, or false if the
+   * server does not exist.
    */
   async remove(serverId: string): Promise<boolean> {
     const existingServersJson = await this.secrets.get(ASSIGNED_SERVERS_KEY);


src/jupyter/storage.ts
```diff
@@ -0,0 +1,144 @@
...
+export class ServerStorage {
+  private cache?: ColabAssignedServer[];
+
+  constructor(
+    private readonly vs: typeof vscode,
+    private readonly secrets: vscode.SecretStorage,
+  ) {}
+
+  /**
+   *
```
zelliott: Missing a first-line comment?

src/jupyter/storage.ts
```diff
@@ -0,0 +1,144 @@
...
+   * @returns The assigned servers that have been stored.
+   */
+  async get(): Promise<ColabAssignedServer[]> {
+    if (this.cache !== undefined) {
+      return this.cache;
+    }
+    const serversJson = await this.secrets.get(ASSIGNED_SERVERS_KEY);
+    const servers = serversJson
+      ? AssignedServers.parse(JSON.parse(serversJson))
```
zelliott, line 54: Oh that is nice.

src/jupyter/storage.ts
```diff
@@ -0,0 +1,144 @@
...
+      : [];
+    const res = servers.map((server) => {
+      return {
```
zelliott: Consider shorthand.

src/jupyter/storage.ts
```diff
@@ -0,0 +1,144 @@
...
+   * Store an assigned server.
+   *
+   * @param server The server to store.
+   */
+  async store(server: ColabAssignedServer): Promise<void> {
+    const existingServersJson = await this.secrets.get(ASSIGNED_SERVERS_KEY);
+    const existingServers = existingServersJson
+      ? AssignedServers.parse(JSON.parse(existingServersJson))
+      : [];
+    const serversById = new Map(existingServers.map((s) => [s.id, s]));
```
zelliott: nit: Opportunity to pull this into a helper used by both `store` and `remove`.

In general, it feels like there may be some opportunities to DRY up this class a bit, a lot of the methods do similar things. But not blocking because it's all straightforward code.

src/jupyter/storage.ts
```diff
@@ -0,0 +1,144 @@
...
+  };
+};
+
+/**
+ * Server storage for Colab Jupyter servers.
+ *
+ * Implementation assumes full ownership over the backing secret storage file.
+ */
+export class ServerStorage {
+  private cache?: ColabAssignedServer[];
```
zelliott, line 37: Not blocking - I wonder if this should be two separate classes:

* `ServerStorage` - Basic CRUD operations with the secret storage file. No caching.
* `Cache` - Generic cache class that can cache any request. I wouldn't be surprised if you find yourself needing a generic `Cache` class in the future. Bunch of examples in g3 (search for cache.ts). Could support TTLs, eviction criteria, etc.

Up to you, could always be done in the future.

src/jupyter/storage.ts
```diff
@@ -0,0 +1,144 @@
...
+      Array.from(serversById.values()),
+      existingServersJson,
+    );
+  }
+
+  /**
+   * Remove an assigned server.
+   *
+   * @param serverId The ID of the server to remove.
+   * @returns true if a server was stored and has been removed, or false if the server does not exist.
```
zelliott: This was a little surprising to me - do remove/delete APIs typically return `false` in this case? If the server does not exist, this could be seen as a vacuous success. I'm trying to imagine why a client would potentially care whether the server was actually removed or not... I think they'd likely just care that the server is no longer there.

src/jupyter/storage.ts
```diff
@@ -0,0 +1,144 @@
+import { JupyterServerConnectionInformation } from "@vscode/jupyter-extension";
```
zelliott, line 1: Should this module be called `server_storage.ts` instead of just `storage.ts` to avoid ambiguity if other storage modules are introduced?

kevineger: I'm explicitly exporting members so isn't it a bit redundant considering the full path?

src/jupyter/storage.ts
kevineger: Fair enough. In hindsight I should have skipped caching entirely (for now). Considering it's already done, I'm tempted to leave it in. Can always lift it out easily.

Promoting to a generic cache in the future seems ideal. I also think it'd be a reasonable lift to make `ServerStorage` more generic as simply _VS Code secret storage CRUD_ (once we have another use case).

I'll loop back on this if that's alright.

src/jupyter/storage.ts
kevineger: Done.

src/jupyter/storage.ts
kevineger: 😎

src/jupyter/storage.ts
kevineger: Done.

src/jupyter/storage.ts
kevineger: > do remove/delete APIs typically return false in this case?

Consider a [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map). It's [`delete`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete) `@returns`:

```
@returns — true if an element in the Map existed and has been removed, or false if the element does not exist.
```

I made it `Promise<boolean>` since I've always followed that convention for `remove`/`delete` methods. While I haven't needed it yet, I suspect we may at a minimum want to log when we try to `remove` a server which doesn't exist.

src/jupyter/storage.ts
kevineger: Done.

W.r.t. removing other duplication, anything specific you'd like me to address? While `store` and `get` look similar, they're going the opposite ways in the conversion. Personally not seeing a whole lot else that could be minimized, but I'm all ears!

PR #36, opened 2025-02-28
feat: multi-step quick-pick

kevineger:
Influenced by the [VS Code sample](https://github.com/microsoft/vscode-extension-samples/blob/7e4846a7a78428c52e8fe7f7aae16aa444f8d8cb/quickinput-sample/src/multiStepInput.ts) but adapted for our usage and with simpler ergonomics. Oh... and with tests 😉.

Why use `expect(selected).to.equal(undefined);` over `expect(selected).to.be.undefined;`? The latter violates `@typescript-eslint/no-unused-expressions` and rather than disable it inline (or for all tests) I'd rather tolerate the minorly less readable former version.

### DIFF
--- a/src/common/multi-step-quickpick.ts
+++ b/src/common/multi-step-quickpick.ts
@@ -11,12 +11,17 @@ import vscode from "vscode";
  * Represents an action that can be taken during an input flow.
  */
 export class InputFlowAction extends Error {
+  /** Navigate back in the input flow. */
   static back = new InputFlowAction("back");
+  /** Cancel the input flow. */
   static cancel = new InputFlowAction("cancel");
 }
 
 /**
  * Represents a chainable step in the input flow.
+ *
+ * An {@link InputStep} can be returned to chain the next step in the input
+ * flow. If undefined is returned, the input flow is complete.
  */
 export type InputStep = (
   input: MultiStepInput,
@@ -24,12 +29,16 @@ export type InputStep = (
 
 /**
  * The base options for all quick-input types.
+ *
+ * @param title
  */
 export interface QuickInputOptions {
+  /** The title of the input. */
   title: string;
+  /** The current step count. */
   step: number;
+  /** The total step count. */
   totalSteps: number;
-  ignoreFocusOut?: boolean;
 }
 
 /**
@@ -37,8 +46,11 @@ export interface QuickInputOptions {
  */
 export interface QuickPickOptions<T extends QuickPickItem>
   extends QuickInputOptions {
+  /** Items to pick from. */
   items: T[];
+  /** The optional placeholder text shown when no value has been input. */
   placeholder?: string;
+  /** The actively picked item. */
   activeItem?: T;
 }
 
@@ -46,9 +58,15 @@ export interface QuickPickOptions<T extends QuickPickItem>
  * The options for an input box input.
  */
 export interface InputBoxOptions extends QuickInputOptions {
+  /** The current value of the input box. */
   value: string;
+  /** The prompt text providing some ask or explanation to the user. */
   prompt: string;
+  /** A function that validates the input value. If a string is returned, it is
+   * set as the validation message indicating a problem with the current input
+   * value. If undefined is returned, the validation message is cleared. */
   validate: (value: string) => string | undefined;
+  /** The optional placeholder text shown when no value has been input. */
   placeholder?: string;
 }
 
@@ -58,7 +76,7 @@ export interface InputBoxOptions extends QuickInputOptions {
 export class MultiStepInput {
   private constructor(private readonly vs: typeof vscode) {}
 
-  private steps: InputStep[] = [];
+  private readonly steps: InputStep[] = [];
   private current?: QuickInput;
 
   /**
@@ -128,7 +146,6 @@ export class MultiStepInput {
         input.title = opts.title;
         input.step = opts.step;
         input.totalSteps = opts.totalSteps;
-        input.ignoreFocusOut = opts.ignoreFocusOut ?? false;
         input.placeholder = opts.placeholder;
         input.items = opts.items;
         if (opts.activeItem) {
@@ -137,23 +154,23 @@ export class MultiStepInput {
         input.buttons = [this.vs.QuickInputButtons.Back];
 
         const nav = this.configureNavigation(input, reject);
-        disposables.push(nav.onDidHide, nav.onDidTriggerButton);
 
         disposables.push(
+          nav.onDidHide,
+          nav.onDidTriggerButton,
           input.onDidChangeSelection((selectedItems) => {
             resolve(selectedItems[0]);
           }),
         );
 
-        if (this.current) {
-          this.current.dispose();
-        }
+        this.current?.dispose();
         this.current = input;
         this.current.show();
       });
     } finally {
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-      disposables.forEach((d) => d.dispose());
+      for (const disposable of disposables) {
+        disposable.dispose();
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/common/multi-step-quickpick.ts
```diff
@@ -0,0 +1,233 @@
...
+  QuickInput,
+  Event,
+  QuickInputButton,
+} from "vscode";
+import vscode from "vscode";
+
+/**
+ * Represents an action that can be taken during an input flow.
+ */
+export class InputFlowAction extends Error {
```
zelliott, line 13: Does this extend `Error` because you're throwing it, and there's some linter complaining about throwing non-errors?

If that's the case, I'd prefer to silence the linter instead of having this extend `Error`, because the latter doesn't make sense to me.

src/common/multi-step-quickpick.ts
```diff
@@ -0,0 +1,233 @@
...
+) => Thenable<InputStep | undefined>;
+
+/**
+ * The base options for all quick-input types.
+ */
+export interface QuickInputOptions {
+  title: string;
+  step: number;
+  totalSteps: number;
+  ignoreFocusOut?: boolean;
```
zelliott: What does `ignoreFocusOut` do? Do you mind adding a JSDoc comment?

src/common/multi-step-quickpick.ts
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/**
+ * The options for a quick pick input.
+ */
+export interface QuickPickOptions<T extends QuickPickItem>
+  extends QuickInputOptions {
+  items: T[];
+  placeholder?: string;
+  activeItem?: T;
```
zelliott, line 54: Can this be `activeIndex`? If the answer is "yes but why does it matter", fine to keep as is, not a huge preference.

src/common/multi-step-quickpick.ts
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/**
+ * The options for an input box input.
+ */
+export interface InputBoxOptions extends QuickInputOptions {
+  value: string;
+  prompt: string;
```
zelliott, line 64: What does `prompt` do? Do you mind adding a JSDoc comment?

src/common/multi-step-quickpick.ts
```diff
@@ -0,0 +1,233 @@
...
+  validate: (value: string) => string | undefined;
```
zelliott, line 68: Does `validate` return an user-facing error string on failure, and `undefined` on success? Do you mind adding a JSDoc comment explaining this?

src/common/multi-step-quickpick.ts
```diff
@@ -0,0 +1,233 @@
...
+  placeholder?: string;
+}
+
+/**
+ * A chainable multi-step runner for quick-inputs.
+ */
+export class MultiStepInput {
+  private constructor(private readonly vs: typeof vscode) {}
+
+  private steps: InputStep[] = [];
```
zelliott: Make `readonly` if possible.

src/common/multi-step-quickpick.ts
```diff
@@ -0,0 +1,233 @@
...
+        input.buttons = [this.vs.QuickInputButtons.Back];
+
+        const nav = this.configureNavigation(input, reject);
+        disposables.push(nav.onDidHide, nav.onDidTriggerButton);
+
+        disposables.push(
+          input.onDidChangeSelection((selectedItems) => {
+            resolve(selectedItems[0]);
+          }),
+        );
```
zelliott: Could move this into the `push` above if you wanted. But I understand if you think separating improves readability.

src/common/multi-step-quickpick.ts
```diff
@@ -0,0 +1,233 @@
...
+
+        if (this.current) {
```
zelliott: Could do `this.current?.dispose()`

src/common/multi-step-quickpick.ts
```diff
@@ -0,0 +1,233 @@
...
+          this.current.dispose();
+        }
+        this.current = input;
+        this.current.show();
+      });
+    } finally {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+      disposables.forEach((d) => d.dispose());
```
zelliott: nit: In g3, `for..of` iteration is preferred: go/tsjs-practices/iteration#forof.

src/common/multi-step-quickpick.ts
```diff
@@ -0,0 +1,233 @@
...
+    }
+  }
+
+  /**
+   * Creates and shows an input box.
+   *
+   * @param opts The options for the input box.
+   * @returns The entered value.
+   */
+  async showInputBox(opts: InputBoxOptions): Promise<string> {
```
zelliott, line 183: Left some comments above that also apply to this method.

src/common/multi-step-quickpick.ts
```diff
@@ -0,0 +1,233 @@
...
+        const nav = this.configureNavigation(input, reject);
+        disposables.push(nav.onDidHide, nav.onDidTriggerButton);
+
+        // Handle value confirmation.
+        disposables.push(
+          input.onDidAccept(() => {
+            const value = input.value;
+            input.enabled = false;
+            input.busy = true;
+            if (!opts.validate(value)) {
```
zelliott: I did a double-take when reading this (i.e. _"if not validate, then resolve??"_). But then I remembered that `validate` returns a truthy value when it _fails_ and a falsey value when it succeeds. Seems like a bit of a foot gun.

`validate` could return a `ValidationResult`:

```
type ValidationResult =
  | { isValid: true }
  | { isValid: false; error: string };
```

But I understand if that's a bit too verbose and you'd prefer to keep things as-is.

src/common/multi-step-quickpick.unit.test.ts
```diff
@@ -0,0 +1,435 @@
...
+
+  describe("run", () => {
+    function inputStepStub(): sinon.SinonStub<
+      [input: MultiStepInput],
+      Thenable<InputStep | undefined>
+    > {
+      return sinon.stub();
+    }
+
+    it("a single step", async () => {
```
zelliott: nit: English sentences - `it("runs a single step"`.

src/common/multi-step-quickpick.unit.test.ts
```diff
@@ -0,0 +1,435 @@
...
+      });
+      vsCodeStub.window.createInputBox.returns(
+        inputBoxStub as Partial<InputBox> as InputBox,
+      );
+    });
+
+    it('rejects on "back"', async () => {
+      const step: InputStep = async (input: MultiStepInput) => {
+        await input.showInputBox(ENTER_A_VALUE);
+        return undefined;
```
zelliott, line 294: Shorthand `return;` should work? Apply throughout this file.

src/common/multi-step-quickpick.ts
kevineger: Experimented and discussed in person the tradeoffs of making `InputFlowAction` a simple enum.

Throwing a non-error forces Sinon stub setups (e.g., `.throws(InputFlowAction.back);`) to use `.callsFake` with a `Promise.reject(InputFlowAction.BACK)` and include a corresponding lint disable for `@typescript-eslint/prefer-promise-reject-errors`.

It also means we can't use Chai's `rejectedWith` and instead need to `.rejected.then((err) => (err).to.be.equal(InputFlowAction.BACK));`.

If the logic didn't use function chaining to navigate through the multi-step input flow, it'd make sense to not throw and instead return these `InputFlowAction`-s. But that change would be a rewrite and likely require a lot of duplication to users for handling the input flow (back and cancel).

Leaving as is.

src/common/multi-step-quickpick.ts
kevineger: If the UI should stay open even when loosing UI focus. Defaults to false.

I thought I needed it, but I didn't end up using it up the PR chain. Removed.

src/common/multi-step-quickpick.ts
kevineger: It's not an index. It's the actual item, `T`. See usage in `showQuickPick` below.

src/common/multi-step-quickpick.ts
kevineger: Added full doc.

src/common/multi-step-quickpick.ts
kevineger: Added full doc.

src/common/multi-step-quickpick.ts
kevineger: Done.

src/common/multi-step-quickpick.ts
kevineger: Done.

src/common/multi-step-quickpick.ts
kevineger: Done.

src/common/multi-step-quickpick.ts
kevineger: Done.

src/common/multi-step-quickpick.ts
kevineger: Done.

src/common/multi-step-quickpick.ts
kevineger: Extracted to a variable which should make the truthy value more obvious (`const validationMessage = opts.validate(value);`).

src/common/multi-step-quickpick.unit.test.ts
kevineger: Done. Applied globally.

src/common/multi-step-quickpick.unit.test.ts
kevineger: 'Promise<void>' is not assignable to type 'Thenable<InputStep | undefined>'

src/common/multi-step-quickpick.ts
zelliott: Ah - to clarify - I was asking _could_ it be just the index instead (aka `activeIndex?: number`). Then you could get the item via `items[activeIndex]`. But no difference really.

PR #37, opened 2025-02-28
feat: Colab server picker

kevineger:
This will be used in a subsquent PR for interactively selecting the server to create.

While note required for this change, I did remove `id` from `ColabServerDescriptor` and instead inlined it to the type for the `COLAB_SERVERS` definition. It's going to be fully removed in the next PR (provider rework).

### DIFF
--- a/src/colab/server-picker.ts
+++ b/src/colab/server-picker.ts
@@ -3,6 +3,9 @@ import { InputStep, MultiStepInput } from "../common/multi-step-quickpick";
 import { ColabServerDescriptor } from "../jupyter/servers";
 import { Accelerator, Variant } from "./api";
 
+/**
+ * Supports prompting the user to pick a Colab server to be created.
+ */
 export class ServerPicker {
   constructor(private readonly vs: typeof vscode) {}
 
@@ -10,7 +13,6 @@ export class ServerPicker {
    * Prompt the user through a multi-step series of inputs to pick a Colab
    * server type.
    *
-   * @param vs The vscode module.
    * @param availableServers The available servers to pick from.
    * @returns The selected server, or undefined if the user cancels.
    */
@@ -25,7 +27,7 @@ export class ServerPicker {
       variantToAccelerators.set(server.variant, accelerators);
     }
     if (variantToAccelerators.size === 0) {
-      return undefined;
+      return;
     }
 
     const state: Partial<Server> = {};
@@ -56,18 +58,18 @@ interface Server {
 /**
  * A partial of {@link Server} with all properties optional except for K.
  */
-type ServerWith<K extends keyof Server> = Partial<Server> &
+type PartialServerWith<K extends keyof Server> = Partial<Server> &
   Required<Pick<Server, K>>;
 
 function isVariantDefined(
   state: Partial<Server>,
-): state is ServerWith<"variant"> {
+): state is PartialServerWith<"variant"> {
   return state.variant !== undefined;
 }
 
 function isAcceleratorDefined(
   state: Partial<Server>,
-): state is ServerWith<"accelerator"> {
+): state is PartialServerWith<"accelerator"> {
   return state.accelerator !== undefined;
 }
 
@@ -125,7 +127,7 @@ async function promptForVariant(
 
 async function promptForAccelerator(
   input: MultiStepInput,
-  state: ServerWith<"variant">,
+  state: PartialServerWith<"variant">,
   acceleratorsByVariant: Map<Variant, Set<Accelerator>>,
 ): Promise<InputStep | undefined> {
   const accelerators = acceleratorsByVariant.get(state.variant) ?? new Set();
@@ -154,7 +156,7 @@ async function promptForAccelerator(
 
 async function promptForAlias(
   input: MultiStepInput,
-  state: ServerWith<"variant">,
+  state: PartialServerWith<"variant">,
 ): Promise<InputStep | undefined> {
   const acceleratorPart =
     state.accelerator && state.accelerator !== Accelerator.NONE
@@ -162,9 +164,7 @@ async function promptForAlias(
       : "";
   const placeholder = `Colab ${variantToString(state.variant)}${acceleratorPart}`;
   const step =
-    state.accelerator !== undefined && state.accelerator !== Accelerator.NONE
-      ? 3
-      : 2;
+    state.accelerator && state.accelerator !== Accelerator.NONE ? 3 : 2;
   const alias = await input.showInputBox({
     title: "Alias your server",
     step,
@@ -173,14 +173,10 @@ async function promptForAlias(
     // CPU (1), Colab GPU A100 (2).
     value: state.alias ?? "",
     prompt: "Provide a local convenience alias to the server.",
-    validate: (value) => {
-      if (value.length > 10) {
-        return "Name must be less than 10 characters.";
-      }
-      return undefined;
-    },
+    validate: (value) =>
+      value.length > 10 ? "Name must be less than 10 characters." : "",
     placeholder,
   });
-  state.alias = alias ? alias : placeholder;
+  state.alias = alias || placeholder;
   return;
 }
--- a/src/colab/server-picker.unit.test.ts
+++ b/src/colab/server-picker.unit.test.ts
@@ -44,11 +44,11 @@ describe("ServerPicker", () => {
       return stub;
     }
 
-    it("with no available servers returns undefined", () => {
+    it("returns undefined when there are no available servers", () => {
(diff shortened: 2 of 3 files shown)


src/colab/server-picker.ts
```diff
@@ -0,0 +1,186 @@
+import vscode, { QuickPickItem } from "vscode";
+import { InputStep, MultiStepInput } from "../common/multi-step-quickpick";
+import { ColabServerDescriptor } from "../jupyter/servers";
+import { Accelerator, Variant } from "./api";
+
+export class ServerPicker {
```
zelliott, line 9: nit: Do you mind adding a JSDoc comment to this class?

src/colab/server-picker.ts
```diff
@@ -0,0 +1,186 @@
...
+  constructor(private readonly vs: typeof vscode) {}
+
+  /**
+   * Prompt the user through a multi-step series of inputs to pick a Colab
+   * server type.
+   *
+   * @param vs The vscode module.
```
zelliott: There is no `vs` param.

src/colab/server-picker.ts
```diff
@@ -0,0 +1,186 @@
...
+    const variantToAccelerators = new Map<Variant, Set<Accelerator>>();
+    for (const server of availableServers) {
+      const accelerators =
+        variantToAccelerators.get(server.variant) ?? new Set();
+      accelerators.add(server.accelerator ?? Accelerator.NONE);
+      variantToAccelerators.set(server.variant, accelerators);
+    }
+    if (variantToAccelerators.size === 0) {
+      return undefined;
+    }
```
zelliott: Could use shorthand `return;` (optionally one line) if you wanted here (and further down).

src/colab/server-picker.ts
```diff
@@ -0,0 +1,186 @@
...
+interface Server {
+  variant: Variant;
+  accelerator: Accelerator;
+  alias: string;
+}
+
+/**
+ * A partial of {@link Server} with all properties optional except for K.
+ */
+type ServerWith<K extends keyof Server> = Partial<Server> &
```
zelliott: Given what this type alias does, I would maybe expect it to be named `PartialServerWith`.

src/colab/server-picker.ts
```diff
@@ -0,0 +1,186 @@
...
+): Promise<InputStep | undefined> {
+  const items: VariantPick[] = [];
+  for (const variant of acceleratorsByVariant.keys()) {
+    items.push({
+      value: variant,
+      label: variantToString(variant),
+      // TODO: Add a description for each variant?
+    });
+  }
+  const pick = await input.showQuickPick({
```
zelliott, line 108: I could probably answer this myself by looking at one of your other PRs, but I'm assuming `showQuickPick` handles empty `items` correctly? I'm just double-checking that we don't run into a scenario where `showQuickPick` never resolves if `items` is empty.

src/colab/server-picker.ts
```diff
@@ -0,0 +1,186 @@
...
+  input: MultiStepInput,
+  state: ServerWith<"variant">,
+): Promise<InputStep | undefined> {
+  const acceleratorPart =
+    state.accelerator && state.accelerator !== Accelerator.NONE
+      ? ` ${state.accelerator}`
+      : "";
+  const placeholder = `Colab ${variantToString(state.variant)}${acceleratorPart}`;
+  const step =
+    state.accelerator !== undefined && state.accelerator !== Accelerator.NONE
```
zelliott: Above you check `state.accelerator`, here you check if it's explicitly not `undefined`. Choose one?

src/colab/server-picker.ts
```diff
@@ -0,0 +1,186 @@
...
+    totalSteps: step,
+    // TODO: Incrementally number the same variant/server machines. E.g. Colab
+    // CPU (1), Colab GPU A100 (2).
+    value: state.alias ?? "",
+    prompt: "Provide a local convenience alias to the server.",
+    validate: (value) => {
+      if (value.length > 10) {
+        return "Name must be less than 10 characters.";
+      }
+      return undefined;
```
zelliott: nit: Shorthand `return;` Also consider ternary.

src/colab/server-picker.ts
```diff
@@ -0,0 +1,186 @@
...
+    },
+    placeholder,
+  });
+  state.alias = alias ? alias : placeholder;
```
zelliott: `state.alias = alias || placeholder` (or `??` if that fits your use case better).

src/colab/server-picker.unit.test.ts
```diff
@@ -0,0 +1,295 @@
...
+
+    function stubInputBoxForCall(n: number) {
+      const stub = buildInputBoxStub();
+      vsCodeStub.window.createInputBox
+        .onCall(n)
+        .returns(stub as Partial<InputBox> as InputBox);
+      return stub;
+    }
+
+    it("with no available servers returns undefined", () => {
```
zelliott: Same ole "english sentence" comment: `it("returns undefined with there are no available servers`.

That is, match the naming structure of your `"sets the right step when accelerators are available"` test.

src/jupyter/servers.ts
```diff
@@ -40,7 +38,11 @@ export type ColabAssignedServer = ColabJupyterServer & {
 /**
  * The mapping of all potentially available ID to Colab Jupyter servers.
  */
-export const COLAB_SERVERS = new Set<ColabServerDescriptor>([
+export const COLAB_SERVERS = new Set<
+  ColabServerDescriptor & /*Note: id will be removed as part of the Jupyter server provider rework.*/ {
```
zelliott: Yuck this comment is ugly as well.

Does it format nicer and allow for `//` syntax if you put it immediately above the `readonly id: string` line?

src/colab/server-picker.ts
```diff
@@ -0,0 +1,186 @@
...
+      accelerators.add(server.accelerator ?? Accelerator.NONE);
+      variantToAccelerators.set(server.variant, accelerators);
+    }
+    if (variantToAccelerators.size === 0) {
+      return undefined;
+    }
+
+    const state: Partial<Server> = {};
+    await MultiStepInput.run(this.vs, (input) =>
+      promptForVariant(input, state, variantToAccelerators),
```
zelliott, line 35: Non-blocking: I'll admit found how we're building up the `state` partial a little hard to grok. I think I expected this thenable to eventually return the final state, instead of mutating `state` itself. But you linked to the [VS Code prior art](https://github.com/microsoft/vscode-extension-samples/blob/7e4846a7a78428c52e8fe7f7aae16aa444f8d8cb/quickinput-sample/src/multiStepInput.ts), so let's run with it.

src/colab/server-picker.ts
kevineger: Fixed. Will look into adding https://www.npmjs.com/package/eslint-plugin-jsdoc in a follow-up PR.

src/colab/server-picker.ts
kevineger: Done.

src/colab/server-picker.ts
kevineger: Open to follow-up improvements if you want to brainstorm. I think going with the function chaining approach kind of locks us into building up the state, or at least I couldn't find a convenient way to generically type the return as we build it up. I tried to make it clear what the state "has to" be at each phase with the `ServerWith` type (e.g. `ServerWith<"variant">`).

src/colab/server-picker.ts
kevineger: Done.

src/colab/server-picker.ts
kevineger: Done.

src/colab/server-picker.ts
kevineger: Can always focus out or go `BACK`.

src/colab/server-picker.ts
kevineger: Done.

src/colab/server-picker.ts
kevineger: Done.

src/colab/server-picker.ts
kevineger: `||` since `??` expects LHS to be null or undefined, `alias` is a string.

src/colab/server-picker.unit.test.ts
kevineger: Done. _when* there are_ 😉.

Applied globally.

src/jupyter/servers.ts
kevineger: Done.

PR #41, opened 2025-03-04
Fix: Machine polling implementation

wh1ter4bb1t-js:
fix: add ccu-info implementation to jupyter provider

Adds the CCUInformation implementation to the jupyter provider.
This was originally in https://github.com/googlecolab/colab-vscode/pull/24
but moved out into this seperate PR to make the PR smaller and
easier to understand.

### DIFF
--- a/src/colab/api.ts
+++ b/src/colab/api.ts
@@ -5,58 +5,7 @@
  * choices are due to historical reasons and are not ideal.
  */
 
-export interface FreeCcuQuotaInfo {
-  /**
-   * Number of tokens remaining in the "USAGE-mCCUs" quota group (remaining
-   * free usage allowance in milli-CCUs).
-   */
-  remainingTokens: number;
-  /**
-   * Next free quota refill timestamp (epoch) in seconds.
-   */
-  nextRefillTimestampSec: number;
-}
-
-/**
- * Cloud compute unit (CCU) information.
- */
-export interface CcuInfo {
-  /**
-   * The current balance of the paid CCUs.
-   *
-   * Naming is unfortunate due to historical reasons and free CCU quota
-   * balance is made available in a separate field for the same reasons.
-   */
-  currentBalance: number;
-  /**
-   * The current rate of consumption of the user's CCUs (paid or free) based on
-   * all assigned VMs. VMs consume paid CCUs if the user's paid CCU balance is
-   * positive and free CCU quota if the paid balance is zero.
-   */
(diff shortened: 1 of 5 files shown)


src/jupyter/assignments.ts
```diff
@@ -8,6 +8,7 @@ import fetch, {
 } from "node-fetch";
 import vscode from "vscode";
 import { RuntimeProxyInfo, Variant } from "../colab/api";
+import { CcuInformation as CcuInfo } from "../colab/ccu-info";
```
kevineger, line 11: Nit: don't rename. Internal guidance - go/tsjs-style#renaming-imports.

src/jupyter/assignments.ts
```diff
@@ -39,6 +40,8 @@ export class AssignmentManager implements vscode.Disposable {
   onDidAssignmentsChange: vscode.Event<void>;
 
   private readonly assignmentsChange: vscode.EventEmitter<void>;
+  private onChangeServersEmitter: vscode.EventEmitter<void>;
```
kevineger, line 43: `onDidChangeServersEmitter` for consistency

kevineger: Why do we have this member in the first place? Looks like we only ever `fire` it?

src/jupyter/assignments.ts
```diff
@@ -59,9 +63,16 @@ export class AssignmentManager implements vscode.Disposable {
...
-    const ccuInfo = await this.client.ccuInfo();
-    const eligibleGpus = new Set(ccuInfo.eligibleGpus);
-    const ineligibleGpus = new Set(ccuInfo.ineligibleGpus);
+    if (!this.ccuInfo) {
+      const ccuInfo = await CcuInfo.initialize(this.vs, this.client);
+      this.ccuInfo = ccuInfo;
+      this.ccuInfo.onDidChangeCcuInfo.event(() => {
+        this.onChangeServersEmitter.fire();
+      }, this);
+    }
```
kevineger, line 72: Putting this logic here seems kind of _side-effecty_. We're effectively saying _poll for CCU info changes after the first call to `availableServers`_.

Would it make more sense to start "listening" to CCU info changes on construction and simply fire an event when the available servers change? Then the Jupyter server provider can react to those events.

Happy to discuss in person if that isn't clear.

kevineger:
Closing so someone else can take this on.

PR #42, opened 2025-03-05
fix: add validation test for alias names

wh1ter4bb1t-js:
Add a test for testing that a validation error is shown if a user tries
to create an alias greater than the limit (currently 10).

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -3,8 +3,11 @@ import vscode from "vscode";
 import { GoogleAuthProvider } from "./auth/provider";
 import { RedirectUriCodeProvider } from "./auth/redirect";
 import { ColabClient } from "./colab/client";
+import { ServerPicker } from "./colab/server-picker";
+import { AssignmentManager } from "./jupyter/assignments";
 import { getJupyterApi } from "./jupyter/jupyter-extension";
 import { ColabJupyterServerProvider } from "./jupyter/provider";
+import { ServerStorage } from "./jupyter/storage";
 
 /* cSpell:disable */
 const CLIENT_ID =
@@ -33,11 +36,24 @@ export async function activate(context: vscode.ExtensionContext) {
   const colabClient = new ColabClient(new URL("https://localhost:8888"), () =>
     GoogleAuthProvider.getSession(vscode),
   );
+  const serverStorage = new ServerStorage(vscode, context.secrets);
+  const assignmentManager = new AssignmentManager(
+    vscode,
+    colabClient,
+    serverStorage,
+  );
+  const serverPicker = new ServerPicker(vscode);
   const serverProvider = new ColabJupyterServerProvider(
     vscode,
+    assignmentManager,
+    serverPicker,
     jupyter,
-    colabClient,
   );
 
-  context.subscriptions.push(disposeUriHandler, authProvider, serverProvider);
+  context.subscriptions.push(
+    disposeUriHandler,
+    authProvider,
+    assignmentManager,
+    serverProvider,
+  );
 }
--- a/src/jupyter/provider.ts
+++ b/src/jupyter/provider.ts
@@ -1,33 +1,22 @@
-import { randomUUID } from "crypto";
+import { randomUUID, UUID } from "crypto";
 import {
   Jupyter,
   JupyterServer,
   JupyterServerCollection,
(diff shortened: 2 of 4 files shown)


src/colab/server-picker.unit.test.ts
```diff
@@ -145,6 +145,32 @@ describe("ServerPicker", () => {
...
+      acceleratorQuickPickStub.onDidChangeSelection.yield([
+        { value: Accelerator.T4, label: "T4" },
+      ]);
+      await aliasInputShown;
+      aliasInputBoxStub.value = "s".repeat(11);
+      aliasInputBoxStub.onDidChangeValue.yield(aliasInputBoxStub.value);
+
+      expect(aliasInputBoxStub.validationMessage).to.equal(errorMessage);
+      expect(prompt).to.eventually.throw();
+    });
```
kevineger: Nit: either move closer to usage or (prefered) use in-line and instead do a substring/regex match.

```suggestion
    it("returns a validation error message if over character limit", async () => {
      const variantQuickPickStub = stubQuickPickForCall(0);
      const acceleratorQuickPickStub = stubQuickPickForCall(1);
      const aliasInputBoxStub = stubInputBoxForCall(0);

      const variantPickerShown = variantQuickPickStub.nextShow();
      const prompt = serverPicker.prompt(ALL_SERVERS);
      await variantPickerShown;
      const acceleratorPickerShown = acceleratorQuickPickStub.nextShow();
      variantQuickPickStub.onDidChangeSelection.yield([
        { value: Variant.GPU, label: "GPU" },
      ]);
      await acceleratorPickerShown;
      const aliasInputShown = aliasInputBoxStub.nextShow();
      acceleratorQuickPickStub.onDidChangeSelection.yield([
        { value: Accelerator.T4, label: "T4" },
      ]);
      await aliasInputShown;
      aliasInputBoxStub.value = "s".repeat(11);
      aliasInputBoxStub.onDidChangeValue.yield(aliasInputBoxStub.value);

      expect(aliasInputBoxStub.validationMessage).to.equal(/less than 10/);
      expect(prompt).to.eventually.throw();
    });
```

src/colab/server-picker.unit.test.ts
```diff
@@ -145,6 +145,32 @@ describe("ServerPicker", () => {
       });
     });
 
+    it("returns a validation error message if over character limit", async () => {
+      const variantQuickPickStub = stubQuickPickForCall(0);
+      const acceleratorQuickPickStub = stubQuickPickForCall(1);
```
kevineger: Since the point of this test is to exercise alias input validation, pick a variant (`DEFAULT`) which does not have accelerators so we can skip the accelerator selection. Look at / copy from the _prompting for an accelerated is skipped when there are none_ test case.

More brief and focused on the behaviour-under-test.

src/colab/server-picker.unit.test.ts
```diff
@@ -145,6 +145,32 @@ describe("ServerPicker", () => {
...
+      const aliasInputShown = aliasInputBoxStub.nextShow();
+      acceleratorQuickPickStub.onDidChangeSelection.yield([
+        { value: Accelerator.T4, label: "T4" },
+      ]);
+      await aliasInputShown;
+      aliasInputBoxStub.value = "s".repeat(11);
+      aliasInputBoxStub.onDidChangeValue.yield(aliasInputBoxStub.value);
+
+      expect(aliasInputBoxStub.validationMessage).to.equal(errorMessage);
+      expect(prompt).to.eventually.throw();
```
kevineger: Why is `prompt` throwing? I suspect it's because of an async timeout.

Since this test is strictly testing that the validation message informs the user they've exceeded the character length, I'd:

- `void prompt = ...`
- remove this `expect`

src/colab/server-picker.unit.test.ts
wh1ter4bb1t-js: Done.

src/colab/server-picker.unit.test.ts
wh1ter4bb1t-js: Done.

src/colab/server-picker.unit.test.ts
wh1ter4bb1t-js: Done.

PR #43, opened 2025-03-05
fix: Add unique alias validator

wh1ter4bb1t-js:
Adds logic to validator to give validation error in the case a non-unique alias
is used.

- Refactors server picker to pass the list of server labels through to the
  `promptForAlias` function.
- Refactors validator function out into it's own factory function.
- Adds test to check if error appears if attempting to create an alias that
  matches an already set server label.


src/colab/server-picker.ts
```diff
@@ -20,19 +20,21 @@ export class ServerPicker {
     availableServers: ColabServerDescriptor[],
   ): Promise<ColabServerDescriptor | undefined> {
     const variantToAccelerators = new Map<Variant, Set<Accelerator>>();
+    const serverLabels: string[] = [];
     for (const server of availableServers) {
       const accelerators =
         variantToAccelerators.get(server.variant) ?? new Set();
       accelerators.add(server.accelerator ?? Accelerator.NONE);
       variantToAccelerators.set(server.variant, accelerators);
+      serverLabels.push(server.label);
```
kevineger, line 29: `availableServers` are the `ColabServerDescriptor`-s (e.g. the _templates_ of servers which can be created), not the servers which have been assigned. This change as is will prevent the user from setting an alias like "Colab CPU" but will not guard the user from setting an alias "My Server" if it already exists.

kevineger:
Closing so someone else can take this on.

PR #45, opened 2025-03-06
fix: zod-ify auth storage

kevineger:
Bye-bye gnarly type-guards!

Currently only supports a single session, since we only ever need the one scope. Despite this, the implementation is designed to be extensible to multiple sessions in the future (stores an array of sessions). We are likely to do this if and when we support Drive-specific functionality.

### DIFF
--- a/src/auth/provider.ts
+++ b/src/auth/provider.ts
@@ -228,14 +228,7 @@ export class GoogleAuthProvider
       );
     }
     const json: unknown = await response.json();
-    try {
-      return UserInfoSchema.parse(json);
-    } catch (error: unknown) {
-      if (error instanceof z.ZodError) {
-        throw new Error(`Unexpected response object: ${error.message}`);
-      }
-      throw error;
-    }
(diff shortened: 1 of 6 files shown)


src/auth/provider.ts
```diff
@@ -239,134 +228,21 @@ export class GoogleAuthProvider
...
     const json: unknown = await response.json();
-    if (!isUserInfo(json)) {
-      throw new Error(`Invalid user info, got: ${JSON.stringify(json)}`);
+    try {
+      return UserInfoSchema.parse(json);
+    } catch (error: unknown) {
+      if (error instanceof z.ZodError) {
+        throw new Error(`Unexpected response object: ${error.message}`);
+      }
+      throw error;
```
zelliott: Just for my own understanding, why throw a different error? Why not just throw the `z.ZodError`?

kevineger: I guess we may as well just throw it. I'll globally address this elsewhere we're using Zod for consistency.

PR #46, opened 2025-03-07
chore: lint TS file names

kevineger:


eslint.config.mjs
```diff
@@ -21,6 +22,7 @@ export default tseslint.config(
     plugins: {
       "@stylistic/ts": stylisticTs,
       import: importPlugin,
+      "check-file": checkFile,
```
wh1ter4bb1t-js, line 24: nit: should we put string keys next to each other? make all keys strings? keep this alphabetized? I think a couple ideas here to improve findability as the list of plugins will likely grow.
```suggestion
      "@stylistic/ts": stylisticTs,
      "check-file": checkFile,
      import: importPlugin,
```

kevineger: Done. I'm indifferent.

PR #48, opened 2025-03-12
feat: list Colab assignments

kevineger:
Adds a new method to the Colab API client to list assignments.

A few minor additional changes:

- Properly cases "CCU" as "Ccu" (camel).
- Removes `?authuser=0` from all endpoints and applies it globally.
- De-dups test data.
- Fixed _getCCUInfo_ describe to _ccuInfo_ (the SUT).
- Matched on a subset of the error instead of the full response (less brittle)
  for applicable tests.
- Modified `SinonMatcher` for the fetch matching to more robustly match query params.

### DIFF
--- a/src/colab/client.unit.test.ts
+++ b/src/colab/client.unit.test.ts
@@ -7,7 +7,7 @@ import * as sinon from "sinon";
 import { AuthenticationSession } from "vscode";
 import {
   Accelerator,
-  CCUInfo,
+  CcuInfo,
   Assignment,
   Shape,
   SubscriptionState,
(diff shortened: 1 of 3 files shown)


src/colab/client.unit.test.ts
```diff
@@ -229,7 +250,7 @@ function matchAuthorizedRequest(
   method: "GET" | "POST",
 ): SinonMatcher {
   return sinon.match({
-    url: sinon.match(`${DOMAIN}/${endpoint}?authuser=0`),
+    url: sinon.match(new RegExp(`${DOMAIN}/${endpoint}?.*authuser=0`)),
```
amtoney524, line 253: (double checking) should ? be escaped if it's matching for a query string?

kevineger: No need to escape within the `RegExp` constructor.

PR #50, opened 2025-03-12
feat: reconcile assignments on activation

kevineger:
The extension maintains a list of assignments that originate from itself. Upon activation, that list needs to be reconciled with those that Colab maintains as it's possible they were pruned (either due to inactivity, exceeding life-cycle or manually in Colab-web).

### DIFF
--- a/src/jupyter/assignments.unit.test.ts
+++ b/src/jupyter/assignments.unit.test.ts
@@ -142,6 +142,130 @@ describe("AssignmentManager", () => {
     });
   });
 
+  describe("reconcileAssignedServers", () => {
+    it("does nothing when there are no stored servers", async () => {
+      storageStub.list.resolves([]);
+
+      await assignmentManager.reconcileAssignedServers();
+
+      sinon.assert.notCalled(colabClientStub.listAssignments);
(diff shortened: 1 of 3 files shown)


src/jupyter/assignments.unit.test.ts
```diff
@@ -142,6 +142,130 @@ describe("AssignmentManager", () => {
...
+        sinon.assert.calledOnceWithExactly(storageStub.store, twoServers);
+      });
+
+      it("reconciles ignoring assignments originating out of VS Code", async () => {
+        storageStub.list.resolves(servers);
+        const colabAssignment: Assignment = {
+          ...defaultAssignment,
+          runtimeProxyInfo: {
+            ...defaultAssignment.runtimeProxyInfo,
+            url: "https://not-from-vs-code.com",
```
amtoney524, line 255: (curious) when the assignment is from vs code, is a specific url expected here?

kevineger: The URLs will _look the same_ regardless of their assignment origin. The difference will be the `AssignmentManager` won't know about (have persisted) assignments which originated outside of the extension.

PR #55, opened 2025-03-21
feat: add rename server alias command

amtoney524:
This change adds a "Colab: Rename Server Alias" command to the command palette.

renameServerAlias

The Rename Server Alias command lists the assigned servers in storage, allowing the user to make a selection and provide a new alias for the selected server.

### DIFF
--- a/src/colab/server-commands.ts
+++ b/src/colab/server-commands.ts
@@ -26,6 +26,7 @@ export async function renameServerAlias(
 
     return async () => {
       const alias = await input.showInputBox({
+        buttons: [vs.QuickInputButtons.Back],
         placeholder: selectedServer.label,
         prompt: PROMPT_SERVER_ALIAS,
         step: 2,


src/extension.ts
```diff
@@ -59,5 +60,8 @@ export async function activate(context: vscode.ExtensionContext) {
     authProvider,
     assignmentManager,
     serverProvider,
+    vscode.commands.registerCommand("colab.renameServerAlias", () => {
+      void renameServerAlias(vscode, serverStorage);
+    }),
```
kevineger: Nit and totally up to you! Consider omitting parents (`{ }`) for conciseness.

```suggestion
    vscode.commands.registerCommand(
      "colab.renameServerAlias",
      () => void renameServerAlias(vscode, serverStorage),
    ),
```

src/colab/server-commands.ts
```diff
@@ -0,0 +1,42 @@
+import vscode from "vscode";
+import { MultiStepInput } from "../common/multi-step-quickpick";
+import { ColabAssignedServer } from "../jupyter/servers";
+import { ServerStorage } from "../jupyter/storage";
+
+/** Prompt the user to select and rename the local alias used to identify an assigned Colab server. */
```
kevineger: Totally optional style nit, consider multiline. The linter enforces this for non-inline comments, so this inline long comment is totally fine.

```suggestion
/**
 * Prompt the user to select and rename the local alias used to identify an
 * assigned Colab server.
 */
```

src/colab/server-commands.ts
```diff
@@ -0,0 +1,42 @@
...
+export async function renameServerAlias(
+  vs: typeof vscode,
+  serverStorage: ServerStorage,
+): Promise<void> {
+  const servers: ColabAssignedServer[] = await serverStorage.list();
```
kevineger: Totally fine to leave as is, but is there any benefit to annotating the `ColabAssignedServer[]` return type from `.list()`?

src/colab/server-commands.ts
```diff
@@ -0,0 +1,42 @@
...
+        step: 1,
+        title: "Select a Server",
+        totalSteps,
+      })
+    ).value;
+
+    return async () => {
+      const alias = await input.showInputBox({
+        placeholder: selectedServer.label,
+        prompt: "Provide a local convenience alias to the server.",
```
kevineger:
```suggestion
        prompt: "Provide a local convenience alias for the server.",
```

src/colab/server-commands.ts
```diff
@@ -0,0 +1,42 @@
...
+        buttons: [],
+        items: servers.map((s) => ({ label: s.label, value: s })),
+        step: 1,
+        title: "Select a Server",
+        totalSteps,
+      })
+    ).value;
+
+    return async () => {
+      const alias = await input.showInputBox({
```
kevineger, line 28: Could we get some code re-use with the input [where this originally get set](https://github.com/googlecolab/colab-vscode/blob/9802bf3c7c97732ea5c9a2772d45f6b84b741db3/src/colab/server-picker.ts#L157)?

- De-dups code
- Ensures consistency (e.g. same prompt and validation, etc.)

Totally fine if that's simply a shared constant for the `prompt` and a shared exported `function` for the `validate`. Your call on what abstraction makes sense!

src/colab/server-commands.ts
```diff
@@ -0,0 +1,42 @@
...
+        placeholder: selectedServer.label,
+        prompt: "Provide a local convenience alias to the server.",
+        step: 2,
+        title: "Update your Server Alias",
+        totalSteps,
+        validate: (value) =>
+          value.length > 10 ? "Name must be less than 10 characters." : "",
+        value: selectedServer.label,
+      });
+      if (!alias) return undefined;
```
kevineger: Optional nit, omit the `undefined`.

```suggestion
      if (!alias) return;
```

src/colab/server-commands.ts
```diff
@@ -0,0 +1,42 @@
...
+
+      void serverStorage.store([{ ...selectedServer, label: alias }]);
```
kevineger, line 40: Since the `alias` could be the same as `selectedServer.label`, should we check that and return early? Reduces redundant `store`-s.

src/colab/server-commands.unit.test.ts
```diff
@@ -0,0 +1,123 @@
...
+        });
+      void renameServerAlias(vsCodeStub.asVsCode(), serverStorageStub);
+      sinon.assert.calledOnce(serverStorageStub.list);
+      await quickPickStub.nextShow();
+      expect(quickPickStub.items).to.eql([
+        { label: defaultServer.label, value: defaultServer },
+        { label: additionalServer.label, value: additionalServer },
+      ]);
+    });
+
```
kevineger, line 72: Totally your call, but consider a `describe("renaming the default server", ...` to _tell a behavioural story_ across the test cases and reduce setup duplication for _validates the input alias_ and _updates the alias of the selected server_.

Obviously personal preference, but I like the way that reads, e.g. _... renaming the default server validates the input alias_ and _... renaming the default server updates the alias_ (removed the suffix of the latter since the describe addition makes it redundant).

src/colab/server-commands.unit.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      ]);
+
+      await inputBoxStub.nextShow();
+      inputBoxStub.value = "new_alias";
+      inputBoxStub.onDidChangeValue.yield(inputBoxStub.value);
+      inputBoxStub.onDidAccept.yield();
+
+      await expect(rename).to.eventually.be.fulfilled;
+      sinon.assert.calledOnceWithExactly(serverStorageStub.store, [
+        { ...defaultServer, label: "new_alias" },
```
kevineger: Consider adding a test case where the same alias (the placeholder) is used. Functionally equivalent (to the user), but behaviourally different. If you do, and we early return, add an assertion that we don't call `store`.

src/common/multi-step-quickpick.unit.test.ts
```diff
@@ -217,13 +218,25 @@ describe("MultiStepQuickPick", () => {
...
         ...optsToCompare,
         activeItems: [activeItem],
-        // TODO: Back button.
       });
       quickPickStub.onDidChangeSelection.yield([{ label: "foo" }]);
 
       await expect(input).to.eventually.be.fulfilled;
     });
 
+    it("renders the back button by default when the buttons option is undefined", () => {
```
kevineger: I'm curious, do you think the default should be the opposite? In other words if a callers wants buttons, they must specify them? I'm honestly not sure what makes more sense! Your call!

src/colab/server-commands.ts
amtoney524: done, thanks!

src/colab/server-commands.ts
amtoney524: I don't think so, removed :)

src/colab/server-commands.ts
amtoney524: I had wondered the same (e.g., should we wait for 3 references before refactoring) but exporting prompt / validate sgtm. They're being exported from server-picker for now, though happy to find a better long-term home

src/colab/server-commands.ts
amtoney524: done!

src/colab/server-commands.ts
amtoney524: I think this needs to be undefined to satisfy the current return value of InputStep (otherwise it returns a void promise)

src/colab/server-commands.ts
amtoney524: sgtm, made the change and added test coverage

src/colab/server-commands.unit.test.ts
amtoney524: sure! added another describe to stick with the existing convention

src/colab/server-commands.unit.test.ts
amtoney524: done!

src/common/multi-step-quickpick.unit.test.ts
amtoney524: ok I really like that idea! I opened PR #61 for that change - hoping to merge that one first, and will clean up accordingly here

src/extension.ts
amtoney524: done!

src/common/multi-step-quickpick.unit.test.ts
kevineger: Approved https://github.com/googlecolab/colab-vscode/pull/61!

amtoney524: thanks!

PR #57, opened 2025-03-21
feat: client methods for keeping Colab servers alive

kevineger:
A couple random tidbits:

- I chose to `.transform` the `Kernel` API type so we can consistently be camelCase. We may want to consider equivalent transforms on things like `sub` -> `subscriptionState`.
- Rather than define `Kernels` (an array), it felt more ergnomic to define the `Kernel` and use `z.array` in the client.
- Expanded `issueRequest` to work with empty-bodied returns.
- Removed some unecessary test duplication which added little to no value.

### DIFF
--- a/src/colab/api.ts
+++ b/src/colab/api.ts
@@ -182,7 +182,8 @@ export const AssignmentsSchema = z.object({
   assignments: z.array(AssignmentSchema),
 });
 
-// This can be obtained by querying the Jupyter REST API's /api/kernels.
+// This can be obtained by querying the Jupyter REST API's /api/spec.yaml
+// endpoint.
 export const KernelSchema = z
   .object({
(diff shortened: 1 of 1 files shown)


src/colab/api.ts
```diff
@@ -181,3 +181,28 @@ export type Assignment = z.infer<typeof AssignmentSchema>;
 export const AssignmentsSchema = z.object({
   assignments: z.array(AssignmentSchema),
 });
+
+export const KernelSchema = z
```
wh1ter4bb1t-js, line 187: maybe add a link to where this is defined like `https://jupyter-server.readthedocs.io/en/latest/developers/rest-api.html#get--api-spec.yaml`

kevineger: Done.

PR #59, opened 2025-03-24
feat: keep Colab servers alive

kevineger:
Demo of the keep-alive notifications. I manually configured the `ServerKeepAliveController` to run on much shorter intervals than the default/real implementation will. With that said, it's still a lengthy and boring demo:

[keep-alive notifications.webm](https://github.com/user-attachments/assets/de4d730c-8edd-4b09-9947-59374031f953)

Our logic here differs from what we do in Colab (web) for two main reasons:

1. We don't have the APIs needed across VS Code and the Jupyter extension to associate an active notebook editor to the Colab server that yielded the connected kernel. https://github.com/microsoft/vscode-jupyter/issues/16487 was filed on the Jupyter extension repository both verifying I'm not overlooking something but also as a feature request to make that possible. See issue for details.
2. The interaction pattern in VS Code is likely a bit different than Colab (web). Drawing from personal experiences, it's more common for editor tabs to be closed and re-opened compared to web tabs, like Colab (web).

As highlighted in the implementation comment, we keep servers alive (by sending the `/keep-alive` request) if:

- the server has been active
- the server's lifetime has been extended by the user and is still within the extension period
- the user extends the server's lifetime

Otherwise, we cease to send `/keep-alive` requests. If a server was previously not kept alive but there was recent activity on it, it will be kept alive again according to the aforementioned rules.

The `console` logging should be fine for now. Longer term we'll move that to telemetry. In practice, when properly configured we shouldn't see any of those logs anyways.

(At the time of sending this PR) The test structure is as follows. This behaviour-centric testing should make understanding the class functionality easier.

```txt
ServerKeepAliveController
  lifecycle
    ✔ disposes the interval timeout
    ✔ skips when a keep-alive is already in flight
    ✔ aborts slow keep-alive attempts
  with no assigned servers
    ✔ does nothing
  with an "active" server
    ✔ sends a keep-alive request
  with an "idle" server
    ✔ prompts the user to keep it running
    ✔ counts down the time to extend
    which the user does not extend
      ✔ does not send keep-alive requests
      ✔ does not prompt for extension again
      ✔ starts sending keep-alive requests when used again
    which the user extends
      ✔ sends a keep-alive request
      and then uses
        ✔ sends a keep-alive request
        ✔ does not prompt to keep it running
  with a mix of "active" and "idle" servers
    ✔ only prompts to "idle" servers
    ✔ only sends keep-alive requests for "active" and extended servers
    ✔ swallows keep-alive failures
  with a server that has multiple kernels
    ✔ respects the most recent kernel activity
    ✔ does not send a keep-alive request if all kernels are idle
```

---

I suspect the choices made here won't be final, but this feels like the best option right now.

---

In a follow-up change, I intend to abstract some of the interval/polling logic into a generic class since there's an additional use case for this and it makes sense for it to be implemented robustly once.

### DIFF
--- a/src/colab/keep-alive.ts
+++ b/src/colab/keep-alive.ts
@@ -81,7 +81,6 @@ export class ServerKeepAliveController implements Disposable {
     const timeout = setTimeout(
       () => {
         abort.abort("Timed out while keeping servers alive");
-        this.inFlight = undefined;
       },
       // The underlying calls to get the assigned servers and send the
       // "keep-alive" signal are quick. Twice the prompt time should be plenty.
--- a/src/colab/keep-alive.unit.test.ts
+++ b/src/colab/keep-alive.unit.test.ts
@@ -11,6 +11,15 @@ import { ServerKeepAliveController } from "./keep-alive";
 const NOW = new Date();
 const ONE_SECOND_MS = 1000;
 const ONE_MINUTE_MS = ONE_SECOND_MS * 60;
+const ABORTING_KEEP_ALIVE = async (
+  _: string,
+  signal?: AbortSignal,
+): Promise<void> =>
+  new Promise((_, reject) => {
+    signal?.addEventListener("abort", () => {
+      reject(new Error("Aborted"));
+    });
+  });
 
 const CONFIG = {
   inactivityThresholdMs: ONE_MINUTE_MS * 60, // 1 hour.
@@ -89,10 +98,7 @@ describe("ServerKeepAliveController", () => {
       colabClientStub.listKernels
         .withArgs(defaultServer.endpoint)
         .resolves([DEFAULT_KERNEL]);
-      colabClientStub.keepAlive.callsFake(
-        // eslint-disable-next-line @typescript-eslint/no-empty-function
-        async () => new Promise(() => {}),
-      );
+      colabClientStub.keepAlive.callsFake(ABORTING_KEEP_ALIVE);
       serverKeepAliveController = new ServerKeepAliveController(
         vsCodeStub.asVsCode(),
         colabClientStub,
@@ -116,10 +122,7 @@ describe("ServerKeepAliveController", () => {
       colabClientStub.listKernels
         .withArgs(defaultServer.endpoint)
         .resolves([DEFAULT_KERNEL]);
-      colabClientStub.keepAlive.onFirstCall().callsFake(
-        // eslint-disable-next-line @typescript-eslint/no-empty-function
-        async () => new Promise(() => {}),
-      );
+      colabClientStub.keepAlive.onFirstCall().callsFake(ABORTING_KEEP_ALIVE);
       serverKeepAliveController = new ServerKeepAliveController(
         vsCodeStub.asVsCode(),
         colabClientStub,


src/colab/keep-alive.ts
```diff
@@ -0,0 +1,217 @@
...
+   * "keep-alive signal for a given server.
+   */
+  idleExtensionPromptTimeMs: number;
+  /**
+   * How long (in milliseconds) to extend the "keep-alive" signal.
+   */
+  idleExtensionMs: number;
+}
+
+const DEFAULT_CONFIG: Config = {
```
zelliott, line 30: It might be useful to `export` this if a client wants to override one member but keep the rest. Then they can do:

```ts
const MY_SPECIAL_CONFIG = {
  ...DEFAULT_CONFIG,
  idleExtensionMs: 100 * 60 * 60 // 60 minutes.
};
```

src/colab/keep-alive.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  private async maybeKeepServersAlive(): Promise<void> {
+    if (this.inFlight) {
+      console.log("Keep alive already in flight. Skipping.");
+      return;
+    }
+
+    const abort = new AbortController();
+    this.inFlightAbort = abort;
+    const timeout = setTimeout(
```
zelliott, line 81: I know you're refactoring in https://github.com/googlecolab/colab-vscode/pull/60, so feel free to apply changes there.

The built-in https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static might be helpful here, see also the example at https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/any_static#examples for combining multiple `AbortSignal`s. But what you have here looks fine too.

src/colab/keep-alive.ts
```diff
@@ -0,0 +1,217 @@
...
+  private extendedKeepAlive = new Map<UUID, Date>();
+  private tombstones = new Set<UUID>();
+
+  constructor(
+    private readonly vs: typeof vscode,
+    private readonly colabClient: ColabClient,
+    private readonly assignmentManager: AssignmentManager,
+    private readonly config: Config = DEFAULT_CONFIG,
+  ) {
+    this.poller = this.startEndpointPolling();
```
zelliott, line 54: nit: It feels a little odd to me for the controller to start polling on construction. I think I might expect a separate `start` method, but will defer to your judgment.

If you do end up creating a separate `start` method, then make sure that it handles various edge cases (e.g. called multiple times, called when disposed).

src/colab/keep-alive.ts
```diff
@@ -0,0 +1,217 @@
...
+      this.inFlight = this.keepServersAlive(abort.signal);
+      await this.inFlight;
+    } finally {
+      clearTimeout(timeout);
+      this.inFlight = undefined;
+    }
+  }
+
+  private async keepServersAlive(signal: AbortSignal): Promise<void> {
+    const assignments = await this.assignmentManager.getAssignedServers();
```
zelliott, line 99: Why is the `AbortSignal` not passed to `getAssignedServers`?

src/colab/keep-alive.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  private async keepServerAlive(
+    assignment: ColabAssignedServer,
+    signal: AbortSignal,
+  ): Promise<void> {
+    const kernels = await this.colabClient.listKernels(
+      assignment.endpoint,
+      signal,
+    );
+    if (await this.shouldKeepAlive(assignment, kernels)) {
```
zelliott, line 119: Suppose all of the async calls up-to-and-including `shouldKeepAlive` eat up all of the timeout time. User clicks "Extend", so `shouldKeepAlive` returns true. We call `keepAlive`, but the `AbortSignal` is aborted because the timeout time has passed, so it immediately fails, and user sees an error. Am I getting that right?

I don't think it's the worst behavior in the world, but it does feel a bit odd. I wonder if the solution is to have separate timeouts for each network request. Not sure, feel free to ack.

src/colab/keep-alive.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  /**
+   * Determines if a server should be kept alive.
+   *
+   * Returns true if:
+   *
+   * - the server has been active
+   * - the server's lifetime has been extended by the user and is still within
+   *   the extension period
+   * - the user extends the server's lifetime
```
zelliott: Are these OR'd? Also, is bullet point 3 already encompassed by bullet point 2? They both have _"user extending the server's lifetime"_, I'm not sure how 2 and 3 are different.

src/colab/keep-alive.ts
```diff
@@ -0,0 +1,217 @@
...
+      );
+      return true;
+    }
+
+    this.tombstones.add(assignment.id);
+    return false;
+  }
+
+  private async extendKeepAlive(serverLabel: string): Promise<boolean> {
+    return new Promise<boolean>((resolve) => {
```
zelliott, line 189: See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers. You might prefer this because it'll reduce nesting.

src/colab/keep-alive.unit.test.ts
```diff
@@ -0,0 +1,525 @@
...
+const CONFIG = {
+  inactivityThresholdMs: ONE_MINUTE_MS * 60, // 1 hour.
+  keepAliveIntervalMs: ONE_MINUTE_MS * 5, // 5 minutes.
+  idleExtensionPromptTimeMs: ONE_SECOND_MS * 10, // 10 seconds.
+  idleExtensionMs: ONE_MINUTE_MS * 30, // 30 minutes.
+};
+
+const DEFAULT_KERNEL: Kernel = {
+  id: "456",
+  name: "Kermit the Kernel",
```
zelliott, line 33: Nice

src/colab/keep-alive.ts
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * Keeps Colab servers alive while they are recently used, or if the user
+ * explicitly extends their lifetime.
+ */
+export class ServerKeepAliveController implements Disposable {
+  private inFlight?: Promise<void>;
+  private inFlightAbort?: AbortController;
+  private poller: NodeJS.Timeout;
+  private extendedKeepAlive = new Map<UUID, Date>();
+  private tombstones = new Set<UUID>();
```
zelliott: Can these be `readonly`?

src/colab/keep-alive.ts
```diff
@@ -0,0 +1,217 @@
...
+      console.log("Keep alive already in flight. Skipping.");
+      return;
+    }
+
+    const abort = new AbortController();
+    this.inFlightAbort = abort;
+    const timeout = setTimeout(
+      () => {
+        abort.abort("Timed out while keeping servers alive");
+        this.inFlight = undefined;
```
zelliott: Is this needed? Won't the `finally` block run on `inFlight` abort and do this?

src/colab/keep-alive.ts
kevineger: I wish I could easily use it. [Previously](https://github.com/googlecolab/colab-vscode/pull/12#discussion_r1932259042) we discussed the same thing.

As before, `withResolvers` requires ES2024. I tried upgrading our target but could not get the produced JS to target the newer ES. I believe we spoke in person about this and you said you were interested in tackling this? If you're still willing to take a pass that'd be awesome and appreciated 🙏.

For this PR, as before, sticking with this.

src/colab/keep-alive.ts
kevineger: They're different. I'll rephrase.

src/colab/keep-alive.ts
kevineger: Because that is simply a storage lookup and `vscode.SecretStorage.get` does not accept an `AbortSignal`.

zelliott: It's async, hypothetically it could take longer than the timeout? For APIs that don't accept `AbortSignal`, you can race against one. See http://google3/research/colab/frontend/common/base/promise.ts;l=302;rcl=728396757. But also seems low priority.

src/colab/keep-alive.ts
zelliott: Ah sorry for forgetting and bringing this up again. Filed b/406360866 and assigned to me.

src/colab/keep-alive.ts
kevineger: I'll take that as a follow-up. In practice I can't imagine the overhead is worth it but will take a closer look.

src/colab/keep-alive.ts
kevineger: SG. I'll apply in https://github.com/googlecolab/colab-vscode/pull/60 to do it once.

src/colab/keep-alive.ts
kevineger: I'll export it once I have a use case for it if that's alright.

src/colab/keep-alive.ts
kevineger: Done.

src/colab/keep-alive.ts
kevineger: Until there's a use case of deferring starting the polling - I'll keep it in the ctor.

src/colab/keep-alive.ts
kevineger: The abort signal is really the time to complete the task we effectively run on a cron. I don't think it's too weird since none of these errors will be user visible, in practice they should never happen and if we do abort, that's because we've overrun and are re-running the task. In the case you mention, the user's "response" to extending is preserved and they won't be prompted again.

Going to leave as is.

src/colab/keep-alive.ts
kevineger: I think I added it because I didn't setup my test well enough:

```ts
colabClientStub.keepAlive.onFirstCall().callsFake(
  // eslint-disable-next-line @typescript-eslint/no-empty-function
  async () => new Promise(() => {}),
);
```

To mimic `fetch` reacting to the abort, I updated the test to do:

```ts
colabClientStub.keepAlive.onFirstCall().callsFake(
  async (_, signal) =>
    new Promise((_, reject) => {
      signal?.addEventListener("abort", () => {
        reject(new Error("Aborted"));
      });
    }),
);
```

... and remove this, leaning just on the `finally`.

Thanks for noticing and the push! Much better.

src/colab/keep-alive.ts
kevineger: Went to apply this to https://github.com/googlecolab/colab-vscode/pull/60, but I'm not sure we should.

`AbortSignal.timeout()` "_is based on active rather than elapsed time, and will effectively be paused if the code is running in a suspended worker, or while the document is in a back-forward cache ('[bfcache](https://web.dev/articles/bfcache)')._"

I believe VS Code extensions run in a WebWorker and those can get suspended. Since here we care about _wall clock_ time, not active time processing, sticking with `setTimeout` (here and in https://github.com/googlecolab/colab-vscode/pull/60).

PR #60, opened 2025-03-24
feat: abstract task running logic

kevineger:
The same interval-based task running logic is needed in multiple places (keep-alive and CCU info). This centralizes it into a standalone class and refactors the existing classes to leverage it.

I also took the chance to fix an issue with the `CcuInformation` class (renamed as `CcuInformationManager`) where we were exposing `onDidChangeCcuInfo: EventEmitter<void>;` instead of simply the `vscode.Event`. Additionally took a pass at improving test clarity + behavioural coverage.

Lint change is to support `chai` expressions like `expect(foo).to.be.true;`.

### DIFF
--- a/src/colab/ccu-info.ts
+++ b/src/colab/ccu-info.ts
@@ -11,9 +11,9 @@ const TASK_TIMEOUT_MS = 1000 * 10; // 10 seconds.
  */
 export class CcuInformationManager implements Disposable {
   readonly onDidChangeCcuInfo: vscode.Event<void>;
-  private emitter: vscode.EventEmitter<void>;
+  private readonly emitter: vscode.EventEmitter<void>;
   private _ccuInfo?: CcuInfo;
-  private runner: SequentialTaskRunner;
+  private readonly runner: SequentialTaskRunner;
 
   constructor(
     private readonly vs: typeof vscode,
@@ -26,7 +26,7 @@ export class CcuInformationManager implements Disposable {
     this.runner = new SequentialTaskRunner(
       { intervalTimeoutMs: POLL_INTERVAL_MS, taskTimeoutMs: TASK_TIMEOUT_MS },
       (signal) => this.updateCcuInfo(signal),
-      OverrunPolicy.AllowToComplete,
+      OverrunPolicy.AbandonAndRun,
     );
   }
 
--- a/src/common/task-runner.ts
+++ b/src/common/task-runner.ts
@@ -34,7 +34,7 @@ export enum OverrunPolicy {
 export class SequentialTaskRunner implements Disposable {
   private inFlight?: Promise<void>;
   private inFlightAbort?: AbortController;
-  private timeout: NodeJS.Timeout;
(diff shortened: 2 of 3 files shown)


src/colab/ccu-info.ts
```diff
@@ -1,93 +1,68 @@
...
-export class CcuInformation implements Disposable {
-  onDidChangeCcuInfo: EventEmitter<void>;
-  private currentCcuInfo?: CcuInfo;
-  private poller: NodeJS.Timeout;
-  private isFetching = false;
+export class CcuInformationManager implements Disposable {
+  readonly onDidChangeCcuInfo: vscode.Event<void>;
+  private emitter: vscode.EventEmitter<void>;
+  private _ccuInfo?: CcuInfo;
+  private runner: SequentialTaskRunner;
```
zelliott: `readonly`? Also apply to other members in this class if applicable. Also apply to `keep-alive.ts` if applicable.

src/common/task-runner.ts
```diff
@@ -0,0 +1,79 @@
...
+    clearInterval(this.timeout);
+    this.inFlightAbort?.abort();
+  }
+
+  private async run(): Promise<void> {
+    if (this.inFlight) {
+      switch (this.overrun) {
+        case OverrunPolicy.AllowToComplete:
+          return;
+        case OverrunPolicy.AbandonAndRun:
```
zelliott, line 57: Do you have a use case in mind for this?

eslint.config.mjs
```diff
@@ -66,6 +66,7 @@ export default tseslint.config(
     files: ["**/*.unit.test.ts"],
     rules: {
       "@typescript-eslint/unbound-method": "off",
+      "@typescript-eslint/no-unused-expressions": "off",
```
zelliott, line 69: Why?

src/common/task-runner.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * Runs a task at a regular interval, ensuring that only one task is running at
+ * a time.
+ */
+export class SequentialTaskRunner implements Disposable {
+  private inFlight?: Promise<void>;
+  private inFlightAbort?: AbortController;
+  private timeout: NodeJS.Timeout;
```
zelliott: `readonly`

src/common/task-runner.ts
kevineger: Woops, had meant to set the `CcuInformationManager`'s `runner` to this. Done.

src/common/task-runner.ts
kevineger: Done.

src/colab/ccu-info.ts
kevineger: Done.

eslint.config.mjs
kevineger: Expressions like the ones used in this PR - `expect(colabClientStub.keepAlive.firstCall.args[1]?.aborted).to.be.true;`. Added note to PR description.

PR #61, opened 2025-03-25
refactor: make quickinput buttons configurable

amtoney524:
This change allows callers of MultiStepInput to configure which buttons are shown

### DIFF
--- a/src/common/multi-step-quickpick.ts
+++ b/src/common/multi-step-quickpick.ts
@@ -37,6 +37,8 @@ export interface QuickInputOptions {
   step: number;
   /** The total step count. */
   totalSteps: number;
+  /** Buttons for actions in the UI. */
+  buttons?: QuickInputButton[];
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/common/multi-step-quickpick.ts
```diff
@@ -72,7 +74,7 @@ export interface InputBoxOptions extends QuickInputOptions {
  * A chainable multi-step runner for quick-inputs.
  */
 export class MultiStepInput {
-  private constructor(private readonly vs: typeof vscode) {}
```
kevineger, line 75: Why remove?

amtoney524: good q! intention was to expose `vs` on the `MultiStepInput` so that we have a reference to it wherever we're using `MultiStepInput`; otherwise, for something like the server picker, we'd need to pass `vs` to functions like `promptForAlias`, `promptForAccelerator` in order to specify the back button

PR #62, opened 2025-03-27
feat: auto-increment the server placeholder

kevineger:
Follows standard file explorer semantics for duplicating files.

- If no server exists, placeholder is `Colab CPU`.
- If `Colab CPU` exists, placeholder is `Colab CPU (1)`.
- If `Colab CPU` and `Colab CPU (1)` exist, placeholder is `Colab CPU (2)`.
- If `Colab CPU` and `Colab CPU (2)` exist (because `1` was removed),
  placeholder is `Colab CPU (1)`.

The above should be easily understood looking at the tests.

Determining the number to use requires the `AssignmentManager`, so the functions
were moved to members of the `ServerPicker` class.

I also re-organized the file a bit so exported members are at the top.

### DIFF
--- a/src/colab/server-picker.ts
+++ b/src/colab/server-picker.ts
@@ -1,13 +1,25 @@
 import vscode, { QuickPickItem } from "vscode";
 import { InputStep, MultiStepInput } from "../common/multi-step-quickpick";
+import { AssignmentManager } from "../jupyter/assignments";
 import { ColabServerDescriptor } from "../jupyter/servers";
 import { Accelerator, Variant } from "./api";
+
+/** Provides an explanation to the user on updating the server alias. */
+export const PROMPT_SERVER_ALIAS =
+  "Provide a local convenience alias to the server.";
+
+/** Validates the server alias. */
+export const validateServerAlias = (value: string) =>
+  value.length > 10 ? "Name must be less than 10 characters." : "";
 
 /**
  * Supports prompting the user to pick a Colab server to be created.
  */
 export class ServerPicker {
-  constructor(private readonly vs: typeof vscode) {}
+  constructor(
+    private readonly vs: typeof vscode,
+    private readonly assignments: AssignmentManager,
+  ) {}
 
   /**
    * Prompt the user through a multi-step series of inputs to pick a Colab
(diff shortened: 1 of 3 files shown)


src/colab/server-picker.ts
```diff
@@ -1,13 +1,25 @@
 import vscode, { QuickPickItem } from "vscode";
 import { InputStep, MultiStepInput } from "../common/multi-step-quickpick";
+import { AssignmentManager } from "../jupyter/assignments";
 import { ColabServerDescriptor } from "../jupyter/servers";
 import { Accelerator, Variant } from "./api";
 
+/** Provides an explanation to the user on updating the server alias. */
+export const PROMPT_SERVER_ALIAS =
```
amtoney524, line 8: thanks for moving these!

src/colab/server-picker.unit.test.ts
```diff
@@ -190,6 +212,116 @@ describe("ServerPicker", () => {
...
+            label: "Colab GPU A100 (1)",
+          },
+        ];
+        assignmentStub.getAssignedServers.resolves(servers);
+        const input = await progressToAliasInput();
+
+        expect(input.placeholder).to.equal("Colab GPU T4 (1)");
+      });
+
+      // To ensure a string sort isn't used, which would put "10" before "2".
```
amtoney524, line 294: is this maybe a stale comment?

src/colab/server-picker.ts
```diff
@@ -47,6 +59,123 @@ export class ServerPicker {
...
+    return;
+  }
+
+  private async getPlaceholder(
+    variant: Variant,
+    accelerator?: Accelerator,
+  ): Promise<string> {
+    const servers = await this.assignments.getAssignedServers();
+    const relevantServers = servers.filter(
+      (s) => s.variant === variant && s.accelerator === accelerator,
```
amtoney524, line 157: checking my understanding -  on an assigned server, the accelerator field can be undefined or Accelerator.NONE?

src/colab/server-picker.unit.test.ts
kevineger: I put it in to rationalize why there's the test above (with _multiple assigned servers_) and this one, with 11 servers.

src/colab/server-picker.ts
kevineger: Correct! A function of our API being a little funky. `src/colab/api.ts` for details.

src/colab/server-picker.ts
kevineger: No sweat

PR #63, opened 2025-04-02
feat: limit TPUs to available accelerators

kevineger:
The CCU API now returns the list of available and ineligible TPUs. With that, we can now properly limit the available TPUs.

Included in this change is the addition of _Colab TPU v6e-1_ availability.

I also commented out now fully deprecated GPU accelerators which is consistent to other API-deprecated fields.

### DIFF
--- a/src/jupyter/assignments.ts
+++ b/src/jupyter/assignments.ts
@@ -262,11 +262,8 @@ function isAcceleratorAvailable(
   if (!accelerator) {
     return false;
   }
-  if (availability.ineligible?.has(accelerator)) {
-    return false;
-  }
-  if (availability.eligible.has(accelerator)) {
-    return true;
-  }
-  return false;
+  return (
+    !availability.ineligible?.has(accelerator) &&
+    availability.eligible.has(accelerator)
+  );
 }


src/colab/ccu-info.unit.test.ts
```diff
@@ -18,6 +18,8 @@ const DEFAULT_CCU_INFO: CcuInfo = {
   assignmentsCount: 3,
   eligibleGpus: [Accelerator.T4],
   ineligibleGpus: [Accelerator.A100, Accelerator.L4],
+  eligibleTpus: [Accelerator.V6E1, Accelerator.V28],
```
metrizable, line 21: v6e1, not sure will be eligible. fine to ship now

src/jupyter/assignments.ts
```diff
@@ -246,3 +250,23 @@ function colabProxyFetch(
...
+
+// TODO: Provide a ⚠️ warning for the servers which are ineligible.
+function isAcceleratorAvailable(
+  accelerator: Accelerator | undefined,
+  availability: {
+    eligible: Set<Accelerator>;
+    ineligible?: Set<Accelerator>;
+  },
+): boolean {
+  if (!accelerator) {
```
metrizable, line 262: maybe just:

```typescript
  if (!accelerator) {
    return false;
  }
  return !availability.ineligible?.has(accelerator) && availability.eligible.has(accelerator);
```

also noting a small semantics oddity re: the disjoint property of these two sets, yet we are checking both. is my thinking right here?

kevineger: [Internal source link](http://google3/research/colab/datalab/tunnel/managedapi.proto;l=27-34;rcl=726937946) where these fields are described.

- Eligible: _The list of eligible GPU accelerators for the requesting user (w.r.t. their CCU balance and maybe their subscription status) in the priority order._
- Ineligible: _The list of ineligible GPU accelerators for the requesting user. Browser code may use this to display the items as disabled, and upsell CCUs if this is non-empty._

Until we have any degree of _pro upselling_ in this extension, I think it makes the most sense to filter these out as is being done. But please correct me if my interpretation/rationale is wrong.

---

Simplified return as suggested.

src/colab/ccu-info.unit.test.ts
kevineger: This is just test data. But for what it's worth, this is the eligible/ineligible combo I'm seeing right now (at least in our sandbox env.).

PR #64, opened 2025-04-04
feat: add remove server command

amtoney524:
This change adds the `Colab: Remove Server` command to the command palette, allowing users to remove server assignments.

removeServer

### DIFF
--- a/src/jupyter/assignments.unit.test.ts
+++ b/src/jupyter/assignments.unit.test.ts
@@ -458,11 +458,6 @@ describe("AssignmentManager", () => {
   });
 
   describe("unassignServer", () => {
-    it("attempts to remove the server from storage", async () => {
-      await assignmentManager.unassignServer(defaultServer);
-      sinon.assert.calledOnceWithExactly(storageStub.remove, defaultServer.id);
-    });
-
     describe("when the server does not exist", () => {
       beforeEach(() => {
         storageStub.remove.resolves(false);


src/colab/client.ts
```diff
@@ -102,6 +102,28 @@ export class ColabClient {
...
+   * @param endpoint - The endpoint to unassign.
+   */
+  async unassign(endpoint: string, signal?: AbortSignal): Promise<void> {
+    const url = new URL(`${TUN_ENDPOINT}/unassign/${endpoint}`, this.domain);
+    const { token } = await this.issueRequest(
+      url,
+      {
+        method: "GET",
+        signal,
+      },
```
kevineger: Optional style nit: 

```suggestion
      { method: "GET", signal },
```

src/colab/server-commands.unit.test.ts
```diff
@@ -147,4 +148,48 @@ describe("Server Commands", () => {
       });
     });
   });
+
+  describe("removeServer", () => {
```
kevineger, line 152: Consider testing the full behaviour of this function/command.

describe _removeServer_

- it _[TODO] when there are no assigned servers_
- describe _with a single assigned server_
  - it _lists the server for selection_
  - it _unassigns the selected server_
- describe _with multiple assigned servers_
  - it _lists the server for selection_
  - it _unassigns the selected server_

src/colab/server-commands.ts
```diff
@@ -41,3 +43,31 @@ export async function renameServerAlias(
...
+// TODO: Handle bug where, if the server of the connected kernel is
+// removed, a fallback kernel is selected but does not connect.
+// TODO: Consider adding a notification that the server was removed.
+// TODO: Update MultiStepInput to handle a single-step case.
+export async function removeServer(
+  vs: typeof vscode,
+  serverStorage: ServerStorage,
+  assignmentManager: AssignmentManager,
+) {
+  const servers = await serverStorage.list();
```
kevineger: Could we remove the `serverStorage` dependency and insteady strictly rely on the `AssignmentManager` (`getAssignedServers`)?

src/jupyter/assignments.ts
```diff
@@ -158,6 +158,17 @@ export class AssignmentManager implements vscode.Disposable {
...
 
+  /**
+   * Unassigns the given server.
+   *
+   * @param server - The server to remove.
+   */
+  async unassignServer(server: ColabAssignedServer): Promise<void> {
+    await this.storage.remove(server.id);
+    this.assignmentsChange.fire();
+    await this.client.unassign(server.endpoint);
```
kevineger, line 180: Just to confirm, we don't need to first delete the kernel (and/or sessions) to have the Jupyter extension _gracefully_ disconnect before we unassign the Colab server (effectively disconnecting the backend socket connection)?

src/jupyter/assignments.ts
```diff
@@ -158,6 +158,17 @@ export class AssignmentManager implements vscode.Disposable {
     return this.assignOrRefresh(server);
   }
 
+  /**
+   * Unassigns the given server.
+   *
+   * @param server - The server to remove.
+   */
+  async unassignServer(server: ColabAssignedServer): Promise<void> {
+    await this.storage.remove(server.id);
```
kevineger: Check TS doc for this, _@returns true if a server was stored and has been removed, or false if the server does not exist._. We should probably use the return to determine if the `unassign` call is made.

Additionally, cover this behaviour with a test case.

src/colab/server-commands.unit.test.ts
kevineger: The _TODO_ above isn't immediately obvious to me, curious what you think!

I think we could either (a) constrain/disable the command when there are no servers, (b) render the picker with a prompt or placeholder that says there are no servers to remove or (c) pop a notification.

My knee-jerk reaction is (a).

I suppose the above would/should likely apply to the rename-alias command for consistency.

src/colab/client.ts
amtoney524: done!

src/colab/server-commands.ts
amtoney524: sgtm, updated to work off `AssignmentManager` only

src/colab/server-commands.unit.test.ts
amtoney524: I also agree with a and would apply that treatment to rename server alias as well - would this be alright to address in a follow up cl?

also, for the test cases, since single and multiple servers have the same treatment right now, I'm not sure if it makes sense to have separate describe blocks at this point. I think the main difference would be the number of items displayed in the picker, and asserting behavior for multiple items would assert the single item case by default. Let me know if I'm thinking about this the wrong way through!

src/jupyter/assignments.ts

src/jupyter/assignments.ts
amtoney524: I ended up adding logic to clean up any sessions before unassigning the server - why? if we unassign the server and immediately list kernels after, they're still available. Cleaning up the sessions and then unassigning the server seems consistent with the behavior in Colab web and, according to the [Jupyter server docs](https://jupyter-server.readthedocs.io/en/latest/developers/architecture.html#delete-session-workflow), should shut down the kernel

src/jupyter/assignments.ts
```diff
@@ -158,6 +158,25 @@ export class AssignmentManager implements vscode.Disposable {
...
+   *
+   * @param server - The server to remove.
+   */
+  async unassignServer(server: ColabAssignedServer): Promise<void> {
+    const removed = await this.storage.remove(server.id);
+    if (!removed) {
+      return;
+    }   
+    this.assignmentsChange.fire();
+    await Promise.all(
```
kevineger, line 175: _Creates a Promise that is resolved with an array of results when all of the provided Promises resolve, or rejected when any Promise is rejected._

Considering callers could end up in some kind of hybrid state where some sessions were deleted and others weren't, consider dropping a note in the TS doc?

Something to the effect of _Deletes all kernel sessions for the specified server before unassigning. Only unassigns if all session deletions succeed._

src/jupyter/assignments.unit.test.ts
```diff
@@ -456,6 +456,108 @@ describe("AssignmentManager", () => {
...
+      it("does not unassign the server", async () => {
+        await assignmentManager.unassignServer(defaultServer);
+        sinon.assert.notCalled(colabClientStub.unassign);
+      });
+
+      it("does not emit an assignment change event", async () => {
+        const listener = sinon.stub();
+        assignmentManager.onDidAssignmentsChange(listener);
+        await assignmentManager.unassignServer(defaultServer);
+        sinon.assert.notCalled(colabClientStub.unassign);
```
kevineger: Nit: split test body into arrange-act-assert blocks. Internal guidance: go/unit-testing-practices?polyglot=typescript#structure

Apply globally.

```suggestion
        const listener = sinon.stub();
        assignmentManager.onDidAssignmentsChange(listener);

        await assignmentManager.unassignServer(defaultServer);
        
        sinon.assert.notCalled(colabClientStub.unassign);
```

src/jupyter/assignments.unit.test.ts
```diff
@@ -456,6 +456,108 @@ describe("AssignmentManager", () => {
...
+        sinon.assert.notCalled(colabClientStub.listSessions);
+      });
+
+      it("does not unassign the server", async () => {
+        await assignmentManager.unassignServer(defaultServer);
+        sinon.assert.notCalled(colabClientStub.unassign);
+      });
+
+      it("does not emit an assignment change event", async () => {
+        const listener = sinon.stub();
```
kevineger, line 479: Should we add a `sinon.assert.notCalled` on the `listener`?

src/jupyter/assignments.unit.test.ts
```diff
@@ -456,6 +456,108 @@ describe("AssignmentManager", () => {
     });
   });
 
+  describe("unassignServer", () => {
+    it("attempts to remove the server from storage", async () => {
+      await assignmentManager.unassignServer(defaultServer);
+      sinon.assert.calledOnceWithExactly(storageStub.remove, defaultServer.id);
+    });
```
kevineger: I'd consider removing this test case. It isn't framed around the public _behaviour_ of the SUT. Consider this: a caller of the `AssignmentManager` may not even know/care how servers are maintained (storage, memory, etc.).

We can add the `storageStub.remove` sinon asserts to the relevant `it` blocks in the other tests where appropriate.

src/jupyter/assignments.unit.test.ts
```diff
@@ -456,6 +456,108 @@ describe("AssignmentManager", () => {
...
+      });
+
+      it("lists sessions", async () => {
+        colabClientStub.listSessions.resolves([]);
+        await assignmentManager.unassignServer(defaultServer);
+        sinon.assert.calledOnceWithExactly(
+          colabClientStub.listSessions,
+          defaultServer.endpoint,
+        );
+      });
```
kevineger: As above regarding the _attempts to remove the server from storage_ test case, consider focusing on testing the public behaviour over the implementation.

Internal guidance: go/unit-testing-practices?polyglot=typescript#behavior-testing

The _deletes sessions_ test effectively covers this.

src/colab/server-commands.unit.test.ts
kevineger: Totally fine to address in another PR, just drop a TODO!

> RE: distinct test cases for no servers, one server and many servers

You're totally right in that it exercises the exact same code, though I tend to think about this from a behaviour perspective. In English, _the `AssignmentManager` can remove a server when it's the only one_ and _the `AssignmentManager` can remove a server when there are multiple_ are two distinct behaviours callers rely on. 

If I'm a new dev to the project or I haven't seen this class/SUT before - can I look at just the tests and know exactly how it behaves? For example, with just the one test that has multiple servers, one "reasonable" assumption would be that when `removeServer` is called when we only have a single server, it skips the quick-pick.

These are just super nitty points though that really don't matter at the end of the day, but hopefully that makes at least some sense! Totally fine to ignore. Also multiple ways to accomplish the above without the terse `describe`/`it` structure I gave as an example. Your discretion!

src/colab/server-commands.unit.test.ts
```diff
@@ -147,4 +148,40 @@ describe("Server Commands", () => {
...
+    it("lists assigned servers for selection", async () => {
+      const additionalServer = { ...defaultServer, label: "bar" };
+      const assignmentManagerStub: SinonStubbedInstance<AssignmentManager> =
+        sinon.createStubInstance(AssignmentManager, {
```
kevineger: Consider definine the stub in the `describe` block and assigning it in a `beforeEach` as the other tests do.

Then the test body is nice and slim with only the stub setup (`assignmentManager.WithArgs(...).resolves(...)`).

src/colab/client.unit.test.ts
```diff
@@ -157,6 +158,26 @@ describe("ColabClient", () => {
...
+    fetchStub
+      .withArgs(
+        matchAuthorizedRequest(path, "POST", {
+          "X-Goog-Colab-Token": token,
+        }),
+      )
+      .resolves(new Response(undefined, { status: 200 }));
+    await expect(client.unassign(endpoint)).to.eventually.be.fulfilled;
+    sinon.assert.calledTwice(fetchStub);
+  });
```
kevineger: Nit: here too.

```suggestion
  it("successfully unassigns the specified assignment", async () => {
    const endpoint = "mock-endpoint";
    const path = `tun/m/unassign/${endpoint}`;
    const token = "mock-xsrf-token";
    fetchStub
      .withArgs(matchAuthorizedRequest(path, "GET"))
      .resolves(
        new Response(withXSSI(JSON.stringify({ token })), { status: 200 }),
      );
    fetchStub
      .withArgs(
        matchAuthorizedRequest(path, "POST", {
          "X-Goog-Colab-Token": token,
        }),
      )
      .resolves(new Response(undefined, { status: 200 }));

    await expect(client.unassign(endpoint)).to.eventually.be.fulfilled;
    
    sinon.assert.calledTwice(fetchStub);
  });
```

src/colab/client.ts
```diff
@@ -130,6 +151,39 @@ export class ColabClient {
...
+    );
+  }
+
+  /**
+   * Deletes the given session
+   *
+   * @param endpoint - The endpoint to delete the session from.
+   * @param sessionId - The ID of the session to delete.
+   */
+  async deleteSession(endpoint: string, sessionId: string) {
```
kevineger: For consistency, accept a `signal?: AbortSignal` and pass it along.

amtoney524: whoops, nice catch!

src/colab/server-commands.unit.test.ts
amtoney524: done, thanks for the tip!

src/jupyter/assignments.ts
amtoney524: Dropped a note in the TS Doc and maybe this is also a good case for the notification follow ups (e.g., something fails and we don't call unassign, we could consider popping a toast)

src/jupyter/assignments.unit.test.ts
amtoney524: yes ^^ nice catch!

src/jupyter/assignments.unit.test.ts
amtoney524: done, here and throughout!

src/jupyter/assignments.unit.test.ts
amtoney524: you're right, removed this one in favor of the deletes session cases

src/colab/server-commands.unit.test.ts
amtoney524: dropped a todo regarding constraints in `server-commands`! I'll work on that in the next PR and address the tests over there as well - thanks for sharing your thoughts on this :)

src/colab/client.unit.test.ts
amtoney524: done!

src/jupyter/assignments.unit.test.ts
amtoney524: removed! It wasn't clear which it blocks would be relevant since this behavior is common to all of them - happy to go back and adjust as needed

PR #65, opened 2025-04-07
feat: refresh OAuth tokens

kevineger:
Note: this is a breaking change. Since we aren't released, there's no point in making this change foreward compatible. Other contributors building the extension will need to manually run the following to clear their existing stored auth session: `await this.secrets.delete(SESSIONS_KEY);`.

Here we are now storing the refresh token and using it to ensure the access token returned on authentication sessions is valid.

For details on the Google auth client, check out the [README](https://github.com/googleapis/google-auth-library-nodejs/blob/main/README.md). To refresh the token, we could have instead leaned on the `tokens` event ([docs](https://github.com/googleapis/google-auth-library-nodejs/blob/main/README.md#handling-token-events)) the client emits when it detects a refresh is needed and does one. However, I felt it was best to explicitly control the refresh since:

- The logic is more simple than what we'd need to manage async storage updates responding to a sync event.
- We can refresh JIT when we need it, avoiding unecessary refreshes.
- This is largely consistent with what other similar extensions do (e.g. Google Cloud Code).
- If in the future we wanted to pre-refresh the token so it's always valid, we could instrument pretty basic interval logic.

As described in [Obtaining a new Refresh Token](https://github.com/googleapis/google-auth-library-nodejs/blob/main/README.md#handling-token-events), we obtain a refresh token on the first authorization by prompting for `consent` and setting the `access_type` to `offline`.

While we could fully stub out all of the `vscode.authentication` members, here I felt it was more robust and maintainable to author the thin fakes. The covered functionality is pretty trivial and it makes working with the tests a lot easier.

### DIFF
--- a/src/auth/provider.ts
+++ b/src/auth/provider.ts
@@ -22,8 +22,8 @@ const REFRESH_MARGIN_MS = 5 * 60 * 1000; // 5 minutes
  * Registers itself with the VS Code authentication API and emits events when
  * authentication sessions change.
  *
- * Access tokens for sessions are refreshed JIT when they are accessed when they
- * are close to, or past, their expiry.
+ * Session access tokens are refreshed JIT upon access if they are near or past
+ * their expiry.
  */
 export class GoogleAuthProvider
   implements vscode.AuthenticationProvider, vscode.Disposable
@@ -31,7 +31,7 @@ export class GoogleAuthProvider
   readonly onDidChangeSessions: vscode.Event<vscode.AuthenticationProviderAuthenticationSessionsChangeEvent>;
   private readonly authProvider: vscode.Disposable;
   private readonly emitter: vscode.EventEmitter<vscode.AuthenticationProviderAuthenticationSessionsChangeEvent>;
-  private session: AuthenticationSession | undefined;
+  private session?: Readonly<AuthenticationSession>;
   private isInitialized = false;
 
   /**
@@ -122,6 +122,8 @@ export class GoogleAuthProvider
 
   /**
    * Get the list of managed sessions.
+   *
+   * The session's access token is refreshed if it is near or past its expiry.
    *
    * @param scopes - An optional array of scopes. If provided, the sessions
    * returned will match these permissions. Otherwise, all sessions are
@@ -337,9 +339,7 @@ export class GoogleAuthProvider
 
   private assertInitialized(): void {
     if (!this.isInitialized) {
-      throw new Error(
-        `${this.constructor.name} has not been initialized. Call initialize() first.`,
-      );
+      throw new Error(`Must call initialize() first.`);
     }
   }
 }
--- a/src/auth/provider.unit.test.ts
+++ b/src/auth/provider.unit.test.ts
@@ -117,6 +117,9 @@ describe("GoogleAuthProvider", () => {
   describe("lifecycle", () => {
     it('registers the "Google" authentication provider', async () => {
       await authProvider.initialize();
+      // Expect the provider-specific rejection surrounding the scopes not
+      // matching the required set. This validates that the provider was
+      // registered and is being used.
       await expect(
         vsCodeStub.authentication.getSession(PROVIDER_ID, [
           "make",
@@ -137,13 +140,13 @@ describe("GoogleAuthProvider", () => {
     it("is not functional until initialized", async () => {
       await expect(
         authProvider.getSessions(undefined, {}),
-      ).to.eventually.be.rejectedWith(/Call initialize/);
+      ).to.eventually.be.rejectedWith(/call initialize/);
       await expect(
         authProvider.createSession([]),
-      ).to.eventually.be.rejectedWith(/Call initialize/);
+      ).to.eventually.be.rejectedWith(/call initialize/);
       await expect(
         authProvider.removeSession(""),
-      ).to.eventually.be.rejectedWith(/Call initialize/);
+      ).to.eventually.be.rejectedWith(/call initialize/);
     });
 
     describe("initialize", () => {
@@ -340,9 +343,7 @@ describe("GoogleAuthProvider", () => {
           const externalCallbackUri = `vscode://google.colab?nonce%3D${nonce}%26windowId%3D1`;
           vsCodeStub.env.asExternalUri
             .withArgs(matchUri(callbackUri))
-            .callsFake((_uri) =>
-              Promise.resolve(vsCodeStub.Uri.parse(externalCallbackUri)),
-            );
+            .resolves(vsCodeStub.Uri.parse(externalCallbackUri));
           return Promise.resolve(code);
         });
     });
@@ -384,9 +385,7 @@ describe("GoogleAuthProvider", () => {
     });
 
     function matchUri(regExp: RegExp) {
-      return sinon.match((uri: vscode.Uri) => {
-        return regExp.test(uri.toString());
-      });
+      return sinon.match((uri: vscode.Uri) => regExp.test(uri.toString()));
     }
 
     describe("with a successful login", () => {
--- a/src/colab/client.ts
+++ b/src/colab/client.ts
@@ -39,7 +39,7 @@ export class ColabClient {
 
   constructor(
     private readonly domain: URL,
-    private accessToken: () => Promise<string>,
+    private readonly getAccessToken: () => Promise<string>,
   ) {
     // TODO: Temporary workaround to allow self-signed certificates
     // in local development.
@@ -201,7 +201,7 @@ export class ColabClient {
     schema?: T,
   ): Promise<z.infer<T>> {
     endpoint.searchParams.append("authuser", "0");
-    const token = await this.accessToken();
+    const token = await this.getAccessToken();
     const requestHeaders = new fetch.Headers(init.headers);
     requestHeaders.set("Accept", "application/json");
     requestHeaders.set("Authorization", `Bearer ${token}`);


src/auth/provider.ts
```diff
@@ -1,28 +1,38 @@
...
 /**
  * Provides authentication using Google OAuth2.
  *
- * Registers itself with the VS Code authentication API and emits events
- * when authentication sessions change.
+ * Registers itself with the VS Code authentication API and emits events when
+ * authentication sessions change.
+ *
+ * Access tokens for sessions are refreshed JIT when they are accessed when they
+ * are close to, or past, their expiry.
```
zelliott: This comment reads clunky, maybe something like: _"Session access tokens are refreshed JIT upon access if they are near or past their expiry."_

src/auth/provider.ts
```diff
@@ -1,28 +1,38 @@
...
  */
 export class GoogleAuthProvider
   implements vscode.AuthenticationProvider, vscode.Disposable
 {
   readonly onDidChangeSessions: vscode.Event<vscode.AuthenticationProviderAuthenticationSessionsChangeEvent>;
   private readonly authProvider: vscode.Disposable;
   private readonly emitter: vscode.EventEmitter<vscode.AuthenticationProviderAuthenticationSessionsChangeEvent>;
+  private session: AuthenticationSession | undefined;
```
zelliott: nit: Consider `private session?: AuthenticationSession;`

src/auth/provider.ts
```diff
@@ -230,6 +334,41 @@ export class GoogleAuthProvider
     const json: unknown = await response.json();
     return UserInfoSchema.parse(json);
   }
+
+  private assertInitialized(): void {
+    if (!this.isInitialized) {
+      throw new Error(
+        `${this.constructor.name} has not been initialized. Call initialize() first.`,
```
zelliott: Are you minifying any of your code? If so, `this.constructor.name` might be something horrible like `Hbz` then.

src/auth/provider.ts
```diff
@@ -142,22 +211,54 @@ export class GoogleAuthProvider
   /**
    * Removes a session by ID.
    *
+   * This will revoke the credentials (if the matching session is managed) and
+   * remove the session from storage.
+   *
    * @param sessionId - The session ID.
    */
   async removeSession(sessionId: string): Promise<void> {
```
zelliott, line 221: This PR introduces a single `session` member to the class, but its methods still communicate that it manages multiple sessions (e.g. `getSessions`, `removeSession(sessionId)`).

Might this provider eventually manage multiple sessions in the future, or does it make sense to refactor its methods to communicate that there's really only a single session?

src/auth/provider.ts
```diff
@@ -142,22 +211,54 @@ export class GoogleAuthProvider
...
+    const expiryDateMs = this.oAuth2Client.credentials.expiry_date;
+    if (expiryDateMs && expiryDateMs > Date.now() + REFRESH_MARGIN_MS) {
+      return;
+    }
+    await this.oAuth2Client.refreshAccessToken();
+    const accessToken = this.oAuth2Client.credentials.access_token;
+    if (!accessToken) {
+      throw new Error("Failed to refresh Google OAuth token.");
+    }
+    this.session = {
```
zelliott, line 256: Could it make sense to make `session` immutable? Is that something that might prevent nasty bugs down the road?

src/auth/provider.ts
```diff
@@ -230,6 +334,41 @@ export class GoogleAuthProvider
...
+
+function isDefinedCredentials(
+  credentials: Credentials,
+): credentials is DefinedCredentials {
+  return (
+    credentials.refresh_token != null &&
+    credentials.access_token != null &&
+    credentials.expiry_date != null &&
+    credentials.scope != null
+  );
```
zelliott, line 371: Just curious, why no zod here?

src/auth/provider.unit.test.ts
```diff
@@ -5,125 +5,321 @@ import {
...
   publisher: PROVIDER_ID,
   name: "colab",
 };
-const REQUIRED_SCOPES = ["profile", "email"];
 const CLIENT_ID = "testClientId";
-const DEFAULT_SESSION: vscode.AuthenticationSession = {
+const SCOPES = Array.from(REQUIRED_SCOPES);
+const NOW = Date.now();
+const HOUR_MS = 60 * 60 * 1000;
+const DEFAULT_ACCESS_TOKEN = "42";
```
zelliott, line 27: 👀

src/auth/provider.unit.test.ts
```diff
@@ -5,125 +5,321 @@ import {
...
+    it('registers the "Google" authentication provider', async () => {
+      await authProvider.initialize();
+      await expect(
+        vsCodeStub.authentication.getSession(PROVIDER_ID, [
+          "make",
+          "it",
+          "error",
+        ]),
+      ).to.eventually.be.rejectedWith(/scopes/);
     });
```
zelliott: Sorry, a little confused by this test. It's not clear to me how this test is asserting that the auth provider is registered. It looks like the test is expecting `getSession` to be rejected. What am I missing?

src/auth/provider.unit.test.ts
```diff
@@ -133,158 +329,205 @@ describe("GoogleAuthProvider", () => {
...
+        .withArgs(sinon.match(isUUID), cancel)
+        .callsFake((n, _token) => {
+          nonce = n;
+          const callbackUri = new RegExp(
+            `vscode://google.colab\\?nonce=${nonce}`,
+          );
+          const externalCallbackUri = `vscode://google.colab?nonce%3D${nonce}%26windowId%3D1`;
+          vsCodeStub.env.asExternalUri
+            .withArgs(matchUri(callbackUri))
+            .callsFake((_uri) =>
```
zelliott: Does sinon have a `.resolves` you can use here?

```ts
.resolves(vsCodeStub.Uri.parse(externalCallbackUri)),
```

src/auth/provider.unit.test.ts
```diff
@@ -133,158 +329,205 @@ describe("GoogleAuthProvider", () => {
...
+      ).to.eventually.be.rejectedWith(/credential information/);
+      sinon.assert.calledOnceWithMatch(
+        vsCodeStub.window.showErrorMessage,
+        sinon.match(/Sign in failed.+/),
+      );
+    });
+
+    function matchUri(regExp: RegExp) {
+      return sinon.match((uri: vscode.Uri) => {
+        return regExp.test(uri.toString());
```
zelliott: nit: Consider shorthand return.

src/auth/provider.unit.test.ts
```diff
@@ -133,158 +329,205 @@ describe("GoogleAuthProvider", () => {
...
+      });
+    }
+
+    describe("with a successful login", () => {
+      beforeEach(() => {
+        sinon
+          .stub(oauth2Client, "getToken")
+          .withArgs({ code, codeVerifier: sinon.match.string })
+          .resolves({
```
zelliott, line 396: Ah, it does! (re my comment above about `.resolves`)

src/colab/client.ts
```diff
@@ -40,7 +39,7 @@ export class ColabClient {
 
   constructor(
     private readonly domain: URL,
-    private session: () => Promise<AuthenticationSession>,
+    private accessToken: () => Promise<string>,
```
zelliott: I could check but I'm a little lazy, can this be `readonly`? I also think it might be more idiomatic TS to call this `getAccessToken` (I know this is a bit preexisting from `session` instead of `getSession` in the before diff).

src/jupyter/assignments.ts
```diff
@@ -246,3 +250,20 @@ function colabProxyFetch(
...
+    ineligible?: Set<Accelerator>;
+  },
+): boolean {
+  if (!accelerator) {
+    return false;
+  }
+  return (
+    !availability.ineligible?.has(accelerator) &&
+    availability.eligible.has(accelerator)
+  );
```
zelliott: Why do we check if the accelerator is both _not ineligible_ and _eligible_? Why doesn't one imply the other?

src/extension.ts
```diff
@@ -38,8 +38,11 @@ export async function activate(context: vscode.ExtensionContext) {
     AUTH_CLIENT,
     redirectUriHandler,
   );
+  await authProvider.initialize();
```
zelliott, line 41: If this fails, what handles that error? What state is the extension in?

src/auth/provider.ts
```diff
@@ -79,56 +121,83 @@ export class GoogleAuthProvider
...
+    scopes: readonly string[] | undefined,
+    options: vscode.AuthenticationProviderSessionOptions,
   ): Promise<vscode.AuthenticationSession[]> {
-    const session = await this.storage.getSession();
-    return session ? [session] : [];
+    this.assertInitialized();
+    if (scopes && !matchesRequiredScopes(scopes)) {
+      return [];
+    }
+    await this.refreshSessionIfNeeded();
```
zelliott, line 143: Does it make sense to document this side-effect in the JSDoc comment? A caller might use `getSessions` differently if they know that it's refreshing under-the-hood.

src/auth/provider.ts
```diff
@@ -79,56 +121,83 @@ export class GoogleAuthProvider
...
+    }
+    return this.session ? [this.session] : [];
   }
 
   /**
    * Creates and stores an authentication session with the given scopes.
    *
-   * @param scopes - Scopes required for the session.
+   * @param scopes - Scopes required for the session. These must strictly be the
+   * collection of {@link REQUIRED_SCOPES}.
```
zelliott, line 154: If this is the requirement, could you type it as `typeof REQUIRED_SCOPES`?

And also, sorry, do you expect this class to support scopes other than `REQUIRED_SCOPES` in the future? I'm guessing so, otherwise you'd just not have the arg altogether.

src/jupyter/assignments.ts
kevineger: This change went in with https://github.com/googlecolab/colab-vscode/pull/63, sorry I hadn't rebased to `origin/main` before mailing PR.

But it's due to how we handle free vs. pro tiers (enables the ability to upsell Pro/Pro+).

src/extension.ts
kevineger: The extension fails to activate. It won't work in any capacity.

VS Code will throw up an error notification to the effect of _Colab failed to activate..._.

src/colab/client.ts
kevineger: Done and done.

src/auth/provider.unit.test.ts
kevineger: Done.

src/auth/provider.unit.test.ts
kevineger: 🌍+🪐+♾️ = 4️⃣2️⃣

src/auth/provider.unit.test.ts
kevineger: If it weren't registered, we wouldn't get the provider-specific failure of it being rejected because the scopes don't match. Here I'm trying to test behaviour strictly through public APIs on the SUT. Calling `vsCodeStub.authentication.getSession` validates that the provider was registered.

I'll add a comment to the test since that does require some implied knowledge.

src/auth/provider.unit.test.ts
kevineger: It does. Sorry, smell of a previous approach which used the provided `uri` param.

src/auth/provider.ts
kevineger: Because we're not parsing anything. We're narrowing the definition of a TS interface.

src/auth/provider.ts
kevineger: Done.

src/auth/provider.ts
kevineger: Sure, I'll define it as `Readonly<AuthenticationSession>`.

src/auth/provider.ts
kevineger: The class `implements vscode.AuthenticationProvider`, so it has to be defined as handling multiple sessions. I can't _refactor its methods_.

In the future, we will support multiple once we have something like Drive support (where the user may start without it, then up-scope to Drive).

src/auth/provider.ts
kevineger: Done.

src/auth/provider.ts
kevineger: Done.

src/auth/provider.ts
kevineger: See other comment about this implementing the VS Code interface and the likely addition of Drive in the future.

src/auth/provider.ts
kevineger: Done.

PR #66, opened 2025-04-07
feat: recaptcha

lisayin:
feat: Implements recaptcha flow.


package.json
```diff
@@ -19,6 +19,13 @@
     "onNotebook:interactive"
   ],
   "contributes": {
+    "commands": [
+      {
+        "command": "recaptcha.showWebView",
+        "title": "Google Recaptcha",
+        "category": "Cat Coding"
```
katlyn-edwards: Heheh I think we want to drop this line

More curious: I wonder if we want to drop this whole command block? I don't know what we want users to trigger this on their own.

src/colab/client.ts
```diff
@@ -37,16 +39,21 @@ interface AssignedAssignment extends Assignment {
  */
 export class ColabClient {
   private readonly httpsAgent?: https.Agent;
+  private readonly recaptchaService?: RecaptchaService;
+  private readonly recaptchaWebview?: RecaptchaWebview;
```
katlyn-edwards: You can drop this line -- the `private readonly recaptchaWebView`in the constructor should be sufficient!

src/colab/client.ts
```diff
@@ -37,16 +39,21 @@ interface AssignedAssignment extends Assignment {
...
     private session: () => Promise<AuthenticationSession>,
+    private readonly recaptchaWebView: RecaptchaWebview
   ) {
     // TODO: Temporary workaround to allow self-signed certificates
     // in local development.
     if (domain.hostname === "localhost") {
       this.httpsAgent = new https.Agent({ rejectUnauthorized: false });
     }
+    this.recaptchaService = new RecaptchaService();
+    this.recaptchaWebview = recaptchaWebView;
```
katlyn-edwards: Ditto here -- since you're already declaring it in the constructor parameters, you can drop this!

src/colab/client.ts
```diff
@@ -37,16 +39,21 @@ interface AssignedAssignment extends Assignment {
...
     private readonly domain: URL,
     private session: () => Promise<AuthenticationSession>,
+    private readonly recaptchaWebView: RecaptchaWebview
   ) {
     // TODO: Temporary workaround to allow self-signed certificates
     // in local development.
     if (domain.hostname === "localhost") {
       this.httpsAgent = new https.Agent({ rejectUnauthorized: false });
     }
+    this.recaptchaService = new RecaptchaService();
```
katlyn-edwards: I wonder if you can put this up on line 43

I'd try dropping this line, then changing 43 to be:
`private readonly recaptchaService: new RecpatchaService();`

src/colab/client.ts
```diff
@@ -91,13 +98,35 @@ export class ColabClient {
         return rest;
       }
       case "to_assign": {
-        return await this.postAssignment(
-          notebookHash,
+        console.log("????????????????????????", this.recaptchaService)
+        if (!this.recaptchaService) {
```
katlyn-edwards: LOL, You can drop this if you're able to inline the proposal to 43, because it won't be "nullable"

src/colab/client.ts
```diff
@@ -91,13 +98,35 @@ export class ColabClient {
...
+          throw new Error("?????????????????")
+        }
+        // prob need to await on this?
+        const promise = new Promise(() => {
```
katlyn-edwards: I think we can drop this, and "await" the this.recaptchaWebView.show() command -- which will async send back a token.

So maybe I'd do:
```
const token = await this.recaptchaWebView.show();
```

src/extension.ts
```diff
@@ -54,10 +60,16 @@ export async function activate(context: vscode.ExtensionContext) {
     jupyter,
   );
 
+
+  const command = vscode.commands.registerCommand('recaptcha.showWebView', () => {
```
katlyn-edwards: I think we can drop this -- re: no needing for users to trigger this.

src/recaptcha/recaptcha-v3.ts
```diff
@@ -0,0 +1,133 @@
...
+import {trustedResourceUrl} from 'safevalues';
+import {setScriptSrc} from 'safevalues/dom';
+import * as vscode from 'vscode';
+
+//import {uuid} from 'uuid'
+const V3SiteKey = "6LfQPtEUAAAAAHBpAdFng54jyuB1V5w5dofknpip"
+
+export class RecaptchaService extends EventTarget {
+    //private readonly loadPromises: Promise<void>[] = [];
+    constructor() {
```
katlyn-edwards: I suspect you can just drop this whole boy (unless it's useful!)

src/recaptcha/recaptcha-v3.ts
```diff
@@ -0,0 +1,133 @@
...
+//    if (!token) {
+//      throw new Error('v3: No token found!');
+//    }
+//    return token;
+//    } else {
+//        throw new Error('grecaptcha never loaded')
+//    }
+//  }
+
+  private generateUniqueCallback(): RecaptchaPublicCallback {
```
katlyn-edwards: Maybe comment this out? I don't see it being used (yet)

lisayin:
would love some help on how to test the webview I'm pretty lost, but I think initial pass on the feature is ready for review?

src/colab/client.ts
```diff
@@ -244,8 +246,18 @@ export class ColabClient {
     variant: Variant,
     accelerator?: Accelerator,
     signal?: AbortSignal,
+    shouldPromptRecaptcha?: boolean,
```
kevineger, line 249: Should this be required?

src/colab/client.ts
```diff
@@ -270,6 +283,9 @@ export class ColabClient {
     if (accelerator) {
       url.searchParams.append("accelerator", accelerator.toString());
     }
+    if (recaptchaResponse) {
+      url.searchParams.append("recaptcha_v3_response", recaptchaResponse);
+    }
```
kevineger, line 288: This branch only happens when posting an assignment. Could we instead leave `buildAssignUrl` to just be the common parts between GET and POST and instead append this search param to the URL in the `postAssignment` method?

src/colab/client.ts
```diff
@@ -244,8 +246,18 @@ export class ColabClient {
     variant: Variant,
     accelerator?: Accelerator,
     signal?: AbortSignal,
+    shouldPromptRecaptcha?: boolean,
   ): Promise<Assignment> {
-    const url = this.buildAssignUrl(notebookHash, variant, accelerator);
+    let recaptchaToken = "";
+    if (shouldPromptRecaptcha) {
+      recaptchaToken = await this.getRecaptchaToken();
```
kevineger, line 253: Is POST `/assignment` the only endpoint that needs this? Would refreshing the token? Anything else?

src/colab/client.unit.test.ts
```diff
@@ -184,6 +188,46 @@ describe("ColabClient", () => {
     });
   });
 
+  it("calls for recaptcha if its required when getting a new assignment", async () => {
```
kevineger, line 191: Nit: test behaviour, not implementation (and choose a name which reflects it). The fact that it "calls for recaptcha" is an implementation detail.

```suggestion
  it("includes recaptcha token when needed for new assignements", async () => {
```

src/colab/client.unit.test.ts
```diff
@@ -428,10 +472,14 @@ function matchAuthorizedRequest(
   method: "DELETE" | "GET" | "POST",
   otherHeaders?: Record<string, string>,
   withAuthUser = true,
+  withRecaptcha = false,
```
kevineger, line 475: Now that there are two of these optional query param assertions, we should consider a structure that isn't just a collection of booleans.

Either (a) a type where they're named (`{withAuthUser: boolean, withRecaptcha: boolean}`), (b) a spread enum or (c) separate the call-site specific matching to the relevant call-sites (in other words, have `matchAuthorizedRequest` only do exactly that and do the Recaptcha matching in that test).

src/extension.ts
```diff
@@ -43,14 +44,21 @@ export async function activate(context: vscode.ExtensionContext) {
...
   const colabClient = new ColabClient(
     new URL(COLAB_DOMAIN),
     new URL(COLAB_GAPI_DOMAIN),
     () =>
       GoogleAuthProvider.getOrCreateSession(vscode).then(
         (session) => session.accessToken,
       ),
+    () => {
+      recaptcha.show();
+      return recaptcha.sendRequestAndWaitForResponse("requestRecaptcha");
```
kevineger, line 58: We only ever call this method with this argument value. Hardcode/const it in the `RecaptchaWebview` class?

package.json
```diff
@@ -101,6 +101,7 @@
     "glob": "^11.0.0",
     "google-auth-library": "^9.15.0",
     "node-fetch": "^2.6.7",
+    "recaptcha-v3": "^1.11.3",
```
kevineger, line 104: Mind listing how/why you chose [this package](https://www.npmjs.com/package/recaptcha-v3)? I haven't looked into it much but seems like there are a few options out there.

[@google-cloud/recaptcha-enterprise](https://www.npmjs.com/package/@google-cloud/recaptcha-enterprise) seems a lot more used? Sorry if overlooking some constraint or reason why we couldn't use it.

project-words.txt
```diff
@@ -15,4 +15,6 @@ unassignment
 unassigns
 veryhighmem
 vulnz
-xssi
\ No newline at end of file
+xssi
+grecaptcha
+recaptcha
```
kevineger, line 20: Nit: sort lines ascending.

src/recaptcha/recaptcha-webview.unit.test.ts
```diff
@@ -0,0 +1,71 @@
...
+const RECAPTCHA_TOKEN = "recaptcha-token";
+
+describe("RecaptchaWebview", () => {
+  let mockContext: Partial<vscode.ExtensionContext>; // Use Partial to allow mocking only necessary properties
+  let webview: Partial<vscode.Webview>;
+  let webviewPanel: Partial<vscode.WebviewPanel>;
+  let vsCodeStub: VsCodeStub;
+  let recaptchaWebview: RecaptchaWebview;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let onDidReceiveMessageEventEmitter: TestEventEmitter<any>;
```
kevineger, line 17: Please avoid using `any` and type this as needed. Apply globally.

src/recaptcha/recaptcha-webview.unit.test.ts
```diff
@@ -0,0 +1,71 @@
...
+import { RecaptchaWebview } from "./recaptcha-webview";
+
+const RECAPTCHA_TOKEN = "recaptcha-token";
+
+describe("RecaptchaWebview", () => {
+  let mockContext: Partial<vscode.ExtensionContext>; // Use Partial to allow mocking only necessary properties
+  let webview: Partial<vscode.Webview>;
+  let webviewPanel: Partial<vscode.WebviewPanel>;
+  let vsCodeStub: VsCodeStub;
+  let recaptchaWebview: RecaptchaWebview;
```
kevineger, line 15: Nit: scope ordering.

```suggestion
  let vsCodeStub: VsCodeStub;
  let mockContext: Partial<vscode.ExtensionContext>;
  let webview: Partial<vscode.Webview>;
  let webviewPanel: Partial<vscode.WebviewPanel>;
  let recaptchaWebview: RecaptchaWebview;
```

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
+//import * as vscode from "vscode";
```
kevineger, line 1: Delete

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+import vscode from "vscode";
+
+const V3SiteKey = "6LfQPtEUAAAAAHBpAdFng54jyuB1V5w5dofknpip";
```
kevineger, line 4: Where does this value come from? Should it be configurable? Is it "secret"?

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export class RecaptchaWebview {
```
kevineger, line 6: Document public classes (TSDoc).

Internal guidance: go/tsjs-style#jsdoc-class-comments

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+  private panel: vscode.WebviewPanel | undefined;
```
kevineger, line 7: Nit

```suggestion
  private panel?: vscode.WebviewPanel;
```

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+  private nextResponseId = 0; // I think theres an equivalent to this when they're doing generateUniqueCallback
```
kevineger, line 8: Did you mean to circle back to this and/or remove/update comment?

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+  private responsePromises: Record<string, (resolve: string) => void> = {};
```
kevineger, line 9: Nit: name should reflect that this maps response IDs to the response resolver.

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  constructor(
+    private readonly vs: typeof vscode,
+    private context: vscode.ExtensionContext,
+  ) {}
+
+  sendRequestAndWaitForResponse(command: string): Promise<string> {
+    return new Promise((resolve) => {
+      const responseId = String(this.nextResponseId++);
```
kevineger, line 18: I'd considering using `uuid`-s instead of an incrementing `number`.

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+        },
+      );
+      this.panel.webview.html = this.getWebviewContent();
+      this.panel.webview.onDidReceiveMessage(
+        (message) => {
+          /* eslint-disable @typescript-eslint/no-unsafe-member-access */
+          if (
+            message.responseId &&
+            message.responseId in this.responsePromises
+          ) {
```
kevineger, line 50: Please address unsafe member access. Can use a type-guard.

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+            // Resolve the recaptcha promise.
```
kevineger, line 51: Nit: comment would be redundant with a naming change to `responsePromises` (e.g. `responseResolvers`).

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  show() {
+    // Track the current panel with a webview
+    const columnToShowIn = this.vs.window.activeTextEditor
+      ? this.vs.window.activeTextEditor.viewColumn
+      : undefined;
+
+    if (this.panel) {
+      // If we already have a panel, show it in the target column
+      this.panel.reveal(columnToShowIn);
```
kevineger, line 32: Sorry I'm struggling a bit understanding the lifecycle here.

For the extension, we have one instance of this class. What happens if we have two overlapping new assignments, which both need Recaptcha-ing?

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+import vscode from "vscode";
+
+const V3SiteKey = "6LfQPtEUAAAAAHBpAdFng54jyuB1V5w5dofknpip";
+
+export class RecaptchaWebview {
+  private panel: vscode.WebviewPanel | undefined;
+  private nextResponseId = 0; // I think theres an equivalent to this when they're doing generateUniqueCallback
+  private responsePromises: Record<string, (resolve: string) => void> = {};
+
+  constructor(
```
kevineger, line 11: Shouldn't this class also be `vscode.Disposable`? It could have a panel which needs to be disposed. Take a look at `ColabJupyterServerProvider` for context (and how it's disposed of in `extension.ts`).

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+      );
+
+      // Reset when the current panel is closed
+      this.panel.onDidDispose(() => {
+        this.panel = undefined;
+      }, null);
+    }
+  }
+
+  private getWebviewContent(): string {
```
kevineger, line 73: I've not used webviews in an extension before. Is it possible to define this HTML template in its own file (ideally an HTML file). Then could we have the `<script>` tag reference actual JS so we get full language support and avoid issues with typos, types, etc.? Would also enable us to write the logic in TS instead of raw JS.

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+                  padding: 10px 20px;
+                  font-size: 16px;
+              }
+          </style>
+      </head>
+      <body>
+          <button id="myButton">Click Me</button>
+          <script>
+              const vscode = acquireVsCodeApi();
+              const button = document.getElementById('myButton');
```
kevineger, line 91: Could we use meaningful `id`-s?

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+          <script src="https://www.google.com/recaptcha/api.js?render=${V3SiteKey}" async defer></script>
+          <style>
+              button {
+                  padding: 10px 20px;
+                  font-size: 16px;
+              }
+          </style>
+      </head>
+      <body>
+          <button id="myButton">Click Me</button>
```
kevineger, line 88: I suspect this needs updating?

src/recaptcha/recaptcha-webview.unit.test.ts
```diff
@@ -0,0 +1,71 @@
+import { expect } from "chai";
+import * as sinon from "sinon";
+import vscode from "vscode";
+import { TestEventEmitter } from "../test/helpers/events";
+import { newVsCodeStub, VsCodeStub } from "../test/helpers/vscode";
+import { RecaptchaWebview } from "./recaptcha-webview";
+
+const RECAPTCHA_TOKEN = "recaptcha-token";
+
+describe("RecaptchaWebview", () => {
```
kevineger, line 10: Could we add coverage for the non-happy-path case(s)?

src/recaptcha/recaptcha-webview.ts
```diff
@@ -0,0 +1,116 @@
...
+                  if (button) {
+                    button.addEventListener('click', () => {
+                      grecaptcha.ready(function() {
+                        grecaptcha.execute('${V3SiteKey}', { action: 'submit' }).then(function(token) {
+                          vscode.postMessage({
+                            responseId: messageResponseId,
+                            token: token
+                          });
+                        }).catch(e => {
+                          console.log("grecaptcha errored with ", e);
```
kevineger, line 105: Typo. Should we be notifying to the user? Failing to Recaptcha will eventually disconnect them, right?

src/recaptcha/recaptcha-webview.unit.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  it("should create a webview panel with the correct title and view type", () => {
+    recaptchaWebview.show();
+    sinon.assert.calledOnce(vsCodeStub.window.createWebviewPanel);
+    expect(vsCodeStub.window.createWebviewPanel.firstCall.args[1]).to.equal(
+      "Colab Recaptcha",
+    );
+  });
+
+  it("returns the message back", async () => {
```
kevineger, line 65: Could we get some tests interacting with the webview?

package.json
lisayin: Katlyn helped me find it but I don't think we're using the enterprise recaptcha?
https://pantheon.corp.google.com/security/recaptcha/6LfQPtEUAAAAAHBpAdFng54jyuB1V5w5dofknpip/overview?e=13803378&inv=1&invt=Abwm2g&mods=dm_deploy_from_gcs&project=colab-recaptcha
different from 
https://pantheon.corp.google.com/apis/api/recaptchaenterprise.googleapis.com/metrics?inv=1&invt=Abwm2A&project=colab-recaptcha


This is just the regular one afaik? https://source.corp.google.com/piper///depot/google3/research/colab/frontend/external/recaptcha/recaptcha.ts;l=61

src/recaptcha/recaptcha-webview.ts
lisayin: It shouldn't be configurable but I don't think it's secret either? It's our key for recaptcha: https://screenshot.googleplex.com/BKLWRmGLCyP9F93

src/recaptcha/recaptcha-webview.ts
lisayin: Added a todo to add it in a different PR.

kevineger:
We're going to hold off on requiring reCAPTCHA for now. Can re-open PR in the future if we need.

PR #68, opened 2025-04-07
feat: add a user-info accessor to the Colab Google API

kevineger:
As with the existing API surface... the Colab backend APIs are a little funky! The team's in the process of standardizing, normalizing and cleaning up the API surface. A big part of this is moving our APIs to One Platform (Google APIs, or GAPIs for short). In the extension, we need to know if the user is a pro/pro+ subscriber or not. Our existing frontend gets this through some funky side-loaded bits of information we don't have access to in the extension. Therefore, we need to integrate with the new GAPI for retrieving this user info.

I've done my best to compartmentalize all the funky API intricacies into `api.ts`, `transform`-ing and _normalizing_ across the two backends so the rest of the client code can be blissfully unaware of the differences.

Currently, this API requires a Drive-scoped OAuth token. This is due to some internal details that don't really matter, but it is a temporary requirement and we'll remove it before launch (hence the TODO).

### DIFF
--- a/src/colab/api.ts
+++ b/src/colab/api.ts
@@ -1,5 +1,5 @@
 /**
- * API types for interacting with Colab's backend.
+ * API types for interacting with Colab's backends.
  *
  * This file not only defines the entire Colab API surface area, but also tries
  * to compartmentalize a lot of the funky intricacies:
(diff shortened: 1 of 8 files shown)


src/colab/api.ts
```diff
@@ -87,6 +106,72 @@ function uppercaseEnum<T extends z.EnumLike>(
...
+      z.object({
+        /** The variant of the assignment. */
+        variant: z.nativeEnum(ColabGapiVariant).transform(normalizeVariant),
+        /** The assigned accelerator. */
+        models: z.array(z.nativeEnum(Accelerator)),
+      }),
+    )
+    .optional(),
+});
+/**
```
metrizable, line 170: nit: blank line prior?

kevineger: I've grouped the `*Schema` objects to their corresponding infered type definition.

PR #70, opened 2025-04-11
feat: enable server commands when assigned servers are present

amtoney524:
This change ensures the `Colab: Rename Server Alias` and `Colab: Remove Server` commands are only available to the user when there are assigned servers.
* Adds a constraint to those commands, enabling them when the `colab.hasAssignedServer` [context key](https://code.visualstudio.com/api/references/when-clause-contexts#add-a-custom-when-clause-context) is truthy
* Sets the context key on activation and in response to the `assignmentsChange` event
* While here, also refactored  `server-commands.unit.test.ts` to include separate blocks for when servers are assigned and when no servers are assigned

constraints

### DIFF
--- a/src/colab/server-commands.unit.test.ts
+++ b/src/colab/server-commands.unit.test.ts
@@ -62,7 +62,7 @@ describe("Server Commands", () => {
       serverStorageStub = sinon.createStubInstance(ServerStorage);
     });
 
-    it("when no servers are assigned the QuickPick is not rendered", async () => {
+    it("does not open the Quick Pick when no servers are assigned", async () => {
       serverStorageStub.list.resolves([]);
 
       await renameServerAlias(vsCodeStub.asVsCode(), serverStorageStub);
@@ -108,7 +108,6 @@ describe("Server Commands", () => {
 
         it("updates the server alias", async () => {
           serverStorageStub.list.resolves([defaultServer]);
-          serverStorageStub.store.resolves();
           const rename = renameServerAlias(
             vsCodeStub.asVsCode(),
             serverStorageStub,
@@ -132,7 +131,6 @@ describe("Server Commands", () => {
 
         it("does not update the server alias when it is unchanged", async () => {
           serverStorageStub.list.resolves([defaultServer]);
-          serverStorageStub.store.resolves();
           const rename = renameServerAlias(
             vsCodeStub.asVsCode(),
             serverStorageStub,
@@ -162,7 +160,7 @@ describe("Server Commands", () => {
       assignmentManagerStub = sinon.createStubInstance(AssignmentManager);
     });
 
-    it("when no servers are assigned the QuickPick is not rendered", async () => {
+    it("does not open the Quick Pick when no servers are assigned", async () => {
       assignmentManagerStub.getAssignedServers.resolves([]);
 
       await removeServer(vsCodeStub.asVsCode(), assignmentManagerStub);


package.json
```diff
@@ -36,7 +36,19 @@
         "command": "colab.removeServer",
         "title": "Remove Server"
       }
-    ]
+    ],
+    "menus": {
```
kevineger: Should we instead be using the `commands`' `enablement` property? If someone key-bound the command, they'd still be able to drop into it by just removing the menu entry, no? Sorry if wrong - haven't played much with these.

https://code.visualstudio.com/api/extension-guides/command#enablement-of-commands

src/colab/server-commands.ts
```diff
@@ -9,7 +9,6 @@ import { PROMPT_SERVER_ALIAS, validateServerAlias } from "./server-picker";
  * assigned Colab server.
  */
 // TODO: Consider adding a notification that the rename was successful.
-// TODO: Add constraint for when there are no servers assigned.
```
kevineger, line 12: With the constraint added, should these functions be a no-op when there are no `servers`? Currently, won't they pop open the quick-picks with no options?

src/colab/server-commands.unit.test.ts
```diff
@@ -56,95 +56,100 @@ describe("Server Commands", () => {
...
-        inputBoxStub.value = "new_alias";
-        inputBoxStub.onDidChangeValue.yield(inputBoxStub.value);
-        inputBoxStub.onDidAccept.yield();
-
-        await expect(rename).to.eventually.be.fulfilled;
-        sinon.assert.calledOnceWithExactly(serverStorageStub.store, [
-          { ...defaultServer, label: "new_alias" },
+          { label: additionalServer.label, value: additionalServer },
         ]);
       });
```
kevineger: Nit: separate arange-act-assert blocks.

```suggestion
      it("lists assigned servers for selection", async () => {
        const additionalServer = { ...defaultServer, label: "bar" };
        serverStorageStub.list.resolves([defaultServer, additionalServer]);

        void renameServerAlias(vsCodeStub.asVsCode(), serverStorageStub);
        await quickPickStub.nextShow();

        sinon.assert.calledOnce(serverStorageStub.list);
        expect(quickPickStub.items).to.eql([
          { label: defaultServer.label, value: defaultServer },
          { label: additionalServer.label, value: additionalServer },
        ]);
      });
```

src/colab/server-commands.unit.test.ts
```diff
@@ -156,33 +161,49 @@ describe("Server Commands", () => {
...
+      assignmentManagerStub.getAssignedServers.resolves([]);
 
       void removeServer(vsCodeStub.asVsCode(), assignmentManagerStub);
       await quickPickStub.nextShow();
 
-      expect(quickPickStub.items).to.eql([
-        { label: defaultServer.label, value: defaultServer },
-        { label: additionalServer.label, value: additionalServer },
-      ]);
+      expect(quickPickStub.items).to.eql([]);
```
kevineger: Oh wow... I had no idea `eql` in Chai was the same as `deepEqual` or `deep.equal`. Mind using the explicit `deep` variant as the rest of the tests do? I agree with the top note on this [SO answer](https://stackoverflow.com/a/36808692) 🤣.

src/jupyter/assignments.ts
```diff
@@ -39,18 +39,27 @@ export class AssignmentManager implements vscode.Disposable {
...
     private readonly client: ColabClient,
     private readonly storage: ServerStorage,
   ) {
     this.assignmentsChange = new vs.EventEmitter<void>();
+    this.disposables.push(this.assignmentsChange);
     this.onDidAssignmentsChange = this.assignmentsChange.event;
+    this.onDidAssignmentsChange(
+      () => this.setHasAssignedServerContext(),
+      this,
+      this.disposables,
```
kevineger: I'm personally not the biggest fan of side-effect APIs. If I'm interpretting the VS Code `Event<T>` interface correctly, providing an array of `Disposable` adds the listener's new `Disposable` to it.

Could we instead just use the return value (`Disposable`) and add it to our list which we `dispose()` of?

src/jupyter/assignments.unit.test.ts
```diff
@@ -626,6 +626,48 @@ describe("AssignmentManager", () => {
       sinon.assert.calledOnce(listener);
     });
   });
+
+  describe("onDidAssignmentsChange", () => {
```
kevineger: Nit: to be stylistically consistent with existing texts, consider moving these assertions to where the behaviour over the public API actions are made? `onDidAssignmentsChange` isn't the public API which callers invoke that exhibits this behaviour. Instead, I'd probably move the assertions to the relevant tests, e.g. `Assignment manager > unassignServer > when the server does exist > _it_ sets the has-assigned-server context to false`.

package.json
amtoney524: nice find, that works even better!

src/colab/server-commands.ts
amtoney524: having these be a no-op sounds good to me, added some early returns and updated the  relevant test cases

src/colab/server-commands.unit.test.ts
amtoney524: nice catch, done

src/colab/server-commands.unit.test.ts
amtoney524: np, updated to the deep variant!

src/jupyter/assignments.ts

src/jupyter/assignments.unit.test.ts
amtoney524: I replaced this block and added test cases to the blocks for the `reconcileAssignedServers`, `assignServer` and `unassignServer` methods, ptal

src/colab/server-commands.unit.test.ts
```diff
@@ -56,95 +56,101 @@ describe("Server Commands", () => {
...
-        const serverStorageStub: SinonStubbedInstance<ServerStorage> =
-          sinon.createStubInstance(ServerStorage, {
-            list: Promise.resolve([defaultServer]),
-          });
-        void renameServerAlias(vsCodeStub.asVsCode(), serverStorageStub);
-        await quickPickStub.nextShow();
-        quickPickStub.onDidChangeSelection.yield([
-          { label: defaultServer.label, value: defaultServer },
-        ]);
+    it("when no servers are assigned the QuickPick is not rendered", async () => {
```
kevineger: Nit: test cases should read as full sentences, including the _it_.

```suggestion
    it("does not open the quick-pick when no servers are assigned", async () => {
```

src/colab/server-commands.unit.test.ts
```diff
@@ -56,95 +56,101 @@ describe("Server Commands", () => {
...
+          expect(inputBoxStub.validationMessage).equal("");
+        });
 
-        await inputBoxStub.nextShow();
-        inputBoxStub.value = defaultServer.label;
-        inputBoxStub.onDidChangeValue.yield(inputBoxStub.value);
-        inputBoxStub.onDidAccept.yield();
+        it("updates the server alias", async () => {
+          serverStorageStub.list.resolves([defaultServer]);
+          serverStorageStub.store.resolves();
```
kevineger: I believe by default `sinon` stubs will resolve `Promise<void>` functions? May be able to remove this.

If so, consider applying globally.

src/colab/server-commands.unit.test.ts
```diff
@@ -156,33 +162,48 @@ describe("Server Commands", () => {
       assignmentManagerStub = sinon.createStubInstance(AssignmentManager);
     });
 
-    it("lists assigned servers for selection", async () => {
-      const additionalServer = { ...defaultServer, label: "bar" };
-      assignmentManagerStub.getAssignedServers.resolves([
-        defaultServer,
-        additionalServer,
-      ]);
+    it("when no servers are assigned the QuickPick is not rendered", async () => {
```
kevineger: Same naming suggestion here.

src/jupyter/assignments.unit.test.ts
```diff
@@ -280,6 +280,44 @@ describe("AssignmentManager", () => {
         sinon.assert.calledOnceWithExactly(storageStub.store, []);
       });
     });
+
+    describe("sets hasAssignedServer context", () => {
+      it("to true when at least one server is stored after reconciling", async () => {
```
kevineger, line 285: It sets the context key in more cases than just following reconciliation. We'll want to have similar test assertions for them:

- `assignServer`
- `unassignServer` (assertion when it removes the only vs. removes one of many)

kevineger: I recommend nesting the relevant assertion/`it` across all these cases (and the variants, e.g. reconcile none/single/multiple).

src/colab/server-commands.unit.test.ts
amtoney524: done!

src/colab/server-commands.unit.test.ts
amtoney524: done!

src/jupyter/assignments.unit.test.ts
amtoney524: yes! `assignServer` and `unassignServer` should have test cases for setting the context key; for reconciliation, I thought testing based on the effects of reconciliation would be a good start, though happy to rework this bit in a separate pr!

src/colab/server-commands.unit.test.ts
amtoney524: yes you're right, done!

PR #72, opened 2025-04-14
feat: add subscription watcher

kevineger:
This background watcher will be used to monitor for changes to the user's subscription tier. In the extension, this information will be used to determine what to show when CCU runs low/out, what commands to show (e.g. "Upgrade to Pro"), etc.


kevineger:
Closing, at least for now... On second thought, I may not need this watcher and can instead JIT-query for the subscription tier where I need it.

PR #73, opened 2025-04-14
feat: provide subscription-aware commands

kevineger:
When the user's already paid for pro or pro-plus, omit it from the list of Jupyter commands.

### DIFF
--- a/src/jupyter/provider.ts
+++ b/src/jupyter/provider.ts
@@ -93,32 +93,7 @@ export class ColabJupyterServerProvider
     _value: string | undefined,
     _token: CancellationToken,
   ): ProviderResult<JupyterServerCommand[]> {
-    const commands = [
-      {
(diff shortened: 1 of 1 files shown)


src/jupyter/provider.ts
```diff
@@ -99,11 +102,23 @@ export class ColabJupyterServerProvider
         label: OPEN_COLAB_WEB_LABEL,
         description: "Open Colab web.",
       },
-      {
-        label: UPGRADE_TO_PRO_LABEL,
-        description: "More machines, more quota, more Colab!",
-      },
     ];
+    return (
```
amtoney524: wdyt about getting commands from a separate function so we can use async/wait syntax? worth it or no?

kevineger: Done.

PR #78, opened 2025-04-22
fix: convert remainingTokens to a number

kevineger:
The Proto-defined field is an `Int64`. As the [docs](https://protobuf.dev/programming-guides/json) outline, `Int64`
fields are JSON serialized as strings since they can eclipse the max number size in JavaScript (`2^53 - 1`).

In practice, this value should never be more than 6,000 (the current max allowed tokens). We can safely convert, though I did add some safety to the refinement/validation largely for clarity.

### DIFF
--- a/src/colab/client.unit.test.ts
+++ b/src/colab/client.unit.test.ts
@@ -100,7 +100,7 @@ describe("ColabClient", () => {
   });
 
   it("successfully gets CCU info", async () => {
-    const mockResponse: CcuInfo = {
+    const mockResponse = {
       currentBalance: 1,
       consumptionRateHourly: 2,
       assignmentsCount: 3,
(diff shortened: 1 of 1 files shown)


src/colab/api.ts
```diff
@@ -206,7 +206,24 @@ export const CcuInfoSchema = z.object({
        * Number of tokens remaining in the "USAGE-mCCUs" quota group (remaining
        * free usage allowance in milli-CCUs).
        */
-      remainingTokens: z.number(),
+      // The API is defined in Protobuf and the field is an Int64. The ProtoJSON
+      // format (https://protobuf.dev/programming-guides/json) returns Int64 as
+      // a string so the value needs to be converted to a number. It's not
+      // expected we'll ever fail the `isSafeInteger` check since in practice
+      // the value is << 2^53 - 1 (the max value a number type can safely
```
amtoney524, line 213: nit: `<` instead of `<<` ?

kevineger: The `<<` here is intentional.  `<<` means _much-less-than_ ([Wikipedia](https://simple.wikipedia.org/wiki/List_of_mathematical_symbols)). It's somewhat ambiguous but given that our values should never be greater than 6,000, they are much less than `2^53-1`.

PR #80, opened 2025-04-23
build: esbuild all the things

kevineger:
Holy smokes... ESBuild is quick! But that's not why we're using it...

https://code.visualstudio.com/api/working-with-extensions/bundling-extension

Bundling helps ensure the extension will work across the various VS Code platforms (different OS-es and web). We need to bundle before we can publish.

The bundle structure is to output `out/extension.js` and `out/extension.js.map` when building the extension only and to preserve folder structure when building for tests.

There are various bundlers we could use, e.g. Webpack, Rollup, etc. I opted to use ESBuild for a few reasons.

- It's fast, like really fast.
- It's gaining a ton of popularity and has an involved community.
- It's recommended in the VS Code docs and what they use as their sample.
- It's popular across other extensions.
- Code config is nice and simple to setup, read and maintain.
- It has baked in problem solvers for VS Code (so you get red squigglies).

I've opted to not just bundle the extension for publishing, but also for development and testing. It's best practice to develop and test in the same way the extension will be run in production. This way we can catch any issues early.

The [docs](https://code.visualstudio.com/api/working-with-extensions/bundling-extension) touch on this in detail, but an important callout here is that ESBuild does not do any type checking. The standard and recommended approach is to run `tsc -noEmit` in parallel (which I'm doing with [concurrently](https://www.npmjs.com/package/concurrently)). This will ensure that we catch any type errors while still getting the speed and benefits of ESBuild.

In this change, we've got a few source-file diffs. This is largely due to the fact that VS Code still requires CJS modules. Attempting to stub exports from `node-fetch` (an ESM module), especially when imported as a namespace (`import * as nodeFetch`), leads to TypeErrors. This is because ES module exports, particularly in CJS interop scenarios, are non-configurable/non-writable ("read-only") by design.

This is fixed by removing direct attempts to stub properties on the imported `node-fetch` module namespace.

In making this change, I also standardized the npm _scripts_ to be more ergonomic.


esbuild.config.mjs
```diff
@@ -0,0 +1,137 @@
...
+    console.log(`📦 ${name} bundle (${fileName}) - ${size} KB`);
+  }
+}
+
+function buildReporter(name) {
+  return {
+    name: "build-reporter",
+    setup(build) {
+      build.onEnd((result) => {
+        if (result.errors.length > 0) {
```
zelliott, line 33: If @katlyn-edwards was here, she would balk at this and ask you to do a truthiness check in order to win at code golf.

But I'm a reasonable PR reviewer.

esbuild.config.mjs
```diff
@@ -0,0 +1,137 @@
...
+          console.log("\n🛑 Watch mode stopped");
+          process.exit(0);
+        });
+      } else {
+        const startTime = performance.now();
+        await esbuild.build(opts);
+        const endTime = performance.now();
+        const duration = endTime - startTime;
+
+        console.log(`🏃 Build completed in ${duration.toFixed(2)} ms`);
```
zelliott, line 129: obviously optional: If duration is slower than some threshold, can we use a 🐌 ?

kevineger: If we ever eclipse 1s, I'll circle back to this P0 😉.

esbuild.config.mjs
kevineger: 🤢... 🤫

PR #82, opened 2025-05-16
test: add end-to-end test for notebook execution

amtoney524:
This PR adds an initial E2E test that exercises the following CUJ:
1. The user triggers the Colab server provider
1. The user is authenticated and creates a Colab server
1. The user can successfully execute the notebook

Note that running the test locally requires `COLAB_EXTENSION_ENVIRONMENT="production"` and setting the `TEST_ACCOUNT_*` environment variables. 

Dev Dependencies:
* [vscode-extension-tester](https://www.npmjs.com/package/vscode-extension-tester) is used to install / run the extension in a test instance of VS Code. Additionally, it provides VS Code specific page objects, which are super convenient in that we don't have to go through the VS Code DOM as much.
* [clipboardy](https://www.npmjs.com/package/clipboardy) is used to access the clipboard as part of testing the OAuth flow.

Future work:
* set up the test for [debugging](https://github.com/redhat-developer/vscode-extension-tester/wiki/Debugging-Tests)
* with clipboardy, we're using a dynamic import to get around the following compiler error `The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. `. The solution for this seems like it could be more involved, so proposing to save that for a followup PR.

### DIFF
--- a/src/test/extension.e2e.test.ts
+++ b/src/test/extension.e2e.test.ts
@@ -20,7 +20,7 @@ describe("Colab Extension", function () {
   let workbench: Workbench;
 
   before(async () => {
-    // Wait for the extension to be activated.
+    // Wait for the extension to be installed.
     workbench = new Workbench();
     driver = workbench.getDriver();
     await driver.sleep(2000);


src/test/extension.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  Key,
+  ModalDialog,
+  WebDriver,
+  Workbench,
+  until,
+} from "vscode-extension-tester";
+
+describe("Colab Extension", function () {
+  this.timeout("2m"); // Override the default test suite timeout of 2s.
+  dotenv.config();
```
kevineger, line 17: Consider a top level constant (e.g. `ELEMENT_WAIT_MS`).

kevineger, line 17: Should we instead define a mocharc, or provide this as an option in the `extest` invocation so that each e2e test gets the default larger timeout?

src/test/extension.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  const waitMs = 5000; // A default timeout for explicit waits.
+  let driver: WebDriver;
+  let workbench: Workbench;
+
+  before(async () => {
+    // Wait for the extension to be activated.
+    workbench = new Workbench();
+    driver = workbench.getDriver();
+    await driver.sleep(2000);
```
kevineger, line 26: Is this the sleep waiting for the extension activation? Would it be possible to use the VS Code API to detect this instead? I think something like `extensions.getExtension("google.colab")?.isActive` inside a `driver.wait` (with some upper limit timeout) might work?

src/test/extension.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    // Create a notebook that we can execute. Note that it's created with a
+    // single code cell by default.
+    await workbench.executeCommand("Create: New Jupyter Notebook");
+    await workbench.executeCommand("Notebook: Edit Cell");
+    const cell = await driver.switchTo().activeElement();
+    await cell.sendKeys("1 + 1");
```
kevineger: It'd be great to push this notebook setup into a `describe`. Something like `describe("with a notebook", ...`.

src/test/extension.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  });
+
+  it("executes the notebook", async () => {
```
kevineger: Nit: test does a lot more than execute the notebook.

```suggestion
  it("authenticates and executes the notebook on a Colab server", async () => {
```

---

Great job on this test! These e2e tests are always so tricky to setup the first time.

To improve readability of body, consider lifting the independent actions into helper functions. E.g.:

- `selectColabKernelProvider(): void`
- `completeOAuthFlow(): void`
- `newColabCpuServer(kernel: string, variant: Variant, accelerator?: Accelerator)`
- `executeCell(idx: number): boolean`

src/test/extension.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    // Accept the dialog allowing the Colab extension to sign in using Google.
+    let dialog = new ModalDialog();
+    await dialog.pushButton("Allow");
+
+    // Begin the sign-in process by copying the OAuth URL to the clipboard and
+    // opening it in a browser window. Why do this instead of triggering the
+    // "Open" button in the dialog? We copy the URL so that we can use a new
+    // driver instance for the OAuth flow, since the original driver instance
+    // does not have a handle to the spawned window.
```
kevineger: Nit

```suggestion
    // does not have a handle to the window that would be spawned with "Open".
```

src/test/extension.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    // Input the test account email address.
+    const emailInput = await oauthDriver.findElement(
+      By.css("input[type='email']"),
+    );
+    await emailInput.sendKeys(process.env.TEST_ACCOUNT_EMAIL ?? "");
+    await emailInput.sendKeys(Key.ENTER);
+
+    // Input the test account password. Note that we wait for the page to settle
+    // to avoid getting stale element reference
```
kevineger: Nit

```suggestion
    // to avoid getting a stale element reference.
```

src/test/extension.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    await continueButton.click();
+
+    // Click Continue to allow the Colab extension to access the account.
+    await oauthDriver.wait(
+      until.urlContains("accounts.google.com/signin/oauth/v2/consent"),
+      waitMs,
+    );
+    continueButton = await oauthDriver.findElement(
+      By.xpath("//span[text()='Continue']"),
+    );
```
kevineger: Totally out of scope for this initial PR, but just spit balling for subsequent changes: should we have a wrapper for `findElement`?

Something like:

```
async findElement(driver: WebDriver, locator: Locator, descriptor: string): WebElementPromise
```

Then the implementation could provide human readable errors. E.g. call site for this case:

```
continueButton = await findElement(oauthDriver, By.xpath("//span[text()='Continue']"), "OAuth2 consent continue button");
```

src/test/extension.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    // The test account should be authenticated. Close the browser window.
+    await oauthDriver.wait(until.urlContains("127.0.0.1"), waitMs);
+    await oauthDriver.quit();
+
+    // Now that we're authenticated, we can resume creating a Colab server via
+    // the open kernel selector.
+    inputBox = await InputBox.create();
+    await inputBox.selectQuickPick("New Colab Server");
+    // Select the variant.
```
kevineger: Done! 🎉

```suggestion
```

src/test/extension.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    inputBox = await InputBox.create();
+    await inputBox.selectQuickPick("CPU");
+    // Alias the server with the default name.
```
kevineger: Nit: remove

package.json
```diff
@@ -83,9 +84,11 @@
     "format": "prettier . --write",
     "format:check": "prettier . --check",
     "pretest": "npm run build && npm run lint",
+    "pretest:e2e": "npm run build",
     "test:unit": "npm run build:tests && mocha --enable-source-maps --require ./out/test/test/unit-test-setup.js './out/test/**/*.unit.test.js'",
     "test:integration": "npm run build && node ./out/test/test/integration-test-runner.js",
     "test:integration:headless": "xvfb-run -s '-screen 0 1024x768x24' npm run test:integration",
+    "test:e2e": "extest setup-and-run ./out/test/**/*.e2e.test.js",
```
kevineger: Nit: group `pre*` by the rule it's a _pre_ of.
```suggestion
    "test:unit": "npm run build:tests && mocha --enable-source-maps --require ./out/test/test/unit-test-setup.js './out/test/**/*.unit.test.js'",
    "test:integration": "npm run build && node ./out/test/test/integration-test-runner.js",
    "test:integration:headless": "xvfb-run -s '-screen 0 1024x768x24' npm run test:integration",
    "pretest:e2e": "npm run build",
    "test:e2e": "extest setup-and-run ./out/test/**/*.e2e.test.js",
```

.env.template
```diff
@@ -1,4 +1,6 @@
 # One of "production", "sandbox", "local"
 COLAB_EXTENSION_ENVIRONMENT=
 COLAB_EXTENSION_CLIENT_ID=
-COLAB_EXTENSION_CLIENT_NOT_SO_SECRET=
\ No newline at end of file
+COLAB_EXTENSION_CLIENT_NOT_SO_SECRET=
+TEST_ACCOUNT_EMAILL=
```
kevineger: Typo

```suggestion
TEST_ACCOUNT_EMAIL=
```

src/test/extension.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    await dialog.pushButton("Allow");
+
+    // Begin the sign-in process by copying the OAuth URL to the clipboard and
+    // opening it in a browser window. Why do this instead of triggering the
+    // "Open" button in the dialog? We copy the URL so that we can use a new
+    // driver instance for the OAuth flow, since the original driver instance
+    // does not have a handle to the spawned window.
+    dialog = new ModalDialog();
+    await dialog.pushButton("Copy");
+    const clipboardy = await import("clipboardy");
```
kevineger: Could you add a TODO to remove this dynamic import? The reason it's required is a bit complicated. VS Code extensions are limited and required to target CommonJS. They have an ongoing effort to support ESM, but it's not ready. As a result, this causes issues with several newer modern libraries, like clipboardy, which use the newer ES module target. Old versions of clipboardy are likely to have CJS support, but the version we've added here does not. In an ideal world, we could have our extension target ESM... but unfortunately that's not a world we live in 🤣.

Let's fix this in a follow up change as your PR description states, but we'll likely have to modify the target for the e2e test build and/or modify the ts config.

package.json
kevineger, line 1: We don't need to solve this as part of this PR, but I'm not able to successfully run the e2e test. May be an issue with how I'm using VS Code (over SSH)?

```
SessionNotCreatedError: session not created: probably user data directory is already in use, please specify a unique value for --user-data-dir argument, or don't use --user-data-dir
     at Object.throwDecodedError (node_modules/selenium-webdriver/lib/error.js:523:15)
     at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:524:13)
     at Executor.execute (node_modules/selenium-webdriver/lib/http.js:456:28)
     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
     ```

File: src/test/extension.e2e.test.ts (hunk 8 above)
Comment (kevineger) [reply src/test/extension.e2e.test.ts]: Woops, this was in reference to a `// TODO`. Was reviewing in VS Code and may have been on an older commit. Disregard!

File: src/test/extension.e2e.test.ts (hunk 9 above)
Comment (kevineger) [reply src/test/extension.e2e.test.ts]: Same here. Sorry, disregard!

File: .gitignore [hunk 13]
```diff
@@ -2,4 +2,5 @@
 .vscode-test/
 node_modules/
 out/
+colab-0.0.1.vsix
```
kevineger: This should likely be all `.vsix` files.

package.json
amtoney524: yeah +1, it would be good to understand what's happening here! I filed a bug internally to track

.gitignore
amtoney524: good call, done!

src/test/extension.e2e.test.ts
amtoney524: done and filed a bug internally to track - thanks for the extra context here :)

.env.template
amtoney524: whoops, thanks for the catch!

package.json
amtoney524: done!

src/test/extension.e2e.test.ts
amtoney524: yeah, good point - it's interesting that `driver.wait` includes a param for a custom `message`, but `driver.findElement` does not 

With `findElement`, the default error messages are helpful (e.g., `NoSuchElementError: no such element: Unable to locate element: {"method":"css selector","selector":"input[type='emaill']"}`) , and I imagine there are cases where a custom error message would be useful as well!

src/test/extension.e2e.test.ts
amtoney524: done!

src/test/extension.e2e.test.ts
amtoney524: done!

src/test/extension.e2e.test.ts
amtoney524: done!

src/test/extension.e2e.test.ts
amtoney524: done!

amtoney524: having this defined as a mocha config sounds good to me. I can include this when setting up the test for debugging and created a bug to track internally

src/test/extension.e2e.test.ts
amtoney524: I was hoping this would work 😭 and my comment's probably incorrect, I think we're likely waiting on installation since technically the notebook has to exist before the extension's activated right? I'll update this to "to be installed" for the time being

Not sure if it's worth it, or if there's an easier way, but we could consider waiting for the extension to appear in the extensions view first ([reference](https://github.com/redhat-developer/vscode-extension-tester/wiki/ExtensionsViewSection)). Will explore this in a follow up PR

src/test/extension.e2e.test.ts
amtoney524: updated the title on the `it` block! for the helper functions, hoping it's alright to punt that to a follow up cl. I think it would be good to spend a bit more time on this since some of the lines can be blurry - for example, `selectColabKernelProvider` yields the sign in with Google dialog, which takes us into the `OAuthFlow`; if these are in separate functions, could some of that information be less obvious (e.g., should `completeOauthFlow` assume the dialog's present and start from there?)

PR #83, opened 2025-05-27
chore: generate config from env vars

kevineger:
This change requires developers to set up their environment variables and generate the config (`npm run generate:config`) before building the project. Failure to do so will result in an actionable build error reminding them to.

We've got a few configuration values which are environment specific, or should be omitted from source code. At the end of the day, we're shipping code to the VS Marketplace so we can never truly hide these values; but we can at least exclude them from the source code so they aren't picked up by things like scrapers, etc.

- Added an `.env.template` file to document the required environment variables.
- Added a `scripts/generate-config.mts` script to generate the `src/colab-config.ts` file from the environment variables.
- Updated the build process to validate the config exists and output an actionable error if it does not.
- Updated the GitHub workflow to generate + upload/download the config file.
- Modified the tsconfig so it includes the scripts.


scripts/generate-config.mts
```diff
@@ -0,0 +1,74 @@
+import fs from "fs/promises";
+import path from "path";
+import { fileURLToPath } from "url";
+import dotenv from "dotenv";
+
+const __filename = fileURLToPath(import.meta.url);
```
zelliott, line 6: TIL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta

PR #86, opened 2025-05-30
refactor: centralize the definition of commands

kevineger:
An additional minor change in here is removing the steps for the _remove server_ command as there's only the one step.

In a following change, the commands need to be referenced in multiple places. This is effectively a prefactor for that. But even if that change wasn't going to happen, this is a good cleanup.

### DIFF
--- a/src/colab/commands/constants.ts
+++ b/src/colab/commands/constants.ts
@@ -30,10 +30,10 @@ export const OPEN_COLAB_WEB: Command = {
   description: "Open Colab web.",
 };
 
-/** Command to open the Colab signup page, to upgrade to pro. */
-export const UPGRADE_TO_PRO: Command = {
-  label: "$(accounts-view-bar-icon) Upgrade to Pro",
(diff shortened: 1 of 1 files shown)


src/colab/commands/constants.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/** Identifying information for a Colab-registered command. */
+export interface RegisteredCommand extends Command {
+  /** The ID of the registered command. */
+  id: string;
+}
+
+/** Command to open the toolbar command selection. */
+export const COLAB_TOOLBAR: RegisteredCommand = {
+  id: "colab.toolbarCommand",
```
amtoney524, line 17: peeked at how this is used downstream, looks great!

src/colab/commands/constants.ts
```diff
@@ -0,0 +1,49 @@
+/** Identifying information for a Colab command. */
```
amtoney524, line 1: this is nice!! nit: worth alphabetizing the commands below?

PR #88, opened 2025-05-30
feat: add a notebook toolbar button for interacting with Colab

kevineger:
[VS Code Colab toolbar button.webm](https://github.com/user-attachments/assets/22de22e2-596a-4d89-bca5-3424a943a2a0)

### DIFF
--- a/project-words.txt
+++ b/project-words.txt
@@ -2,6 +2,7 @@ authuser
 colab
 gapi
 Googlers
+grayscale
 highmem
 milli
 pkce


src/colab/commands/notebook.unit.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      .onFirstCall()
+      // Arbitrarily select the first command.
+      .callsFake(findCommand(OPEN_COLAB_WEB.label));
+
+    await expect(notebookToolbar(vsCodeStub.asVsCode(), assignmentManager)).to
+      .eventually.be.fulfilled;
+
+    sinon.assert.calledOnceWithMatch(
+      vsCodeStub.window.showQuickPick,
+      commandsLabeled([OPEN_COLAB_WEB.label, UPGRADE_TO_PRO.label]),
```
amtoney524: should we also assert for the server commands here?

kevineger: Thanks for the catch! Blind copy-pasta without actually modifying the test 😅

PR #91, opened 2025-06-30
refactor: generalize auth login flows

kevineger:


src/auth/flows/flows.ts
```diff
@@ -0,0 +1,79 @@
...
+  redirectUri?: string;
+  /** An optional disposable to be disposed of once the flow is complete. */
+  disposable?: vscode.Disposable;
+}
+
+/**
+ * An OAuth2 flow that can be triggered to obtain an authorization code.
+ */
+export interface OAuth2Flow {
+  options: OAuth2FlowDescriptor;
```
zelliott, line 46: nit: Thoughts on a `supports(host: Host)` method? Might be easier to understand than the `options` / `Descriptor` pattern.

src/auth/flows/proxied.ts
```diff
@@ -0,0 +1,72 @@
...
+  OAuth2Flow,
+  OAuth2FlowDescriptor,
+  OAuth2TriggerOptions,
+  FlowResult,
+} from "./flows";
+
+const PROXIED_REDIRECT_URI = `${CONFIG.ColabApiDomain}/vscode/redirect`;
+
+export class ProxiedRedirectFlow implements OAuth2Flow, vscode.Disposable {
+  options: OAuth2FlowDescriptor = {
```
zelliott: If you stick with the `options` pattern, can this be made deeply readonly?

src/auth/login.ts
```diff
@@ -0,0 +1,135 @@
...
+            client,
+            flowResult,
+            pkce.codeVerifier,
+          );
+
+          flowResult.disposable?.dispose();
+          return res;
+        },
+      );
+    } catch (err) {
```
zelliott, line 73: Earlier you had `catch (err: unknown)` - is there any kind of eslint/tsconfig setting that can help this repo be consistent with type usage?

src/auth/login.ts
```diff
@@ -0,0 +1,135 @@
...
+      const innerMsg = err instanceof Error ? err.message : "unknown error.";
```
zelliott: veeeeeery nit: There were some instances above where you didn't have periods. This one does. Pick one?

src/auth/login.ts
```diff
@@ -0,0 +1,135 @@
...
+  oAuth2Client: OAuth2Client,
+  flowResult: FlowResult,
+  pkceVerifier: string,
+) {
+  const tokenResponse = await oAuth2Client.getToken({
+    code: flowResult.code,
+    codeVerifier: pkceVerifier,
+    redirect_uri: flowResult.redirectUri,
+  });
+  if (tokenResponse.res?.status !== 200) {
```
zelliott, line 109: I'm not quite sure what the type of `tokenResponse.res` is here, but can this be `!tokenResponse.res?.ok` instead?

src/auth/auth-provider.ts
```diff
@@ -51,10 +46,9 @@ export class GoogleAuthProvider
    */
   constructor(
     private readonly vs: typeof vscode,
-    private readonly packageInfo: PackageInfo,
     private readonly storage: AuthStorage,
     private readonly oAuth2Client: OAuth2Client,
-    private readonly codeProvider: CodeProvider,
+    private readonly login: (scopes: string[]) => Promise<Credentials>,
```
amtoney524, line 53: update JS doc to match?

src/auth/login.ts
kevineger: It's a `GaxiosResponse`. No `ok`.

```ts
export interface GaxiosResponse<T = any> {
    config: GaxiosOptions;
    data: T;
    status: number;
    statusText: string;
    headers: Headers;
    request: GaxiosXMLHttpRequest;
}
```

src/auth/login.ts
kevineger: Done.

src/auth/login.ts
kevineger: Looks like I can enable this, will explore in a separate PR.

https://www.typescriptlang.org/tsconfig/#useUnknownInCatchVariables

src/auth/flows/flows.ts
kevineger: Yeah can do. I've got actually adding the logic for flow selection behind a TODO below. I'll address it in that change shortly since I'm not 100% sure on the exact impl. details yet.

src/auth/auth-provider.ts
kevineger: Done.

src/auth/flows/proxied.ts
kevineger: Temporarily leaving it, so yes. Done.

PR #92, opened 2025-06-30
feat: implement standalone extension URI handler

kevineger:
This functionality is currently handled in the
`RedirectUriCodeProvider`. Sending this standalone change as a prefactor to that class slimming down to just managing the code-exchange portion of the flow (external promise lifecycle).

### DIFF
--- a/src/system/uri-handler.unit.test.ts
+++ b/src/system/uri-handler.unit.test.ts
@@ -6,18 +6,10 @@ import { ExtensionUriHandler } from "./uri-handler";
 
 describe("ExtensionUriHandler", () => {
   let vs: VsCodeStub;
-
-  beforeEach(() => {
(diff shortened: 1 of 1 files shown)


src/system/uri-handler.unit.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    vs = newVsCodeStub();
+  });
+
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  let handler: ExtensionUriHandler;
+
+  beforeEach(() => {
```
amtoney524, line 11: consolidate these with `beforeEach` and `afterEach` above?

kevineger: Done.

PR #93, opened 2025-06-30
feat: implement generic loopback server

kevineger:
This class with be leveraged in a downstream change for a local server OAuth2 flow.

The unit tests leverage a fake server to remain targeted and quick to run. We should consider adding an integration test that spins up a real server. This will be inherently slower, but will ensure that the server works as expected in a real environment.

### DIFF
--- a/src/common/loopback-server.ts
+++ b/src/common/loopback-server.ts
@@ -86,21 +86,22 @@ export class LoopbackServer implements vscode.Disposable {
     if (this.isDisposed) {
       throw new Error("Local server has already been disposed");
     }
-    if (!this.listen) {
-      this.listen = new Promise<number>((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


src/common/loopback-server.ts
```diff
@@ -0,0 +1,106 @@
...
+   *
+   * @returns A promise that resolves to the port number the server is listening
+   * on. If the server fails to start or the port cannot be determined, it
+   * rejects with an error.
+   */
+  async start(): Promise<number> {
+    if (this.isDisposed) {
+      throw new Error("Local server has already been disposed");
+    }
+    if (!this.listen) {
```
zelliott: nit: Invert this to reduce nesting?

kevineger: Done.

PR #94, opened 2025-06-30
feat: add loopback OAuth2 flow

kevineger:


src/auth/flows/loopback.ts
```diff
@@ -0,0 +1,129 @@
...
+      });
+
+      await this.vs.env.openExternal(this.vs.Uri.parse(authUrl));
+      return {
+        code: await code,
+        redirectUri: address,
+        disposable: server,
+      };
+    } catch (err: unknown) {
+      server.dispose();
```
zelliott, line 76: Optional: So on error, it disposes of the server itself, on success, the caller is expected to dispose. That's a little bit of a gotcha, to me. Could we either:

* Update `LocalServerFlow.trigger` to dispose of the server itself regardless of success/failure?
* Update `FlowResult` to include an `error?` field, and have `LocalServerFlow.trigger` never throw, and then have the caller dispose of the disposable again regardless of success/failure?

src/auth/flows/loopback.ts
```diff
@@ -0,0 +1,129 @@
...
+class Handler implements LoopbackHandler {
+  constructor(
+    private readonly serveRoot: string,
+    private readonly codeProvider: CodeManager,
+  ) {}
+
+  handleRequest(req: http.IncomingMessage, res: http.ServerResponse): void {
+    // URL and Host are only missing on malformed requests.
+    assert(req.url);
+    assert(req.headers.host);
```
zelliott, line 91: Do these dump useful error messages on failure?

src/auth/flows/loopback.ts
```diff
@@ -0,0 +1,129 @@
...
+        break;
+      }
+      default: {
+        console.warn("Received unhandled request: ", req);
+      }
+    }
+  }
+
+  handleError?: ((err: Error) => void) | undefined;
+  handleClose?: (() => void) | undefined;
```
zelliott: Do you plan to eventually implement these or have a `LoopbackHandler` implementation that does implement these? If not, remove?

src/auth/flows/loopback.ts
```diff
@@ -0,0 +1,129 @@
...
+    private readonly serveRoot: string,
+    private readonly codeProvider: CodeManager,
+  ) {}
+
+  handleRequest(req: http.IncomingMessage, res: http.ServerResponse): void {
+    // URL and Host are only missing on malformed requests.
+    assert(req.url);
+    assert(req.headers.host);
+    const url = new URL(req.url, `http://${req.headers.host}`);
+    switch (url.pathname) {
```
zelliott, line 98: Do you want to validate that the HTTP method is what you expect (e.g. I'm expecting you only want to handle `GET` requests here)?

src/auth/flows/loopback.ts
```diff
@@ -0,0 +1,129 @@
...
+        res.end("You may now return to the application.");
+        break;
+      }
+      case "/favicon.ico": {
+        const assetPath = url.pathname.substring(1);
+        sendFile(res, path.join(this.serveRoot, assetPath));
+        break;
+      }
+      default: {
+        console.warn("Received unhandled request: ", req);
```
zelliott, line 121: Does it make sense to respond with a 404 here?

src/auth/flows/loopback.ts
```diff
@@ -0,0 +1,129 @@
...
+  handleRequest(req: http.IncomingMessage, res: http.ServerResponse): void {
+    // URL and Host are only missing on malformed requests.
+    assert(req.url);
+    assert(req.headers.host);
+    const url = new URL(req.url, `http://${req.headers.host}`);
+    switch (url.pathname) {
+      case "/": {
+        const state = url.searchParams.get("state");
+        if (!state) {
+          throw new Error("Missing state in redirect URL");
```
zelliott, line 102: Dumb question: What response does the client see when you throw here (and below)? Does it make sense to respond with some kind of non-ok status code?

src/auth/flows/loopback.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  handleError?: ((err: Error) => void) | undefined;
+  handleClose?: (() => void) | undefined;
+}
+
+function sendFile(res: http.ServerResponse, filepath: string): void {
+  fs.readFile(filepath, (err, body) => {
+    if (err) {
+      console.error(err);
+      res.writeHead(404);
```
zelliott: Is this status code right? I _think_ I'd expect to see a 5xx status code here because this read file error represents an internal server error?

src/auth/flows/loopback.ts
kevineger: Done.

src/auth/flows/loopback.ts
kevineger: Yep! Done.

src/auth/flows/loopback.ts
kevineger: https://nodejs.org/api/assert.html#assertvalue-message

Useful enough for me! For example:

```ts
try {
  const a: string | undefined = undefined;
  assert(a);
} catch (e) {
  console.error(e);
}
```

Produces an `AssertionError`:

```
AssertionError [ERR_ASSERTION]: undefined == true
	// ... the stack ...
     {generatedMessage: true, code: 'ERR_ASSERTION', actual: undefined, expected: true, operator: '==', …}
```

src/auth/flows/loopback.ts
kevineger: That's a good point/question. I'll send a standalone PR fixing the `LoopbackServer` to `try`/`catch` the `handleRequest` call to ensure a generic 500 is sent and callers aren't deadlining/hanging.

src/auth/flows/loopback.ts
kevineger: I guess since the only call path to `sendFile` is for the `case "/favicon.ico"`, a 500 would be more appropriate. It's not like we're serving a folder of assets in which case a 404 would make more sense.

src/auth/flows/loopback.ts
kevineger: Done.

src/auth/flows/loopback.ts
kevineger: Had a think (and chatted briefly via DM), I think it makes the most sense to instead have the flow be disposable and push the server to a collection of disposables that will get cleaned up.

PR #97, opened 2025-07-08
fix: correct prod Colab API domain

kevineger:
While the webapp will redirect fine from colab.google.com, we're
returning a 200 with some JS to do that 🫨. To avoid the redirect we
can't easily follow (as if it were a 300), this change configures the
URL to the redirected endpoint.

### DIFF
--- a/scripts/generate-config.mts
+++ b/scripts/generate-config.mts
@@ -30,8 +30,8 @@ try {
   }
   switch (envConfig.env) {
     case "production":
-      colabApiDomain = "https://colab.google.com";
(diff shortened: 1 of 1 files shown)


src/jupyter/provider.ts
```diff
@@ -144,6 +144,7 @@ export class ColabJupyterServerProvider
     } catch (_) {
       // Including the command to upgrade to pro is non-critical. If it fails,
       // just return the commands without it.
+      console.log(_);
```
amtoney524: remove? :)

kevineger: Beat me to it ;)

PR #99, opened 2025-07-11
test: configure e2e test for debugging

amtoney524:
These changes configure the E2E test for debugging.

Here, we specify a [launch configuration](https://github.com/redhat-developer/vscode-extension-tester/wiki/Debugging-Tests) that allows us to set breakpoints in the E2E test.

Additionally, we provide a mocha config to specify the global test timeout depending on whether we're running or debugging the tests. 

Screenshot 2025-07-10 at 8 50 40 PM

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -30,7 +30,8 @@
         "--debug"
       ],
       "console": "integratedTerminal",
-      "internalConsoleOptions": "neverOpen"
+      "internalConsoleOptions": "neverOpen",
+      "preLaunchTask": "npm: build"
     },
     {
       "name": "Extension Integration Tests",


.vscode/launch.json
```diff
@@ -16,6 +16,22 @@
       "runtimeExecutable": "npm",
       "runtimeArgs": ["run-script", "test:unit", "--", "--timeout", "0"]
     },
+    {
```
kevineger, line 19: Should we `preLaunchTask` to build?

src/test/e2e.mocharc.js
```diff
@@ -0,0 +1,4 @@
+const isDebugMode = process.argv.includes("--debug");
```
kevineger, line 1: Love it! 👍

PR #102, opened 2025-07-25
test: support headless mode for e2e tests

amtoney524:
This change adds a new script to run our E2E tests in headless mode and bumps a few waits for decreased flakiness.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "test:unit": "npm run build:tests && mocha --enable-source-maps --require ./out/test/test/unit-test-setup.js './out/test/**/*.unit.test.js'",
     "test:integration": "npm run build && node ./out/test/test/integration-test-runner.js",
     "test:integration:headless": "xvfb-run -s '-screen 0 1024x768x24' npm run test:integration",
-    "pretest:e2e": "npm run build",
+    "pretest:e2e": "npm run build:tests",
     "test:e2e": "extest setup-and-run ./out/test/*.e2e.test.js -m ./out/test/e2e.mocharc.js",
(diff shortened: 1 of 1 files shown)


src/test/extension.e2e.test.ts
```diff
@@ -130,3 +131,26 @@ describe("Colab Extension", function () {
...
+  const isHeadlessMode = process.argv.includes("--headless");
+  if (isHeadlessMode) {
+    // Without the provided user agent, the OAuth flow fails with a 500.
+    return new Builder()
+      .forBrowser("chrome")
+      .setChromeOptions(
+        new chrome.Options()
+          .addArguments("--headless")
+          .addArguments(
+            '--user-agent="Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36"',
```
amtoney524, line 148: This is the example from https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent#chrome_ua_string

src/test/extension.e2e.test.ts
```diff
@@ -10,7 +11,7 @@ import {
   until,
 } from "vscode-extension-tester";
 
-const ELEMENT_WAIT_MS = 5000;
+const ELEMENT_WAIT_MS = 10000;
```
amtoney524, line 14: bumped this since connecting to the kernel can take a bit

package.json
```diff
@@ -89,6 +89,7 @@
     "test:integration:headless": "xvfb-run -s '-screen 0 1024x768x24' npm run test:integration",
     "pretest:e2e": "npm run build",
```
kevineger: Not necessarily in scope for this PR, but shouldn't this be `npm run build:tests`?

PR #103, opened 2025-07-29
test: take screenshot of oauth window on failure

amtoney524:
When encountering a failure during the OAuth sign in flow, capture a screenshot.

The test naming is intended to align with [vscode extension tester's naming](https://github.com/redhat-developer/vscode-extension-tester/wiki/Taking-Screenshots), plus a suffix to identify the oauth window 

Screenshot 2025-07-29 at 4 14 39 PM

### DIFF
--- a/src/test/extension.e2e.test.ts
+++ b/src/test/extension.e2e.test.ts
@@ -29,9 +29,12 @@ describe("Colab Extension", function () {
     await driver.sleep(4000);
   });
 
+  beforeEach(function () {
+    testTitle = this.currentTest?.fullTitle() ?? "";
+  });
+
   describe("with a notebook", () => {
-    beforeEach(async function () {
-      testTitle = this.currentTest?.fullTitle() ?? "";
+    beforeEach(async () => {
       // Create an executable notebook. Note that it's created with a single
       // code cell by default.
       await workbench.executeCommand("Create: New Jupyter Notebook");
@@ -61,10 +64,7 @@ describe("Colab Extension", function () {
       await dialog.pushButton("Copy");
       // TODO: Remove this dynamic import
       const clipboardy = await import("clipboardy");
-      await doOauthSignIn({
-        oauthUrl: clipboardy.default.readSync(),
-        testTitle,
-      });
+      await doOauthSignIn(/* oauthUrl= */ clipboardy.default.readSync());
 
       // Now that we're authenticated, we can resume creating a Colab server via
       // the open kernel selector.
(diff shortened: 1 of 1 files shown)


src/test/extension.e2e.test.ts
```diff
@@ -27,7 +30,8 @@ describe("Colab Extension", function () {
   });
 
   describe("with a notebook", () => {
-    beforeEach(async () => {
+    beforeEach(async function () {
+      testTitle = this.currentTest?.fullTitle() ?? "";
```
kevineger: Should we move this to the top-level `describe` and change it to `beforeEach`?

src/test/extension.e2e.test.ts
```diff
@@ -57,53 +61,10 @@ describe("Colab Extension", function () {
...
-        By.xpath("//span[text()='Continue']"),
-      );
-      await continueButton.click();
-
-      // The test account should be authenticated. Close the browser window.
-      await oauthDriver.wait(until.urlContains("127.0.0.1"), ELEMENT_WAIT_MS);
-      await oauthDriver.quit();
+      await doOauthSignIn({
+        oauthUrl: clipboardy.default.readSync(),
+        testTitle,
```
kevineger: Consider nesting the `function doOauthSignIn` in the `describe` which can access `testTitle` instead of providing it.

src/test/extension.e2e.test.ts
```diff
@@ -132,6 +93,78 @@ describe("Colab Extension", function () {
...
+    );
+    continueButton = await oauthDriver.findElement(
+      By.xpath("//span[text()='Continue']"),
+    );
+    await continueButton.click();
+
+    // The test account should be authenticated. Close the browser window.
+    await oauthDriver.wait(until.urlContains("127.0.0.1"), ELEMENT_WAIT_MS);
+    await oauthDriver.quit();
+  } catch (_) {
```
kevineger: Should this still be thrown? Right now it's swallowing the error.

If the OAuth flow failed, I think the test will just continue?

kevineger: To confirm behaviour, do you mind forcing this to fail and confirm the screenshot + test failure is as expected? (E.g. perhaps use a wrong password.)

src/test/extension.e2e.test.ts
amtoney524: good idea, done!

src/test/extension.e2e.test.ts
amtoney524: nice! this also simplified the `doOauthSignIn` signature

src/test/extension.e2e.test.ts
amtoney524: ah good catch! (confirmed failure offline)

PR #104, opened 2025-07-30
test: move e2e test running to standalone script

kevineger:
With our release pipeline, we've got several options we need to be able to configure when running the e2e tests. This change pushes that complexity into a standalone script that can be run with different parameters.

### DIFF
--- a/scripts/test_e2e.sh
+++ b/scripts/test_e2e.sh
@@ -1,11 +1,12 @@
 #!/bin/bash
 #
-# This script is used to run end-to-end tests for the Colab VSCode extension.
+# This script is used to run end-to-end tests for the Colab VS Code extension.
 #
 # Usage:
 #   ./scripts/test_e2e.sh [--headless] [--vsix=<path_to_vsix_file>] [--auth-driver:<args>...]
 #
 # Arguments:
+#   - storage=<path>: Path to use for extest storage.
 #   - headless: Run tests in headless mode using xvfb.
 #   - vsix=<path_to_vsix_file>: Install the specified VSIX file before running tests.
 #   - auth-driver:<args>: Pass additional arguments to the authentication driver.


scripts/test_e2e.sh
```diff
@@ -0,0 +1,129 @@
+#!/bin/bash
+#
+# This script is used to run end-to-end tests for the Colab VSCode extension.
```
amtoney524: nit: `VS Code`?

scripts/test_e2e.sh
```diff
@@ -0,0 +1,137 @@
...
+#
+# Usage:
+#   ./scripts/test_e2e.sh [--headless] [--vsix=<path_to_vsix_file>] [--auth-driver:<args>...]
+#
+# Arguments:
+#   - headless: Run tests in headless mode using xvfb.
+#   - vsix=<path_to_vsix_file>: Install the specified VSIX file before running tests.
+#   - auth-driver:<args>: Pass additional arguments to the authentication driver.
```
amtoney524, line 12: nit: wondering if should we mention the storage arg here as well (since its in the help text below)?

scripts/test_e2e.sh
```diff
@@ -0,0 +1,137 @@
...
+# 
+# A couple ugly workarounds are made to have this script be portable to Mac which by default
```
amtoney524, line 14: nice comment! (and thanks for fixing this!!!)

scripts/test_e2e.sh
kevineger: Done.

scripts/test_e2e.sh
kevineger: Woops! Thanks.

scripts/test_e2e.sh
kevineger: Thanks for testing on your Mac and letting me know!

PR #106, opened 2025-07-31
test: deflake dialog button push in e2e test

amtoney524:
Adds a util to wait for a successful dialog button press before proceeding with the test.

why? we noticed this was flaky yesterday and want to mitigate that before this runs in a pipeline. The util also generates an error message to notify us which dialog failed.

error before (generic): 
Screenshot 2025-07-31 at 11 26 05 AM

error after (identifies dialog):
Screenshot 2025-07-31 at 11 24 15 AM

### DIFF
--- a/src/test/extension.e2e.test.ts
+++ b/src/test/extension.e2e.test.ts
@@ -106,8 +106,8 @@ describe("Colab Extension", function () {
     button: string;
     dialog: string;
   }) {
-    // ModalDialog.pushButton will throw if the dialog is not found; so, to
-    // reduce flakes, we attempt this until it succeeds or the times out.
+    // ModalDialog.pushButton will throw if the dialog is not found; to reduce
+    // flakes we attempt this until it succeeds or times out.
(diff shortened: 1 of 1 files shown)


src/test/extension.e2e.test.ts
```diff
@@ -92,6 +96,34 @@ describe("Colab Extension", function () {
...
+   */
+  async function pushDialogButton({
+    button,
+    dialog,
+  }: {
+    button: string;
+    dialog: string;
+  }) {
+    // ModalDialog.pushButton will throw if the dialog is not found; so, to
+    // reduce flakes, we attempt this until it succeeds or the times out.
```
kevineger: Nit: wording

```suggestion
    // ModalDialog.pushButton will throw if the dialog is not found; to
    // reduce flakes we attempt this until it succeeds or times out.
```

amtoney524: thanks for the catch!

PR #108, opened 2025-07-31
fix: GPU and TPU assignment parameter bug

kevineger:
`api.ts` really is starting to show some inconsitencies with the Colab API! It's a bit funky to say the least. This change uses the appropriate enum type (num vs. string) over the wire and documents why and where the conversion is necessary.

Tested manually and can assign across all the machine types.

Took the opportunity to rework the `client.unit.test.ts` matcher.

Also added the response body of failed requestse to the error message.

### DIFF
--- a/src/colab/client.unit.test.ts
+++ b/src/colab/client.unit.test.ts
@@ -3,6 +3,7 @@ import { expect } from "chai";
 import fetch, { Response } from "node-fetch";
 import { SinonStub, SinonMatcher } from "sinon";
 import * as sinon from "sinon";
+import { uuidToWebSafeBase64 } from "../utils/uuid";
 import {
   Accelerator,
   CcuInfo,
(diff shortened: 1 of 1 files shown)


src/colab/api.ts
```diff
@@ -310,7 +297,21 @@ export const AssignmentSchema = z
     /** The outcome of the assignment. */
     outcome: z.nativeEnum(Outcome).optional(),
     /** The variant of the assignment. */
-    variant: z.nativeEnum(Variant),
+    // On GET, this is a string (enum) but on POST this is a number.
```
amtoney524, line 300: TIL, thanks for the comment!

src/colab/client.ts
```diff
@@ -296,8 +296,15 @@ export class ColabClient {
...
     if (!response.ok) {
+      let errorBody = "";
+      try {
+        errorBody = await response.text();
+      } catch {
+        // Ignore errors reading the body
+      }
       throw new Error(
-        `Failed to issue request to ${endpoint.toString()}: ${response.statusText}`,
+        `Failed to issue request ${request.method} ${endpoint.toString()}: ${response.statusText}` +
```
amtoney524, line 306: nice!

PR #110, opened 2025-07-31
test: fail e2e test when environment is not set to production

amtoney524:
The test account only works in the production environment. When the test fails due to an unexpected environment, the failure is not obvious (e.g., kernel selection will hang). Because of this, we should fail the test when we detect an unexpected environment.

### DIFF
--- a/src/test/extension.e2e.test.ts
+++ b/src/test/extension.e2e.test.ts
@@ -13,6 +13,7 @@ import {
   VSBrowser,
   until,
 } from "vscode-extension-tester";
+import { CONFIG } from "../colab-config";
 
 const ELEMENT_WAIT_MS = 10000;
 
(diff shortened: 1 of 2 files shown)


src/test/extension.e2e.test.ts
```diff
@@ -23,6 +24,11 @@ describe("Colab Extension", function () {
   let workbench: Workbench;
 
   before(async () => {
+    assert.equal(
+      process.env.COLAB_EXTENSION_ENVIRONMENT,
```
kevineger: This could pass if the config was generated with `sandbox`, then the test was run with the environment variable set to `production`.

What we need to check is the `CONFIG.ColabApiDomain` is `"https://colab.research.google.com"`. Alternatively, we could add `environment` to the `CONFIG` object (I like this better) [here](https://github.com/googlecolab/colab-vscode/blob/83f4a7346a71d9b88b92ff86825b15e71cd9c655/scripts/generate-config.mts#L55-L60).

amtoney524: thanks for the catch! added the `environment` to the `CONFIG` object and updated the assertion to read from there

PR #111, opened 2025-07-31
test: increase e2e test waits on load and cell execution

amtoney524:
Increases e2e test waits for initial load of VS Code and cell execution.

Why? Observed these as points of flakiness when running the e2e test in a Github workflow.

### DIFF
--- a/src/test/extension.e2e.test.ts
+++ b/src/test/extension.e2e.test.ts
@@ -16,6 +16,7 @@ import {
 import { CONFIG } from "../colab-config";
 
 const ELEMENT_WAIT_MS = 10000;
+const CELL_EXECUTION_WAIT_MS = 30000;
 
 describe("Colab Extension", function () {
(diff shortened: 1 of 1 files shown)


src/test/extension.e2e.test.ts
```diff
@@ -14,6 +14,7 @@ import {
 } from "vscode-extension-tester";
 
 const ELEMENT_WAIT_MS = 10000;
+const CELL_EXECUTION_WAIT_MS = 30000;
```
kevineger, line 19: While these times are certainly longer than we'd expect and want, it's a-OK to have them bumped so long as the overall test is sub 2-minutes. We can revisit improving things once the bigger fish have been fried 🍤

amtoney524: ack 🐟, also happy to revisit this once we have a better idea of how the test behaves on its pipelines

PR #112, opened 2025-08-01
test: remove select another kernel QuickPick from e2e test

amtoney524:
Here, we remove the "Select Another Kernel..." QuickPick from the e2e test.

Now that we're failing when QuickPick items aren't found (https://github.com/googlecolab/colab-vscode/pull/109), running the test in a Github workflow fails at this step.

When developing locally, I would get an extra QuickPick due to the Jupyter extension auto-connecting to my python environment. This extra step is not reflected in the workflow run; so, we remove it here.

### DIFF
--- a/src/test/extension.e2e.test.ts
+++ b/src/test/extension.e2e.test.ts
@@ -54,10 +54,6 @@ describe("Colab Extension", function () {
       // Select the Colab server provider from the kernel selector.
       await workbench.executeCommand("Notebook: Select Notebook Kernel");
       await selectQuickPickItem({
-        item: "Select Another Kernel...",
-        quickPick: "Change kernel",
-      });
-      await selectQuickPickItem({
         item: "Colab",
         quickPick: "Select Another Kernel",
       });


amtoney524:
Verified the test failure repro'd on another device

amtoney524:
> Does this still pass as expected locally for you, with the _Change kernel_ quick pick removed?

Yes, but not as is on mac. I have to run the test with the setting `"jupyter.kernels.excludePythonEnvironments": ["usr/bin/python3"]`.

https://github.com/redhat-developer/vscode-extension-tester/wiki/Test-Setup#set-up-and-run-tests

Screenshot 2025-08-04 at 12 03 34 PM

kevineger:
> > Does this still pass as expected locally for you, with the _Change kernel_ quick pick removed?
> 
> Yes, but not as is on mac. I have to run the test with the setting `"jupyter.kernels.excludePythonEnvironments": ["usr/bin/python3"]`.
> 
> https://github.com/redhat-developer/vscode-extension-tester/wiki/Test-Setup#set-up-and-run-tests
> 
> Screenshot 2025-08-04 at 12 03 34 PM

We should sandbox our tests in a workspace and have them run with that setting by default. Can provide the settings for the test run ([docs](https://github.com/redhat-developer/vscode-extension-tester/wiki/Test-Setup#run-tests)).

PR #113, opened 2025-08-01
ci: add e2e tests to build-test workflow

amtoney524:
Adds a step for running E2E tests to the Build and Test workflow

When the test fails, screenshots and the extension host logs are uploaded as workflow artifacts

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -106,3 +106,47 @@ jobs:
       - name: npm run test:integration
         if: matrix.os != 'ubuntu-latest'
         run: npm run test:integration
+
+  e2eTests:
+    name: VS Code E2E Tests
+    strategy:
+      matrix:
+        # macos-latest and windows-latest can be added as we work towards a release.
+        # For now, there's no sense acruing the cost of the runners.
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    needs: generateConfig
+    env:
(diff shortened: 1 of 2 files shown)


src/test/extension.e2e.test.ts
```diff
@@ -34,6 +34,24 @@ describe("Colab Extension", function () {
     testTitle = this.currentTest?.fullTitle() ?? "";
   });
 
+  afterEach(async function () {
+    if (this.currentTest?.state === "failed") {
```
amtoney524: fwiw, I don't love this solution because it messes with the UI state and makes screenshots less valuable (e.g., if we were stuck on a QuickPick, I think we would lose that in the screenshot because we've executed this in the command palette.) Hopefully we're able to find where in the file system the logs reside and upload them

amtoney524: resolving since we identified the exthost logs directory and can upload them as artifacts - nice find, @kevineger!
Screenshot 2025-08-05 at 5 51 05 PM

.github/workflows/build-test.yml
```diff
@@ -106,3 +106,47 @@ jobs:
...
+      - name: npm run test:e2e
+        if: matrix.os != 'ubuntu-latest'
+        run: npm run test:e2e
+      - name: upload screenshots
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: screenshots
+          path: /tmp/test-resources/screenshots/
+      - name: upload exthost.log
```
kevineger, line 147: ❤︎ it!

PR #115, opened 2025-08-04
test: swallow errors in selectQuickPickItem

amtoney524:
This PR updates the `selectQuickPickItem` helper to swallow any errors thrown by `InputBox.create`.

### DIFF
--- a/src/test/extension.e2e.test.ts
+++ b/src/test/extension.e2e.test.ts
@@ -135,8 +135,7 @@ describe("Colab Extension", function () {
           await quickPickItem.select();
           return true;
         } catch (_) {
-          // Swallow the InputBox.create error since we want to fail when our
-          // timeout's reached.
+          // Swallow errors since we want to fail when our timeout's reached.
           return false;
         }
       },


src/test/extension.e2e.test.ts
```diff
@@ -124,15 +124,21 @@ describe("Colab Extension", function () {
...
+          // InputBox.selectQuickPick will not throw if the item is not found.
+          const quickPickItem = await inputBox.findQuickPick(item);
+          if (!quickPickItem) {
+            return false;
+          }
+          await quickPickItem.select();
+          return true;
+        } catch (_) {
+          // Swallow the InputBox.create error since we want to fail when our
+          // timeout's reached.
```
kevineger: Since we only expect the `InputBox.create` to occasionally error and that's the one we want to swallow, would it make sense to move the wrapping `try`/`catch` to just be around that?

kevineger: Alternatively (or in addition?) is there a specific `Error` instance we can key against and check?

amtoney524: I tested out keying against `TimeoutError` here and in the dialog util. With the dialog util, I saw we could also run into `ElementNotInteractable` as well. Because of this, I would air on keeping the try/catch and error swallowing as is, since we can flake after the element's located. I updated the comment to be more generic, and I'm happy to adjust this if we find this is obscuring errors we would otherwise want to see (e.g., maybe we console log errors we're swallowing so they're visible to developers?)

PR #122, opened 2025-08-07
refactor: define ColabRequestError

amtoney524:
Defines an error type that exposes the request method, url, and response.

Why? This will allow us to better handle the thrown error from `client.issueRequest`.

### DIFF
--- a/src/colab/client.ts
+++ b/src/colab/client.ts
@@ -331,8 +331,7 @@ export class ColabClient {
         // Ignore errors reading the body
       }
       throw new ColabRequestError({
-        method: request.method,
-        url: endpoint.toString(),
+        request,
         response,
         responseBody: errorBody,
       });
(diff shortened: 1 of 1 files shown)


src/colab/client.ts
```diff
@@ -354,3 +356,31 @@ function stripXssiPrefix(v: string): string {
   }
   return v.slice(XSSI_PREFIX.length);
 }
+
+class ColabRequestError extends Error {
+  readonly body: string;
```
kevineger: Could we make this optionally `undefined` since not all error responses will have a body and/or there could be a swallowed error preventing us from streaming the value; `undefined` would be more semantically correct than `""`.

```suggestion
  readonly body?: string;
```

src/colab/client.ts
```diff
@@ -354,3 +356,31 @@ function stripXssiPrefix(v: string): string {
...
+  readonly method: string;
+  readonly response: fetch.Response;
+  readonly url: string;
```
kevineger: Nit: could we order these by their scope/usage?

```suggestion
  readonly method: string;
  readonly url: string;
  readonly response: fetch.Response;
  readonly body?: string;
```

src/colab/client.ts
```diff
@@ -354,3 +356,31 @@ function stripXssiPrefix(v: string): string {
   }
   return v.slice(XSSI_PREFIX.length);
 }
+
+class ColabRequestError extends Error {
+  readonly body: string;
+  readonly method: string;
+  readonly response: fetch.Response;
```
kevineger, line 361: Optional: Since we're putting the full `fetch.Response` on the error, wondering if it'd make sense to include the `fetch.Request` instead of teasing apart the `url` + `method`. Your call! Whatever feels more ergonomic.

src/colab/client.ts
amtoney524: done!

src/colab/client.ts
amtoney524: thanks for the suggestion, done! I like that it simplifies the call site and class, and we can always revisit which properties are exposed as this evolves. While here, I also updated `body` to `responseBody` to be more obvious to readers

PR #123, opened 2025-08-07
refactor: centralize header definitions

kevineger:
A prefactor to cleanup header definition/usage across the codebase before I use a header across multiple files.

### DIFF
--- a/src/colab/headers.ts
+++ b/src/colab/headers.ts
@@ -27,7 +27,7 @@ export const CONTENT_TYPE_JSON_HEADER: StaticHeader = {
 };
 
 /**
- * The HTTP header for JSON content type.
+ * The HTTP header for text content type.
  */
 export const CONTENT_TYPE_TEXT_HEADER: StaticHeader = {
   key: "Content-Type",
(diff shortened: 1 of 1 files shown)


src/colab/headers.ts
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * The HTTP header for JSON content type.
+ */
+export const CONTENT_TYPE_JSON_HEADER: StaticHeader = {
+  key: "Content-Type",
+  value: "application/json",
+};
+
+/**
+ * The HTTP header for JSON content type.
```
amtoney524: text?

src/colab/headers.ts
```diff
@@ -0,0 +1,81 @@
...
+ */
+export const CONTENT_TYPE_TEXT_HEADER: StaticHeader = {
+  key: "Content-Type",
+  value: "text/plain",
+};
+
+/**
+ * The HTTP header for the accepting JSON.
```
amtoney524: nit: remove "the"?

kevineger: Woops

src/colab/headers.ts
kevineger: Woops x2

PR #125, opened 2025-08-08
feat: alert users when assignment quotas reached

amtoney524:
Notifies the user when assignment creation fails due to quota.

Previously, the kernel selection flow failed silently when this error was encountered. Now, we display an error message with an action directing the user to remove an existing server.

Screenshot 2025-08-08 at 12 05 39 PM

### DIFF
--- a/src/jupyter/assignments.unit.test.ts
+++ b/src/jupyter/assignments.unit.test.ts
@@ -2,6 +2,7 @@ import { randomUUID } from "crypto";
 import { assert, expect } from "chai";
 import fetch, { Headers } from "node-fetch";
 import sinon, { SinonStubbedInstance } from "sinon";
+import { Uri } from "vscode";
 import {
   Accelerator,
   Assignment,
@@ -599,13 +600,12 @@ describe("AssignmentManager", () => {
           /.*Remove a server.*/,
           "Remove Server at Colab Web",
         );
-        sinon.assert.calledOnceWithMatch(vsCodeStub.env.openExternal, {
-          scheme: "https",
-          authority: "colab.research.google.com",
-          path: "/",
-          query: "",
-          fragment: "",
-        });
+        sinon.assert.calledOnceWithMatch(
+          vsCodeStub.env.openExternal,
+          sinon.match(function (url: Uri) {
+            return url.toString() === "https://colab.research.google.com/";
+          }),
+        );
       });
     });
   });


src/colab/client.ts
```diff
@@ -277,15 +277,22 @@ export class ColabClient {
...
+        url,
+        {
+          method: "POST",
+          headers: { [XSRF_HEADER_KEY]: xsrfToken },
+          signal,
+        },
+        AssignmentSchema,
+      );
+    } catch (error) {
+      if (error instanceof ColabRequestError && error.response.status === 412) {
```
kevineger, line 292: Consider adding a comment that this is for _Precondition Failed_.

src/colab/client.ts
```diff
@@ -379,3 +386,6 @@ class ColabRequestError extends Error {
     this.responseBody = responseBody;
   }
 }
+
+/** Error thrown when the user has too many assignments. */
+export class TooManyAssignmentsError extends Error {}
```
kevineger: Fine to ignore since existing ordering in this file doesn't totally adhere to this, but would be nice to have all exported members towards the top of the file (visiblity DESC).

src/colab/client.unit.test.ts
```diff
@@ -246,6 +246,57 @@ describe("ColabClient", () => {
...
+            host: COLAB_HOST,
+            path,
+            queryParams,
+            otherHeaders: {
+              "X-Goog-Colab-Token": "mock-xsrf-token",
+            },
+          }),
+        )
+        .resolves(
+          new Response(withXSSI(JSON.stringify("")), {
```
kevineger: Can be simplified to `undefined`.

```suggestion
          new Response(undefined), {
```

src/colab/client.unit.test.ts
```diff
@@ -246,6 +246,57 @@ describe("ColabClient", () => {
...
+            status: 412,
+          }),
+        );
+
+      await expect(
+        client.assign(NOTEBOOK_HASH, Variant.DEFAULT),
+      ).to.eventually.be.rejectedWith(TooManyAssignmentsError);
+      sinon.assert.calledTwice(fetchStub);
```
kevineger: Nit: either omit this `calledTwice` check (since it's implicitly covered by the expectation of a `TooManyAssignmentsError`) or separate it with a newline to split arrange-act-assert sections.

src/jupyter/assignments.ts
```diff
@@ -253,11 +253,20 @@ export class AssignmentManager implements vscode.Disposable {
   private async assignOrRefresh(
     toAssign: ColabJupyterServer,
   ): Promise<ColabAssignedServer> {
-    const { assignment, isNew } = await this.client.assign(
-      toAssign.id,
-      toAssign.variant,
-      toAssign.accelerator,
-    );
+    let assignment;
+    let isNew;
```
kevineger: Nit: give these an annotated type. Surprised linter defaults don't complain about this, therefore optional!

src/jupyter/assignments.ts
```diff
@@ -253,11 +253,20 @@ export class AssignmentManager implements vscode.Disposable {
...
+    try {
+      ({ assignment, isNew } = await this.client.assign(
+        toAssign.id,
+        toAssign.variant,
+        toAssign.accelerator,
+      ));
+    } catch (error) {
+      if (error instanceof TooManyAssignmentsError) {
```
kevineger, line 265: Nit: I try (and sometimes succeed 😄) to avoid naming methods `handle...`. Instead, focus on describing what the behaviour is. In this case, something like `notifyMaxAssignmentsExceeded` would be more descriptive.

Internal guidance: go/tott/702.

src/jupyter/assignments.ts
```diff
@@ -253,11 +253,20 @@ export class AssignmentManager implements vscode.Disposable {
...
+        void this.handleTooManyAssignmentsError();
+      }
+      throw error; // Rethrow so error gets logged.
```
kevineger: Uber optional nit, avoid assuming what callers do with the error as those comments can go stale and assume call patterns are fixed.

```suggestion
      throw error;
```

src/jupyter/assignments.ts
```diff
@@ -308,6 +317,34 @@ export class AssignmentManager implements vscode.Disposable {
       },
     };
   }
+
+  private async handleTooManyAssignmentsError() {
+    // TODO: Account for subscription tiers in actions.
+    // TODO: Account for the number of assignments from the VS Code and Colab
+    // UIs in the error message and actions.
+    enum Actions {
```
kevineger: Is defining an `enum` in a method ideal? This effectively re-creates a small object each time the method is called. Should the definition be outside the class method? If the intention was to reduce scope and the benefits there outweigh the (potential) drawback, all good.

src/jupyter/assignments.ts
```diff
@@ -308,6 +317,34 @@ export class AssignmentManager implements vscode.Disposable {
...
+      case Actions.REMOVE_SERVER:
+        this.vs.commands.executeCommand("colab.removeServer");
+        return;
+      case Actions.REMOVE_SERVER_COLAB_WEB:
+        this.vs.env.openExternal(
+          this.vs.Uri.parse("https://colab.research.google.com/"),
+        );
+        return;
+      default:
+      // Do nothing
```
kevineger: Nit: Prefer to have the code be self-documenting.

```suggestion
        return;
```

src/jupyter/assignments.unit.test.ts
```diff
@@ -531,6 +531,89 @@ describe("AssignmentManager", () => {
         });
       });
     });
+
+    describe("when a server is not assigned due to quota", () => {
```
kevineger, line 536: **Totally your call**, but since we know how we intend to enrich this notification (as listed in your TODOs), consider a test structure which lends itself well to those incremental improvements.

- describe: with too many VS Code servers
  - it: notifies the user to remove
- describe: with too many Colab web servers
  - it: notifies the user to remove
- describe: with a mix of too many servers across VS Code and Colab web
  - it: notifies the user to remove

I suspect the gradual improvement would be clean diffs

1. on the _it_, e.g.: _it: notifies the user to remove_ -> _it: notifies the user to remove in VS Code and Colab Web_.
2. on the describe, e.g.: `describe: for a free user` ... `describe: with too many VS Code servers` ... `it: notifies to subscribe or remove`.

src/jupyter/assignments.unit.test.ts
```diff
@@ -531,6 +531,89 @@ describe("AssignmentManager", () => {
...
+
+      it("notifies the user", async () => {
+        try {
+          await assignmentManager.assignServer(
+            defaultServer.id,
+            defaultAssignmentDescriptor,
+          );
+        } catch (_) {
+          // Error shouldn't fail test
+        }
```
kevineger: Check out `to.eventually.be.rejectedWith`. Apply globally.

src/jupyter/assignments.unit.test.ts
```diff
@@ -531,6 +531,89 @@ describe("AssignmentManager", () => {
...
+
+        sinon.assert.calledOnceWithMatch(
+          vsCodeStub.window.showErrorMessage,
+          "You have too many servers. Remove a server to continue.",
```
kevineger: Nit: when possible avoid matching on full error messages, `".*Remove a server.*"` should suffice. Apply globally.

src/jupyter/assignments.unit.test.ts
```diff
@@ -531,6 +531,89 @@ describe("AssignmentManager", () => {
...
+          "Remove Server at Colab Web",
+        );
+        sinon.assert.calledOnceWithMatch(vsCodeStub.env.openExternal, {
+          scheme: "https",
+          authority: "colab.research.google.com",
+          path: "/",
+          query: "",
+          fragment: "",
+        });
```
kevineger: Nit: readability could be improved by asserting over the human-readable representation (`.toString()`).

src/colab/client.ts
amtoney524: done!

src/colab/client.ts
amtoney524: moved this adjacent to the `ColabClient` export - will keep that in mind for the future!

src/colab/client.unit.test.ts
amtoney524: omit sgtm, thanks!

src/colab/client.unit.test.ts
amtoney524: ty!

src/jupyter/assignments.ts

src/jupyter/assignments.ts
amtoney524: good point and +1 to your naming suggestion :)

src/jupyter/assignments.ts

src/jupyter/assignments.ts
amtoney524: good point, done!

src/jupyter/assignments.ts

src/jupyter/assignments.unit.test.ts
amtoney524: done!

src/jupyter/assignments.unit.test.ts
amtoney524: done!

src/jupyter/assignments.unit.test.ts
amtoney524: love this, thank you!

src/jupyter/assignments.unit.test.ts
amtoney524: good point, happy to restructure these test cases as a prefactor to the TODOs

PR #134, opened 2025-08-15
chore: define build and lint config in mts

kevineger:
In doing this, applied the license lint fix to the relevant files. Manually applied the license to the bash and YAML files using https://github.com/google/addlicense. As time permits, I'll look into folding that into a GitHub action for verification and a pre-commit hook for a tighter dev loop. Non-critical for now.

Now that we're on Node 22, we can use https://eslint.org/docs/latest/use/configure/configuration-files#native-typescript-support.

We were missing the `eslint-config-prettier` types, so added those in. Unfortunately there are no types for `eslint-plugin-license-header` and it's not worth defining them ourselves, so suppressed the error but verified things work as expected.

Sorry to bundle this in, but ran `$ shfmt --write --simplify --indent 2 scripts/test_e2e.sh` to clean it up. Will look into instrumenting that into a lint check as well.


esbuild.config.mts
```diff
@@ -0,0 +1,219 @@
...
+    ...baseOptions,
+    entryPoints: [entrypoint],
+    outfile: path.join("out/test", outfile),
+    bundle: false, // Do not bundle, just transpile for test setup files.
+    external: undefined, // 'external' cannot be used when 'bundle' is false.
+    plugins: [buildReporter(name), nodeExternalsPlugin()],
+  };
+}
+
+/**
```
amtoney524, line 150: thanks for adding all these comments!

PR #138, opened 2025-08-19
chore: prepare the extension manifest for release

kevineger:
https://code.visualstudio.com/api/references/extension-manifest

Version corresponds to Semver [best practices](https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase).

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,14 +26,14 @@
     "Notebooks"
   ],
   "tags": [
-    "notebook",
-    "google",
     "colab",
-    "jupyter",
-    "python",
+    "google",
+    "interactive",
     "julia",
-    "r",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -18,6 +25,16 @@
   "categories": [
     "Notebooks"
   ],
+  "tags": [
```
amtoney524, line 28: nit: worth alphabetizing or no?

kevineger: I went in (my perceived) order of scope/importance since I thought that's how'd they render in the marketplace, but looking at some other extensions that doesn't appear to be the case. Fixed.

PR #146, opened 2025-08-26
refactor: toggle background tasks

kevineger:
There's a handful of coupled changes here, sorry there wasn't an easier way to do this. I was finding the lifecycle management of background tasks a bit murky. This was exacerbated when I tried to implement a more granular control over their execution depending on whether the user was signed in to use Colab or not. This prefactor introduces a new `Toggleable` class to help manage the state of these background tasks more easily, instead of starting them on construction.

Couple with this change is some renaming and re-orginization to the consumption files.

### DIFF
--- a/src/common/task-runner.ts
+++ b/src/common/task-runner.ts
@@ -34,6 +34,16 @@ export enum OverrunPolicy {
 }
 
 /**
+ * Specifies when the task runner should start its first run.
+ */
+export enum StartMode {
+  /** Start the task immediately. */
+  Immediately,
+  /** Start the task after the configured {@link Config.intervalTimeoutMs}. */
+  Scheduled,
+}
+
+/**
  * Runs a task at a regular interval, ensuring that only one task is running at
  * a time.
  *
@@ -43,7 +53,7 @@ export enum OverrunPolicy {
 export class SequentialTaskRunner implements Disposable {
   private inFlight?: Promise<void>;
   private inFlightAbort?: AbortController;
-  private timeout?: NodeJS.Timeout;
+  private taskInterval?: NodeJS.Timeout;
 
   constructor(
     private readonly config: Config,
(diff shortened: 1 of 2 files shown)


src/common/task-runner.ts
```diff
@@ -36,22 +36,52 @@ export enum OverrunPolicy {
...
  * a time.
+ *
+ * When disposed, this class stops any scheduled intervals and aborts any
+ * in-flight task.
  */
 export class SequentialTaskRunner implements Disposable {
   private inFlight?: Promise<void>;
   private inFlightAbort?: AbortController;
-  private readonly timeout: NodeJS.Timeout;
+  private timeout?: NodeJS.Timeout;
```
amtoney524: nit: even though the object returned by `setInterval` is a `NodeJS.Timeout`, wdyt about renaming to interval? (when reviewing, I expected a call `setTimeout` based on the naming)

src/common/task-runner.ts
```diff
@@ -36,22 +36,52 @@ export enum OverrunPolicy {
...
+  /**
+   * Stops running.
+   *
+   * If an execution is in-flight, it is aborted. If already stopped, does
+   * nothing.
+   */
+  stop(): void {
     clearInterval(this.timeout);
+    this.timeout = undefined;
     this.inFlightAbort?.abort();
```
amtoney524, line 98: (existing) should we be clearing `inFlightAbort` in the `run` function's `finally` clause?

src/common/task-runner.unit.test.ts
```diff
@@ -87,67 +89,122 @@ describe("SequentialTaskRunner", () => {
...
+    sinon.assert.calledTwice(task);
+    runner.dispose();
+  });
+
+  it("runs immediately when specified", async () => {
+    const runner = buildRunner(task, OverrunPolicy.AllowToComplete, {
+      intervalTimeoutMs: INTERVAL_TIMEOUT_MS,
+      taskTimeoutMs: TASK_TIMEOUT_MS,
     });
+    runner.start(true);
```
amtoney524: very nit: add parameter comment for `immediately`?

kevineger: Worthy nit! Let's make this API more obvious and have `start` accept a `StartMode` enum so call-sites are obvious.

src/common/task-runner.ts
kevineger: Functionally it shouldn't make a difference, but yeah I think it'd be best to clear it when it's no longer needed. Will help avoid accidental usage after the expected lifecycle. Thanks!

src/common/task-runner.ts
kevineger: Renamed!

PR #150, opened 2025-09-23
feat: complete loopback OAuth back to VS Code

kevineger:
Out with the plain text _You may now return to the application_ response, in with the nice full redirect back into VS Code! Upon obtaining the OAuth code, we redirect back to Colab's static "auth success" page, which completes the flow by setting window focus back to the originating VS Code window.

See the demo below for a split-screen view of the "first run experience" of OAuth'ing (including the Xdg-open / URL handler Chrome dialog warning).

[cast.webm](https://github.com/user-attachments/assets/068b99ea-32cc-48ae-a016-b727e20a21bb)

### DIFF
--- a/src/test/helpers/uri.ts
+++ b/src/test/helpers/uri.ts
@@ -122,7 +122,7 @@ export class TestUri implements vscode.Uri {
  * @param regExp - A regular expression to match against the stringified URI.
  * @returns A Sinon matcher for the URI.
  */
-export function matchUri(regExp: RegExp | string) {
+export function matchUri(regExp: RegExp | string): sinon.SinonMatcher {
   if (typeof regExp === "string") {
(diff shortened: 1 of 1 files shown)


src/test/helpers/uri.ts
```diff
@@ -113,3 +114,17 @@ export class TestUri implements vscode.Uri {
...
 }
+
+/**
+ * Creates a Sinon matcher for a {@link vscode.Uri} that matches the provided
+ * regular expression.
+ *
+ * @param regExp - A regular expression to match against the stringified URI.
+ * @returns A Sinon matcher for the URI.
+ */
+export function matchUri(regExp: RegExp | string) {
```
EvanWiederspan: Add return type

PR #156, opened 2025-09-26
feat: track the date servers are assigned

kevineger:
This is needed in a downstream change which adds _Auto-connect_ functionality (connect to most recently created server, or the _default_).

### DIFF
--- a/src/jupyter/storage.unit.test.ts
+++ b/src/jupyter/storage.unit.test.ts
@@ -212,6 +212,29 @@ describe("ServerStorage", () => {
         ]);
       });
 
+      it("does not override the date assigned", async () => {
+        const updatedServer: ColabAssignedServer = {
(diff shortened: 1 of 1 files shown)


src/jupyter/storage.ts
```diff
@@ -97,6 +99,10 @@ export class ServerStorage {
     const existingServersJson = await this.secrets.get(ASSIGNED_SERVERS_KEY);
     const serversById = mapServersById(existingServersJson);
     for (const server of servers) {
+      // This ensures that updating an existing server does not modify the
+      // original assignment date.
+      const dateAssigned =
```
amtoney524, line 104: (double checking) does this have unit test coverage?

kevineger: The happy path does, but good point! I'll add one for the other.

PR #159, opened 2025-10-03
feat: add auth code input for proxied redirect auth flow

amtoney524:
Adds a QuickInput to the ProxiedRedirectFlow

This allows the user to paste the OAuth2 code provided as part of the proxied redirect flow

[successful login](https://github.com/user-attachments/assets/b8c81604-f107-4376-8892-a90eee07d8a8)

[proxied redirect cancellation.webm](https://github.com/user-attachments/assets/f8a0eaae-ee20-422e-ac01-7bce80472e9b)

[proxied redirect timeout.webm](https://github.com/user-attachments/assets/554d200b-4a86-499d-9a81-5f7007e28c6d)

### DIFF
--- a/src/auth/flows/proxied.ts
+++ b/src/auth/flows/proxied.ts
@@ -46,26 +46,30 @@ export class ProxiedRedirectFlow implements OAuth2Flow, vscode.Disposable {
     options.cancel.onCancellationRequested(() => {
       cancelTokenSource.cancel();
     });
-    const code = this.codeManager.waitForCode(
-      options.nonce,
-      cancelTokenSource.token,
-    );
-    const vsCodeRedirectUri = this.vs.Uri.parse(
-      `${this.baseUri}?nonce=${options.nonce}`,
-    );
-    const externalProxiedRedirectUri =
-      await this.vs.env.asExternalUri(vsCodeRedirectUri);
-    const authUrl = this.oAuth2Client.generateAuthUrl({
-      ...DEFAULT_AUTH_URL_OPTS,
-      redirect_uri: PROXIED_REDIRECT_URI,
-      state: externalProxiedRedirectUri.toString(),
-      scope: options.scopes,
-      code_challenge: options.pkceChallenge,
-    });
+    try {
+      const code = this.codeManager.waitForCode(
+        options.nonce,
+        cancelTokenSource.token,
+      );
+      const vsCodeRedirectUri = this.vs.Uri.parse(
+        `${this.baseUri}?nonce=${options.nonce}`,
+      );
+      const externalProxiedRedirectUri =
+        await this.vs.env.asExternalUri(vsCodeRedirectUri);
+      const authUrl = this.oAuth2Client.generateAuthUrl({
+        ...DEFAULT_AUTH_URL_OPTS,
+        redirect_uri: PROXIED_REDIRECT_URI,
+        state: externalProxiedRedirectUri.toString(),
+        scope: options.scopes,
+        code_challenge: options.pkceChallenge,
+      });
 
-    await this.vs.env.openExternal(this.vs.Uri.parse(authUrl));
-    this.promptForAuthorizationCode(options.nonce, cancelTokenSource);
-    return { code: await code, redirectUri: PROXIED_REDIRECT_URI };
+      await this.vs.env.openExternal(this.vs.Uri.parse(authUrl));
+      this.promptForAuthorizationCode(options.nonce, cancelTokenSource);
+      return { code: await code, redirectUri: PROXIED_REDIRECT_URI };
+    } finally {
+      cancelTokenSource.dispose();
+    }
   }
 
   private promptForAuthorizationCode(


amtoney524:
> Overall change LGTM, but if I press `Escape`, the _Signing in to Google..._ progress notification still shows.
> 
> We should either (a) cancel the sign-in flow or (b) modify the notification to have `Re-open Code Input` in addition to `Cancel`.

good idea! added a commit to reflect option a, ptal

src/auth/flows/proxied.ts
```diff
@@ -40,37 +42,50 @@ export class ProxiedRedirectFlow implements OAuth2Flow, vscode.Disposable {
...
-      return { code: await code, redirectUri: PROXIED_REDIRECT_URI };
-    } finally {
-      listener.dispose();
-    }
+    await this.vs.env.openExternal(this.vs.Uri.parse(authUrl));
+    this.displayAuthCodeInput(options.nonce);
+    return { code: await code, redirectUri: PROXIED_REDIRECT_URI };
+  }
+
+  private displayAuthCodeInput(nonce: string) {
```
kevineger: This method does more than display the input. Would `promptForAuthorizationCode` or something of the like be better?

src/auth/flows/proxied.unit.test.ts
```diff
@@ -86,16 +92,28 @@ describe("ProxiedRedirectFlow", () => {
     clock.restore();
   });
```
kevineger, line 94: Could we add a unit test for the validation? Some prior art to reference in other tests.

src/auth/code-manager.ts
```diff
@@ -8,6 +8,9 @@ import vscode from "vscode";
 
 const EXCHANGE_TIMEOUT_MS = 60_000;
 
+export const USER_CANCELLATION_ERROR_MESSAGE =
```
kevineger: If we end up centralizing the cancellation into the token, the following feedback can be ignored as it's no longer relevant:

- This is only used with `new Error`, should we just define it as such?
- Separately, why export? That makes it part of the modules public API but we should be able to change this freely. If just exported for test, prefer to keep the test DAMP over DRY for readability.

src/auth/flows/proxied.ts
```diff
@@ -40,37 +42,50 @@ export class ProxiedRedirectFlow implements OAuth2Flow, vscode.Disposable {
...
-      const nonce = params.get("nonce");
-      const code = params.get("code");
-      if (!nonce || !code) {
-        // A request for `vscode://google.colab` without a code or nonce (while
-        // the user's in the authentication flow) is possible and should be
-        // ignored.
-        return;
-      }
-      this.codeManager.resolveCode(nonce, code);
+    const code = this.codeManager.waitForCode(options.nonce, options.cancel);
```
kevineger: Talked in person, but just to close the loop:

We can create a new `CancellationTokenSource` and pass it here which can be cancelled:

- When `options.cancel` is cancelled (e.g. the user clicked _Cancel)
- Pass it to `displayAuthCodeInput` and `cancel` when an `InputFlowAction.cancel` error is thrown

src/common/multi-step-quickpick.ts
```diff
@@ -45,6 +45,8 @@ export interface QuickInputOptions {
   totalSteps?: number;
   /** Buttons for actions in the UI. */
   buttons?: QuickInputButton[];
+  /** If the UI should stay open even when losing UI focus. */
+  ignoreFocusOut?: boolean;
```
kevineger, line 49: Mind updating the _configures the provided options_ test case so all options are validated as being passed appropriately?

src/auth/flows/proxied.ts
amtoney524: good idea, done!

src/auth/code-manager.ts
amtoney524: went the cancellation token route, thanks a lot for the suggestion 

(also nice catch with export, this was leftover from a previous revision exploring adding a `reject` method to the `CodeManager`)

src/auth/flows/proxied.ts
amtoney524: done!

src/auth/flows/proxied.unit.test.ts
amtoney524: done!

src/common/multi-step-quickpick.ts
amtoney524: good idea, thanks!

src/auth/flows/proxied.ts
```diff
@@ -40,37 +42,60 @@ export class ProxiedRedirectFlow implements OAuth2Flow, vscode.Disposable {
...
-      const nonce = params.get("nonce");
-      const code = params.get("code");
-      if (!nonce || !code) {
-        // A request for `vscode://google.colab` without a code or nonce (while
-        // the user's in the authentication flow) is possible and should be
-        // ignored.
-        return;
-      }
-      this.codeManager.resolveCode(nonce, code);
+    const cancelTokenSource = new this.vs.CancellationTokenSource();
```
kevineger, line 45: I believe this should be `dispose()`-d (in a `try`/`finally`?).

amtoney524: thanks for the catch!

PR #162, opened 2025-10-07
feat: show progress notification when removing server

amtoney524:
Renders a progress notification when removing a server

why? to alert the user that the action's in progress (e.g., may not be obvious over slow network)

### DIFF
--- a/src/colab/commands/server.ts
+++ b/src/colab/commands/server.ts
@@ -63,7 +63,6 @@ export async function renameServerAlias(
 // TODO: Consider making this multi-select.
 // TODO: Handle bug where, if the server of the connected kernel is
 // removed, a fallback kernel is selected but does not connect.
-// TODO: Consider adding a notification that the server was removed.
 // TODO: Update MultiStepInput to handle a single-step case.
 export async function removeServer(
   vs: typeof vscode,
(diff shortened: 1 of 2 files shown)


src/colab/commands/server.ts
```diff
@@ -83,7 +82,14 @@ export async function removeServer(
...
       })
     ).value;
-    await assignmentManager.unassignServer(selectedServer);
+    await vs.window.withProgress(
+      {
+        cancellable: false,
+        location: vs.ProgressLocation.Notification,
+        title: `Removing server "${selectedServer.label}"...`,
+      },
+      () => assignmentManager.unassignServer(selectedServer),
```
kevineger, line 91: Sorry, don't have a chance to pull this down and verify easily - mind checking?

I believe the supplied `task` to `withProgress` can be an `async` function. Should that be the case here with us awaiting it? Or is `await`-ing the `withProgress` effectively the same?

amtoney524: good question! I think this should be the same since the task implicitly returns the promise on L91. The docs mention "progress is shown while running the given callback and while the promise it returned isn't resolved nor rejected" ([source](https://code.visualstudio.com/api/references/vscode-api)).

PR #163, opened 2025-10-08
feat: support retrieving the last known servers

kevineger:
Needed in a downstream change which improves extension activation when the user is not signed in (e.g. first run experience or an activation following a sign-out).

### DIFF
--- a/src/jupyter/assignments.unit.test.ts
+++ b/src/jupyter/assignments.unit.test.ts
@@ -485,11 +485,24 @@ describe("AssignmentManager", () => {
 
       const servers = await assignmentManager.getLastKnownAssignedServers();
 
-      const expectedServers = storedServers.map((s) => {
-        const { connectionInformation, ...rest } = s;
-        return rest;
-      });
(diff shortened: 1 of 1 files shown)


src/jupyter/assignments.unit.test.ts
```diff
@@ -469,6 +469,30 @@ describe("AssignmentManager", () => {
...
+      const storedServers = [
+        { ...defaultServer, id: randomUUID() },
+        { ...defaultServer, id: randomUUID() },
+      ];
+      await serverStorage.store(storedServers);
+
+      const servers = await assignmentManager.getLastKnownAssignedServers();
+
+      const expectedServers = storedServers.map((s) => {
+        const { connectionInformation, ...rest } = s;
```
amtoney524: nit: thoughts on hardcoding the expected list since this logic's similar to what's being tested?

kevineger: Good point! I felt like I had written that code before...

Made the test more DAMP to avoid excessive DRY-ness and re-implementing the SUT.

PR #164, opened 2025-10-08
fix: auth lifecycle in Jupyter provider

kevineger:
This fix actually fixes quite a few issues:

- It ensures users aren't prompted to sign-in when the extension activates.
- It "disconnects" the kernel when a user signs out.
- It displays a command prompting the user to sign in if they previously had servers but since signed out.
- It doesn't force the user to sign in when listing the available Colab operations (until they click one requiring auth).

Additionally, here we add a missing test case for handling the auto-connect command.

### DIFF
--- a/src/jupyter/provider.ts
+++ b/src/jupyter/provider.ts
@@ -120,7 +120,7 @@ export class ColabJupyterServerProvider
   ): Promise<JupyterServerCommand[]> {
     const commands: JupyterServerCommand[] = [];
     // Only show the command to view existing servers if the user is not signed
-    // in but previously had assigned servers. Otherwise, it's the command is
+    // in, but previously had assigned servers. Otherwise, the command is
     // redundant.
     if (
       !this.isAuthorized &&


src/jupyter/provider.ts
```diff
@@ -102,11 +114,34 @@ export class ColabJupyterServerProvider
...
-  provideCommands(
+  async provideCommands(
     _value: string | undefined,
     _token: CancellationToken,
-  ): ProviderResult<JupyterServerCommand[]> {
-    return this.provideRelevantCommands();
+  ): Promise<JupyterServerCommand[]> {
+    const commands: JupyterServerCommand[] = [];
+    // Only show the command to view existing servers if the user is not signed
+    // in but previously had assigned servers. Otherwise, it's the command is
```
amtoney524: nit: remove it's?

src/jupyter/provider.ts
```diff
@@ -102,11 +114,34 @@ export class ColabJupyterServerProvider
...
+      !this.isAuthorized &&
+      (await this.assignmentManager.getLastKnownAssignedServers()).length > 0
+    ) {
+      commands.push(SIGN_IN_VIEW_EXISTING);
+    }
+    commands.push(AUTO_CONNECT, NEW_SERVER, OPEN_COLAB_WEB);
+    if (!this.isAuthorized) {
+      return commands;
+    }
+    try {
```
amtoney524, line 135: nit: should we also take upgrade_to_pro out of the colab toolbar button when not authorized?

src/jupyter/provider.ts
kevineger: Done.

src/jupyter/provider.ts
kevineger: Likely should, yes. Though looking at that implementation, we also need to make that command subscription aware. E.g. show "Upgrade to Pro+" if the user is already pro or "Buy additional compute units" if the user is already Pro+. Scoping that outside of this PR, let's cut a bug internally to circle back.

PR #165, opened 2025-10-09
feat: validate the Jupyter version works

kevineger:
image

All the context is in https://github.com/microsoft/vscode/issues/270285 and the related/linked issues and PRs.

TL;DR: VS Code / Jupyter shipped a bug which caused the Jupyter extension when updated to 2025.9 to fallback to a 2023 version. This old version is no longer supported and working. For now, Colab requires users to pin Jupyter to a working, modern version (so we can use the extensibility API). This validation ensures that users have correctly configured the dependent Jupyter extension so failures to activate aren't confusing.


kevineger:
Disabled the e2e test and integration test since we don't pre-install Jupyter at a working version. _Flying blind_ for a bit here, to unblock some internal testing.

PR #170, opened 2025-10-21
fix: remove the rename server alias command

amtoney524:
This change removes `Rename Server Alias` from the command palette and notebook toolbar

why? when renaming a server, the updated server label is not reflected in the recent kernels list. Because this can be confusing for users, we're opting to remove the command for now. 

We filed an issue at https://github.com/microsoft/vscode-jupyter/issues/17107 for the Jupyter extension team to help take a look.

Note that the command is still defined in `servers.ts` and `constants.ts` as we'll want to add it back once the recent kernels list reflects the renaming.


amtoney524:
> Mind filing an issue on the Jupyter extension and linking it here with the TODO and in the PR description?

good idea, done!

PR #171, opened 2025-10-22
fix: notify users to reload notebooks on server removal

amtoney524:
This change adds a notification to advise the user to reload any notebooks connected to the removed server.

why? The notebook must be reloaded before connecting to a new server; otherwise, cells will fail to execute. See https://github.com/microsoft/vscode-jupyter/issues/17094 for additional details.

### DIFF
--- a/src/colab/commands/server.unit.test.ts
+++ b/src/colab/commands/server.unit.test.ts
@@ -192,85 +192,57 @@ describe("Server Commands", () => {
         ]);
       });
 
-      it("unassigns the selected server", async () => {
-        assignmentManagerStub.getAssignedServers.resolves([defaultServer]);
+      describe("when a server is removed", () => {
+        let remove: Promise<void>;
 
-        const remove = removeServer(
-          vsCodeStub.asVsCode(),
-          assignmentManagerStub,
-        );
-        await quickPickStub.nextShow();
-        quickPickStub.onDidChangeSelection.yield([
-          { label: defaultServer.label, value: defaultServer },
-        ]);
-
-        expect(remove).to.eventually.be.fulfilled;
-        assignmentManagerStub.unassignServer.calledOnceWithExactly(
-          defaultServer,
-        );
-      });
-
-      it("notifies the user while removing the server", async () => {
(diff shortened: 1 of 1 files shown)


src/colab/commands/server.ts
```diff
@@ -90,6 +88,20 @@ export async function removeServer(
...
     return undefined;
   });
 }
+
+async function notifyReloadNotebooks(vs: typeof vscode, serverLabel: string) {
+  const result = await vs.window.showInformationMessage(
+    `Sorry for the inconvenience, to work around [microsoft/vscode-jupyter #17094](https://github.com/microsoft/vscode-jupyter/issues/17094) - please re-open notebooks ${serverLabel} was previously connected to.`,
+    `View Issue`,
+  );
+  if (result) {
```
kevineger: Nit: consider making the button click variable more obvious.

```suggestion
  const viewIssue = await vs.window.showInformationMessage(
    `Sorry for the inconvenience, to work around [microsoft/vscode-jupyter #17094](https://github.com/microsoft/vscode-jupyter/issues/17094) - please re-open notebooks ${serverLabel} was previously connected to.`,
    `View Issue`,
  );
  if (viewIssue) {
```

src/colab/commands/server.unit.test.ts
```diff
@@ -233,6 +233,69 @@ describe("Server Commands", () => {
           sinon.match.func,
         );
       });
+
```
kevineger, line 217: This structure is likely a bit more concise and could improve readability:

- `describe` _when a server is removed_ (a `beforeEach` which does that)
  - `it` _notifies the user to reload_
  - `it` _opens the Jupyter GitHub issue when the notification is clicked_

src/colab/commands/server.ts
amtoney524: nice, done!

src/colab/commands/server.unit.test.ts
amtoney524: good point, added a beforeEach and restructured tests accordingly

PR #174, opened 2025-10-27
fix: robustness improvements to task runner

kevineger:
There are a few changes and improvements going on here, as well as some changes to extension-wide logging.

- Add a `name` for better logging context for a `Task`.
- Grant tasks a configurable "abandon grace period" when overrun. This isn't used by the existing tasks, but will be used in a follow-up PR.
- Simplifies the management of active tasks.
- Dedicated errors to make control flow more obvious.
- Prevent the leaking of unhandled exceptions. Today, if a background task fails, this shows in the extensions view as an "unhandled exception" to users.

Ended up re-writing the tests with a more readable `TestTask` class to simplify control flow and make them more obvious.

### DIFF
--- a/src/common/task-runner.ts
+++ b/src/common/task-runner.ts
@@ -83,6 +83,8 @@ export class SequentialTaskRunner implements Disposable {
     abortCtrl: AbortController;
   };
   private taskInterval?: NodeJS.Timeout;
+  // A lock used to ensure execution is sequential.
+  private isRunning = false;
 
   constructor(
     private readonly config: Config,
@@ -107,12 +109,11 @@ export class SequentialTaskRunner implements Disposable {
       return;
     }
     if (mode === StartMode.Immediately) {
-      void this.run();
+      this.run();
     }
-    this.taskInterval = setInterval(
-      () => void this.run(),
-      this.config.intervalTimeoutMs,
-    );
+    this.taskInterval = setInterval(() => {
+      this.run();
+    }, this.config.intervalTimeoutMs);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/common/task-runner.ts
```diff
@@ -95,34 +124,157 @@ export class SequentialTaskRunner implements Disposable {
...
+      }
     } finally {
       clearTimeout(timeout);
       this.inFlight = undefined;
-      this.inFlightAbort = undefined;
     }
   }
+
+  /**
+   * Wraps a task promise to provide a graceful shutdown mechanism.
```
amtoney524, line 229: thanks for the detailed comment!

src/common/task-runner.ts
```diff
@@ -95,34 +124,157 @@ export class SequentialTaskRunner implements Disposable {
...
+
+        case OverrunPolicy.AbandonAndRun: {
+          const { abortCtrl, promise } = this.inFlight;
+          const abandonError = new OverrunAbandonError(this.task.name);
+          log.warn(abandonError.message);
+          abortCtrl.abort(abandonError);
+          try {
+            await promise;
+          } catch (err: unknown) {
+            if (err instanceof NonGracefulAbandonError) {
```
amtoney524: I could be missing something, but would this end up logging twice? once here and once from the prior call to run, around L190?

kevineger: Totally fair question, but no it won't double log.

Suppose a new `run()` (B) is issued while a previous one (A) is in-flight:

- B enters the `if (this.inFlight)` block and hits the `OverrunPolicy.AbandonAndRun` case.
- It gets the `promise` of the in-flight run A.
- It aborts A.
- It awaits A's promise inside the case's try/catch block (here).
- If A failed to shut down gracefully, its promise rejects with a `NonGracefulAbandonError`.
- This rejection gets caught by this (inside the `AbandonAndRun` case) try/catch block and logs the error.
- This switch `break`-s and proceeds to create a brand new task. `this.inFlight` gets set to the new task B and then task B is awaited in the second try/catch block below.

---

But your question did cause me to take a closer look at this, and find a race condition!

Consider `setInterval` firing again between the state check (`if (this.inFlight)`) and the state set (`this.inFlight = ...`). It isn't atomic because of the `await promise` in the `AbandonAndRun` which could cause two tasks to be running at the same time. This would make the second taska "zombie" task that's not controlleable.

Fixed this by introducing an `isRunning` lock which makes the synchronous `run` re-entrant and prevent the race. Added tests.

PR #176, opened 2025-10-31
feat: support switching OAuth clients for dev

kevineger:
If you switch the configured OAuth client, the `unauthorized_client` 401 error will be thrown and this new logic will run. Generically matched on 401 since I think that's sufficient. Anything beyond that would be on message matching which is brittle.

### DIFF
--- a/src/auth/auth-provider.ts
+++ b/src/auth/auth-provider.ts
@@ -126,7 +126,7 @@ export class GoogleAuthProvider
       await this.oAuth2Client.refreshAccessToken();
     } catch (err: unknown) {
       // This should only ever be the case when developer building from source
-      // switch the OAuth client ID / secret.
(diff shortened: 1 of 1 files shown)


src/auth/auth-provider.ts
```diff
@@ -120,7 +122,23 @@ export class GoogleAuthProvider
       token_type: "Bearer",
       scope: session.scopes.join(" "),
     });
-    await this.oAuth2Client.refreshAccessToken();
+    try {
+      await this.oAuth2Client.refreshAccessToken();
+    } catch (err: unknown) {
+      // This should only ever be the case when developer building from source
+      // switch the OAuth client ID / secret.
```
amtoney524: nit: switches?

kevineger: Done!

PR #178, opened 2025-10-31
feat: implement a generalized async toggle

kevineger:
To support toggling while authorized (reacting to a synchronous event), turning `on` asynchronously is a bit of a challenge. This abstract class solve that problem generically. This will be consumed in a downstream PR which refreshes the connection for servers.

### DIFF
--- a/src/common/toggleable.ts
+++ b/src/common/toggleable.ts
@@ -50,26 +50,7 @@ export abstract class AsyncToggleable<T extends Disposable>
       return;
     }
     this.inFlightTurnOn = new AbortController();
-    this.initializationComplete = this.initialize(this.inFlightTurnOn.signal)
-      .then((resource) => {
-        if (!this.inFlightTurnOn || this.inFlightTurnOn.signal.aborted) {
-          log.trace("Initialization aborted, disposing resource");
-          resource.dispose();
-          return;
-        }
-        this.resource = resource;
-      })
-      .catch((err: unknown) => {
-        if (!this.inFlightTurnOn || this.inFlightTurnOn.signal.aborted) {
-          log.trace("Initialization aborted by error");
-        } else {
-          log.error(`Unable to initialize ${this.constructor.name}`, err);
-        }
-        throw err;
-      })
-      .finally(() => {
-        this.inFlightTurnOn = undefined;
-      });
+    this.initializationComplete = this.doInit(this.inFlightTurnOn.signal);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/common/toggleable.ts
```diff
@@ -18,3 +21,79 @@ export interface Toggleable {
...
+        if (!this.inFlightTurnOn || this.inFlightTurnOn.signal.aborted) {
+          log.trace("Initialization aborted, disposing resource");
+          resource.dispose();
+          return;
+        }
+        this.resource = resource;
+      })
+      .catch((err: unknown) => {
+        if (!this.inFlightTurnOn || this.inFlightTurnOn.signal.aborted) {
+          log.trace("Initialization aborted by error");
```
amtoney524: would it be helpful to log the error that aborted the initialization?

src/common/toggleable.ts
```diff
@@ -18,3 +21,79 @@ export interface Toggleable {
...
+   * No-ops if already on.
+   * If `off()` is called during initialization, it is aborted.
+   */
+  on() {
+    if (this.inFlightTurnOn) {
+      return;
+    }
+    this.inFlightTurnOn = new AbortController();
+    this.initializationComplete = this.initialize(this.inFlightTurnOn.signal)
+      .then((resource) => {
```
amtoney524: nit: not for this pr, but would it be worth refactoring to async/await? (preferred by internal style guide)

src/common/toggleable.unit.test.ts
```diff
@@ -0,0 +1,202 @@
...
+  aborted: Promise<void>;
+}
+
+class TestToggleable extends AsyncToggleable<Disposable> {
+  readonly initializeStub = sinon.stub<[AbortSignal], Promise<Disposable>>();
+  // Promote to public for async hooks in tests.
+  declare initializationComplete;
+
+  nextInitialization(): TestInitialization {
+    let abortResolver: () => void;
```
amtoney524, line 28: will be nice when we can use `Promise.withResolvers`!

kevineger: No kidding! Hope to have time for that soon.

src/common/toggleable.ts
kevineger: Yes!

src/common/toggleable.ts
kevineger: Done!

PR #179, opened 2025-10-31
feat: refresh server connections

kevineger:


src/colab/connection-refresher.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  private handleAssignmentChange(e: AssignmentChangeEvent) {
+    // New servers.
+    for (const s of e.added) {
+      this.scheduleRefresh(s);
+    }
+
+    // Updated servers.
+    for (const s of e.changed) {
```
EvanWiederspan, line 133: It took me a while to understand that this code relies on the fact that assignedChanged is triggered from the Assignments manager after a successful refresh, in order to schedule the next refresh after a successful one - could you add a comment here noting that?

src/colab/connection-refresher.unit.test.ts
```diff
@@ -0,0 +1,381 @@
...
+        expect(
+          refreshAbortSignal?.aborted,
+          "Call to refresh servers should be aborted",
+        ).to.be.true;
+      });
+    });
+
+    describe("dispose", () => {
+      it("aborts in-flight refreshes", async () => {
+        // TODO
```
EvanWiederspan: Intentionally left undone?

src/colab/connection-refresher.unit.test.ts
```diff
@@ -0,0 +1,381 @@
...
+              ...server.connectionInformation,
+              tokenExpiry: new Date(Date.now() + REFRESH_MS),
+            },
+          };
+          assignmentsChangeEmitter.fire({
+            added: [],
+            changed: [server],
+            removed: [],
+          });
+          return Promise.resolve(server);
```
EvanWiederspan, line 270: Don't think you need .resolve since it's a async func

src/colab/connection-refresher.ts
kevineger: Can do. Had a comment in the tests, but yeah would be better here in the context of the code.

Image

src/colab/connection-refresher.ts
```diff
@@ -0,0 +1,221 @@
...
+  AssignmentChangeEvent,
+  AssignmentManager,
+} from "../jupyter/assignments";
+import { ColabAssignedServer } from "../jupyter/servers";
+import { NotFoundError } from "./client";
+
+/* The buffer we give to refresh the token, before it actually expires. */
+const REFRESH_BUFFER_MS = 5 * 60 * 1000; // 5 minutes.
+
+/*
```
EvanWiederspan: nit: /**

src/colab/connection-refresher.unit.test.ts
kevineger: `callsFake` must return a `Promise<ColabAssignedServer>`.

src/colab/connection-refresher.unit.test.ts
kevineger: Woops, thanks. Done.

src/colab/connection-refresher.ts
kevineger: Added as method comment.

src/colab/connection-refresher.ts
kevineger: Done.

PR #180, opened 2025-11-01
feat: support abort signals for assignment

kevineger:
Also throws a known `NotFoundError` when refreshing a server which isn't tracked. This change is unrelated... sorry. Given the size figured this was okay.

While I could add a bunch of tests/asserts that that the abort signals are passed to the `ColabClient`, I couldn't find a way to do this which I liked that didn't feel unnecessary and bloaty. Please feel free to push on this if you feel it's worthwhile.

### DIFF
--- a/src/jupyter/assignments.unit.test.ts
+++ b/src/jupyter/assignments.unit.test.ts
@@ -934,7 +934,7 @@ describe("AssignmentManager", () => {
   });
 
   describe("refreshConnection", () => {
-    it("throws a not found error when refreshing a server that's not tacked", async () => {
+    it("throws a not found error when refreshing a server that's not tracked", async () => {
       await expect(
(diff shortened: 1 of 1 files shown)


src/jupyter/assignments.unit.test.ts
```diff
@@ -933,59 +934,69 @@ describe("AssignmentManager", () => {
   });
 
   describe("refreshConnection", () => {
-    const newToken = "new-token";
-    let refreshedServer: ColabAssignedServer;
+    it("throws a not found error when refreshing a server that's not tacked", async () => {
```
amtoney524: nit: tracked?

kevineger: ty!

PR #181, opened 2025-11-01
fix: trigger notification to reload notebook more generically

amtoney524:
This change notifies the user to reload any notebooks connected to the removed server.

Previously, this was associated with the remove server command, which did not cover other cases where servers may be removed (e.g., reconciliation)

### DIFF
--- a/src/jupyter/assignments.unit.test.ts
+++ b/src/jupyter/assignments.unit.test.ts
@@ -220,6 +220,7 @@ describe("AssignmentManager", () => {
 
       sinon.assert.notCalled(vsCodeStub.commands.executeCommand);
       sinon.assert.notCalled(assignmentChangeListener);
+      sinon.assert.notCalled(vsCodeStub.window.showInformationMessage);
     });
 
     it("does nothing when no servers need reconciling", async () => {
@@ -230,6 +231,7 @@ describe("AssignmentManager", () => {
 
       sinon.assert.notCalled(vsCodeStub.commands.executeCommand);
       sinon.assert.notCalled(assignmentChangeListener);
+      sinon.assert.notCalled(vsCodeStub.window.showInformationMessage);
     });
 
     it("reconciles a single assigned server when it is the only one", async () => {
@@ -251,6 +253,10 @@ describe("AssignmentManager", () => {
         removed: [{ server: defaultServer, userInitiated: false }],
         changed: [],
       });
+      sinon.assert.calledOnceWithMatch(
+        vsCodeStub.window.showInformationMessage,
+        sinon.match(/notebooks Colab GPU A100 was/),
+      );
     });
 
     describe("with multiple servers", () => {
@@ -262,6 +268,7 @@ describe("AssignmentManager", () => {
           defaultServer,
           {
             ...defaultServer,
+            label: "Second Server",
             id: randomUUID(),
             endpoint: "m-s-bar",
             connectionInformation: {
@@ -302,10 +309,18 @@ describe("AssignmentManager", () => {
           removed: [{ server: servers[1], userInitiated: false }],
           changed: [],
         });
+        sinon.assert.calledOnceWithMatch(
+          vsCodeStub.window.showInformationMessage,
+          sinon.match(/notebooks Second Server was/),
+        );
       });
 
       it("reconciles multiple assigned servers when all need reconciling", async () => {
-        await serverStorage.store(servers);
+        const threeServers = [
+          ...servers,
+          { ...defaultServer, label: "Third Server" },
+        ];
+        await serverStorage.store(threeServers);
         colabClientStub.listAssignments.resolves([]);
 
         await assignmentManager.reconcileAssignedServers();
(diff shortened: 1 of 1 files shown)


src/jupyter/assignments.ts
```diff
@@ -486,6 +490,30 @@ export class AssignmentManager implements vscode.Disposable {
...
   }
+
+  private async notifyReloadNotebooks(e: AssignmentChangeEvent) {
+    const numRemoved = e.removed.length;
+    if (numRemoved === 0) {
+      return;
+    }
+
+    const removed = e.removed.map((r) => r.server.label);
+    const serverDescriptor =
```
amtoney524, line 501: open to other suggestions for the multiple server case!

src/jupyter/assignments.unit.test.ts
```diff
@@ -1114,6 +1115,138 @@ describe("AssignmentManager", () => {
       ).to.eventually.equal(defaultServer.label);
     });
   });
+
+  describe("onDidAssignmentsChange", () => {
+    let assignmentChangeEmitter: TestEventEmitter<AssignmentChangeEvent>;
+
+    beforeEach(() => {
+      /* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access */
```
amtoney524, line 1161: I know this isn't ideal, but this was the simplest way I found to test. lmk if there's a different pattern we should consider here!

kevineger: Take a look at how I used the same emitter [here](https://github.com/googlecolab/colab-vscode/blob/3df0500c619f93e067088485f7a49d75ac1b7f72/src/colab/connection-refresher.unit.test.ts#L104) recently.

src/jupyter/assignments.unit.test.ts
```diff
@@ -1114,6 +1115,138 @@ describe("AssignmentManager", () => {
...
+        assignmentChangeEmitter.fire({
+          added: [],
+          removed: [
+            {
+              server: { ...defaultServer, label: "server A" },
+              userInitiated: false,
+            },
+          ],
+          changed: [],
+        });
```
kevineger: Should we lift this into the `beforeEach` of _and the notification to reload notebooks is shown_?

src/jupyter/assignments.unit.test.ts
```diff
@@ -1114,6 +1115,138 @@ describe("AssignmentManager", () => {
...
+        await Promise.resolve();
```
kevineger: Rather that rely on breaking the event loop, can we instead create a `new Promise` that resolves when a fake implementation of `vsCodeStub.window.showInformationMessage` is called? There should be a few examples of this technique in my recent PRs.

src/jupyter/assignments.unit.test.ts
```diff
@@ -1114,6 +1115,138 @@ describe("AssignmentManager", () => {
       ).to.eventually.equal(defaultServer.label);
     });
   });
+
+  describe("onDidAssignmentsChange", () => {
```
kevineger: Since we aren't directly testing the `onDidAssignmentsChange` method, consider naming the top level block _when assignments change_ or something generic.

src/jupyter/assignments.ts
```diff
@@ -486,6 +490,30 @@ export class AssignmentManager implements vscode.Disposable {
...
+      removed.length === 1
+        ? `${removed[0]} was`
+        : `${removed.slice(0, numRemoved - 1).join(", ")} and ${removed[numRemoved - 1]} were`;
+    const viewIssue = await this.vs.window.showInformationMessage(
+      `Sorry for the inconvenience, to work around [microsoft/vscode-jupyter #17094](https://github.com/microsoft/vscode-jupyter/issues/17094) - please re-open notebooks ${serverDescriptor} previously connected to.`,
```
```suggestion
      `To work around [microsoft/vscode-jupyter #17094](https://github.com/microsoft/vscode-jupyter/issues/17094) - please re-open notebooks ${serverDescriptor} previously connected to.`,
```

I think your previous recommendation of simplifying the verbiage will help inform users at a glance that this isn't an _error_ caused by something they did nor the Colab extension.

amtoney524: simplified!

src/jupyter/assignments.unit.test.ts
amtoney524: done!

src/jupyter/assignments.unit.test.ts
amtoney524: yup, updated the tests to use this pattern!

src/jupyter/assignments.unit.test.ts
amtoney524: done!

src/jupyter/assignments.unit.test.ts
amtoney524: synced offline - the same pattern couldn't be used here since the event listener we're testing wouldn't be triggered with a mocked assignments module; instead, we moved assertions that the notification is shown into existing blocks for `reconcileAssignedServers` and `unassignServer`. This should be temporary, contigent on https://github.com/microsoft/vscode-jupyter/issues/17094

PR #186, opened 2025-11-03
test: upload output logs as workflow artifact

amtoney524:
When the e2e test fails, uploads the outputs logs as artifacts

See [this workflow run](http://github.com/googlecolab/colab-vscode/actions/runs/19049133116/job/54404751510) as an example

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -163,7 +163,7 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@v4
         with:
-          name: exthost log
+          name: exthost_log
           path: /tmp/test-resources/settings/logs/*/window1/exthost/exthost.log
       - name: upload output logs
         if: failure()


.github/workflows/build-test.yml
```diff
@@ -165,3 +165,9 @@ jobs:
         with:
           name: exthost log
           path: /tmp/test-resources/settings/logs/*/window1/exthost/exthost.log
+      - name: upload output logs
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: output_logs
```
kevineger, line 172: Could we be consistent one way or another with the `with.name`? E.g. `exthost log` above but `output_logs` here. FWIW I prefer the latter.

amtoney524: good point! renamed the "upload exthost.log" step artifact  to `exhost_log` since only one log file's uploaded in that step

PR #190, opened 2025-11-05
docs: centralize contributing prereqs and document test setup

kevineger:
- Doc updates for brevity and explaining the test + build setup.
- Fixed _recommendations_ for extensions to match doc.
- Set `.env.template` default environment to `production`.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -78,7 +78,7 @@ Unit tests can be debugged by launching _Debug Unit Tests_. The `"args"` can be 
 
 The unit tests rely extensively on a [Sinon](https://sinonjs.org/) stub of the `vscode` module. `vscode` is the _engine_ the Node app runs under and is not an installable npm package. While this stubbing is terse, it enables the entire unit test suite to run sub-1-second. Please peruse existing tests for reference when authoring new ones.
 
-While there are end-to-end tests, running them requires access to accounts with _automation_ exemptions that only Googlers can use. Pull request automation (once approved to be ran by a maintainer) will run them.
+While there are end-to-end tests, running them requires access to accounts with _automation_ exemptions that only Googlers can use. Pull request automation (once approved by a maintainer) will run them.
 
 ### Code Reviews
 


docs/contributing.md
```diff
@@ -25,41 +25,60 @@ Guidelines](https://opensource.google/conduct/).
...
+Incremental build output can be found in the _Terminal_ VS Code panel.
 
-#### Configure your environment
+TypeScript errors and warnings can be found in the _Problems_ VS Code panel, provided a _typecheck_-ing build is running (see [`package.json`](https://github.com/googlecolab/colab-vscode/blob/main/package.json)).
 
-Create OAuth 2.0 _Desktop_ client credentials ([instructions](https://developers.google.com/identity/protocols/oauth2)).
+If you've removed files, you may need to `npm run clean`.
 
-Make a copy of the environment template: `cp .env.template .env`
+#### Tests
```
amtoney524, line 68: nit/optional: worth subheadings for unit and e2e sections?

docs/contributing.md
```diff
@@ -25,41 +25,60 @@ Guidelines](https://opensource.google/conduct/).
...
+Unit tests can be debugged by launching _Debug Unit Tests_. The `"args"` can be modified to filter with the `grep` field as outlined above.
 
-#### Run the extension
+The unit tests rely extensively on a [Sinon](https://sinonjs.org/) stub of the `vscode` module. `vscode` is the _engine_ the Node app runs under and is not an installable npm package. While this stubbing is terse, it enables the entire unit test suite to run sub-1-second. Please peruse existing tests for reference when authoring new ones.
 
-1. Open the repo root with VS Code.
-1. Launch the extension by pressing `F5` or selecting `Run Extension` from VS Code's _Run and Debug_ view.
-1. Create or open a Jupyter notebook file (`.ipynb`).
-1. Test and validate your changes.
+End-to-end tests must have the extension `.e2e.test.ts`. Before they can be ran, the `TEST_ACCOUNT_EMAIL` and `TEST_ACCOUNT_PASSWORD` must be set (in the `.env` file). They can be ran with `npm run test:e2e` or `npm run test:e2e:headless` or debugged by launching _Debug E2E Tests_.
```
amtoney524: because we don't expose the demo account to users, would we suggest they test this with their own creds (or would that potentially trigger abuse)? if not, maybe guidance is that they can run the build and test workflow with their branch?

kevineger: Good call. Modified to make note that while there are e2e tests, running them is restricted to Googlers and PR automation.

docs/contributing.md
kevineger: Four headings is already a bit deep IMO. Considering we're changing the e2e section to be more of a disclaimer, hopefully this is fine. Please re-open if you disagree!

PR #191, opened 2025-11-06
feat: add setting to enable LSP

amtoney524:
This PR defines a setting to enable LSP. 

Extension settings:
Screenshot 2025-11-06 at 6 35 53 PM

Filtered settings:
Screenshot 2025-11-06 at 6 36 03 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,10 +43,10 @@
     "configuration": {
       "title": "Colab",
       "properties": {
-        "colab.codeDiagnostics.LSP": {
+        "colab.codeDiagnostics": {
           "type": "boolean",
           "default": false,
-          "description": "Enable language server protocol (LSP) features when connected to a Colab server."
+          "description": "Enable code diagnostics powered by a connected Colab server."
         },
         "colab.logging.level": {
           "type": "string",


package.json
```diff
@@ -43,6 +43,11 @@
     "configuration": {
       "title": "Colab",
       "properties": {
+        "colab.codeDiagnostics.LSP": {
```
amtoney524: as a starting point, "Code Diagnostics" came from Colab Web's Editor settings

Image

kevineger: I would suspect the vast majority of users don't know what "language server protocol (LSP)" is. I'd like to name the setting (and have a description as) something which generically represents the functionality.

E.g. _Code diagnostics_ - _Enable code diagnostics powered by a connected Colab server._

@EvanWiederspan may have some suggestions too.

Should look at the verbiage the _Python_ and _Pylance_ extensions use. Or perhaps other language server extensions?

amtoney524: nice suggestion, updated to that as a starting point! I think you're right about not using the term "LSP" - I can only find one setting that references that

here's an example from the pylance extension where type checking modes are exposed
Screenshot 2025-11-06 at 6 44 19 PM

PR #196, opened 2025-11-10
feat: colab language client for latest server

EvanWiederspan:
LSP Client that uses the latest assigned runtime as the LSP server


kevineger:
Breaking work across two distinct PRs that I'll open shortly.

PR #202, opened 2025-11-11
test: add safeClick method

amtoney524:
The `waitAndClick` method is used in the E2E test's OAuth flow.

This has recently increased in flakiness, throwing `StaleElementReferenceError` when interacting with the consent page.

As a solution, we've renamed the method to `safeClick` and refactored it to use a `webdriver.wait` so that the flow is retried if it does not succeed

### DIFF
--- a/src/test/extension.e2e.test.ts
+++ b/src/test/extension.e2e.test.ts
@@ -303,7 +303,7 @@ async function safeClick(
         if (e instanceof error.StaleElementReferenceError) {
           return false;
         }
-        return false;
+        throw e;
       }
     },
     ELEMENT_WAIT_MS,


src/test/extension.e2e.test.ts
```diff
@@ -288,11 +288,24 @@ async function waitAndClick(
...
   errorMsg: string,
 ): Promise<void> {
-  await driver.wait(
-    until.elementIsVisible(await driver.findElement(locator)),
-    ELEMENT_WAIT_MS,
-    errorMsg,
-  );
-  const element = await driver.findElement(locator);
-  await element.click();
+  // Attempt up to 3 times to account for the page to settle.
```
kevineger: Would it be better to use `driver.wait` with a lambda function?

---

Currently we're awaiting `findElement` before passing the result to the wait condition. That means we find the element, pass that specific one to the `wait` function, but if the page refreshes during the wait, the reference would be stale. Calling `findElement` again inside the success block would also create a second opportunity for staleness.

---

If we want to fix this as a one-off, we could pass a function directly to `driver.wait`. If it throws the stale error (which I think we can `instanceof` check? `import { error } from "selenium-webdriver";`), we can catch it and return `false` to trigger a retry on the next poll cycle. Something like:

```ts
await driver.wait(async () => {
  try {
    const element = await driver.findElement(locator);
    
    if (await element.isDisplayed() && await element.isEnabled()) {
        await element.click();
        return true;
    }
    return false;
  } catch (e) {
    if (e instanceof error.StaleElementReferenceError) {
      return false;
    }
    throw e;
  }
}, ELEMENT_WAIT_MS, errorMsg);
```

I think this is better than the `while` loops since the location/visibility-check/click is atomic, it relies on Selenium's optimal polling scheduling and it handles Selenium errors directly which will also throw `TimeoutError`-s which remove the need for custom error messages when retries/polls exceed threshold.

---

Alternatively (and better if you have bandwidth for it), we could make this a more reusable util and generalize it as something like:

```ts
async function clickSafe(driver: WebDriver, locator: Locator, timeout = 10000) { ... }
```

If able to do the latter, we should probably put all our web clicks behind it?

amtoney524: thanks a lot for the suggestion, placing this inside a wait makes sense to me! The only place we use `click` is during the OAuth flow; so I went with the `clickSafe` util and updated those references accordingly.

With `clickSafe`, I kept the custom error argument so that we can see where in the flow we timed out. Additionally, I did not add the timeout as an arg, since we're not currently setting that based on the call-site - though happy to go back and adjust if needed

src/test/extension.e2e.test.ts
```diff
@@ -281,18 +283,30 @@ async function notebookLoaded(driver: WebDriver): Promise<void> {
...
+        if ((await element.isDisplayed()) && (await element.isEnabled())) {
+          await element.click();
+          return true;
+        }
+        return false;
+      } catch (e) {
+        if (e instanceof error.StaleElementReferenceError) {
+          return false;
+        }
+        return false;
```
kevineger: All branches `return false;`. Shouldn't we throw `e` if it isn't a `StaleElementReferenceError`?

kevineger (approved):
Thanks so much for tackling this! Should be a great improvement to test robustness!

src/test/extension.e2e.test.ts
amtoney524: thanks for the catch!

Issue #207, opened 2025-11-13
Colab Kernel is not available in VS code after revoking OAuth app access


gogasca:
## Description

I revoked OAuth app access in my Google Account, now Colab is not displayed in Kernel selection

## Environment

```
Version: 1.105.1 (Universal)
Commit: 7d842fb85a0275a4a8e4d7e040d2625abbf7f084
Date: 2025-10-14T22:33:36.618Z
Electron: 37.6.0
ElectronBuildId: 12502201
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin x64 24.6.0
```

## Repro Steps

1. Click on Photo profile from logged in user.
2. Go to Managed your Google Account > Data & Privacy > Third party & services > Google Colaboratory

<img width="2146" height="1516" alt="Image" src="https://github.com/user-attachments/assets/023aaccb-7964-43d7-becd-f3c147f6daab" />

## Expected Behaviour

I get prompted again.

## Actual Behaviour

Colab kernel is no longer displayed and extension becomes unsuable

## Logs

Logs from `Colab` in the `Output` panel (`View` > `Output`, change the drop-down the upper-right of the `Output` panel to `Colab`).

```txt
[2025-11-13T07:22:25.531Z] [Info] Visual Studio Code: 1.105.1
[2025-11-13T07:22:25.531Z] [Info] Remote: N/A
[2025-11-13T07:22:25.531Z] [Info] App Host: desktop
[2025-11-13T07:22:25.531Z] [Info] Jupyter extension version: 2025.9.1
[2025-11-13T07:22:32.547Z] [Error] Unable to refresh access token
                                   Error: invalid_grant
                                       at vo._request (/Users/gogasca/.vscode/extensions/google.colab-0.1.0/out/extension.js:12:3454)
                                       at processTicksAndRejections (node:internal/process/task_queues:105:5)
                                       at t.refreshTokenNoCache (/Users/gogasca/.vscode/extensions/google.colab-0.1.0/out/extension.js:24:10713)
                                       at t.refreshAccessTokenAsync (/Users/gogasca/.vscode/extensions/google.colab-0.1.0/out/extension.js:24:11428)
                                       at Na.initialize (/Users/gogasca/.vscode/extensions/google.colab-0.1.0/out/extension.js:64:1928)
                                       at rk (/Users/gogasca/.vscode/extensions/google.colab-0.1.0/out/extension.js:66:24409)
[...]

PR #218, opened 2025-11-14
fix: handle when OAuth access is revoked

amtoney524:
When initializing the auth provider, we now clear sessions when encountering a 400 error.

This error can be encountered when the user revokes OAuth app access to Colaboratory. See https://github.com/googlecolab/colab-vscode/issues/207 for additional details on how to reproduce.

Now, when the user reloads VS Code, sessions with an invalid grant are cleared, allowing creation of a new session.

### DIFF
--- a/src/auth/auth-provider.ts
+++ b/src/auth/auth-provider.ts
@@ -128,7 +128,11 @@ export class GoogleAuthProvider
       let shouldClearSession = false;
       let reason = "";
 
-      if (err instanceof GaxiosError && err.status === 400) {
+      if (
+        err instanceof GaxiosError &&
+        err.status === 400 &&
+        err.message.includes("invalid_grant")
+      ) {
         reason = "OAuth app access to Colab was revoked";
         shouldClearSession = true;
       } else if (err instanceof GaxiosError && err.status === 401) {
--- a/src/auth/auth-provider.unit.test.ts
+++ b/src/auth/auth-provider.unit.test.ts
@@ -257,7 +257,7 @@ describe("GoogleAuthProvider", () => {
           });
         }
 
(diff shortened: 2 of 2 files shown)


src/auth/auth-provider.ts
```diff
@@ -125,13 +125,21 @@ export class GoogleAuthProvider
...
-      // switches the OAuth client ID / secret.
-      if (err instanceof GaxiosError && err.status === 401) {
-        log.warn(
-          "The configured OAuth client has changed. Clearing session.",
-          err,
-        );
+      let shouldClearSession = false;
+      let reason = "";
+
+      if (err instanceof GaxiosError && err.status === 400) {
```
kevineger: I'm a little worried about only looking at the status being 400. Could mask other user input errors. Is there something on the structured error we can check to confirm this is for an `invalid_grant`?

amtoney524: good point! updated the condition to check that the `message` includes `invalid_grant`

src/auth/auth-provider.ts
```diff
@@ -125,13 +125,25 @@ export class GoogleAuthProvider
...
+        err.message.includes("invalid_grant")
+      ) {
+        reason = "OAuth app access to Colab was revoked";
+        shouldClearSession = true;
+      } else if (err instanceof GaxiosError && err.status === 401) {
+        // This should only ever be the case when developer building from source
+        // switches the OAuth client ID / secret.
+        reason = "The configured OAuth client has changed";
+        shouldClearSession = true;
+      }
```
kevineger, line 143: Totally optional, but readability may be improved if we lifted to a helper function.

amtoney524: chatted offline - going to merge without the helper but can revisit!

PR #274, opened 2025-11-26
feat: add the ability to remove external servers

hjjackyang:
* **Why**? Currently, the UX to remove servers initiated outside VS Code is cumbersome. Also, when there are both VS Code and remotely assigned servers, the existing UX can only show VS Code servers.
* The new added functionality can show all assigned servers and allow users to choose one to remove all in one place within VS Code.
* [Screencast](https://screencast.googleplex.com/cast/NTI2ODQwMDg1NDAwNzgwOHwwOGJhZmFmMy1lZg)
* [Reference bug](https://b.corp.google.com/issues/462727891)

### DIFF
--- a/src/colab/client.ts
+++ b/src/colab/client.ts
@@ -528,7 +528,3 @@ class ColabRequestError extends Error {
     this.responseBody = responseBody;
   }
 }
-
-export const TEST_ONLY = {
-  TUN_ENDPOINT,
-};
--- a/src/colab/client.unit.test.ts
+++ b/src/colab/client.unit.test.ts
@@ -30,7 +30,6 @@ import {
   DenylistedError,
   InsufficientQuotaError,
   TooManyAssignmentsError,
-  TEST_ONLY,
 } from "./client";
 import {
   ACCEPT_JSON_HEADER,
@@ -595,7 +594,7 @@ describe("ColabClient", () => {
             urlMatcher({
               method: "GET",
               host: COLAB_HOST,
-              path: `${TEST_ONLY.TUN_ENDPOINT}/${assignedServer.endpoint}/api/sessions`,
+              path: `/tun/m/${assignedServer.endpoint}/api/sessions`,
               otherHeaders: {
                 [COLAB_TUNNEL_HEADER.key]: COLAB_TUNNEL_HEADER.value,
               },
--- a/src/colab/commands/server.ts
+++ b/src/colab/commands/server.ts
@@ -122,7 +122,3 @@ enum ServerCategory {
 interface RemoveServerItem extends QuickPickItem {
   value?: ColabAssignedServer | UnownedServer;
 }
-
-export const TEST_ONLY = {
-  ServerCategory,
-};
--- a/src/jupyter/assignments.ts
+++ b/src/jupyter/assignments.ts
@@ -457,7 +457,7 @@ export class AssignmentManager implements vscode.Disposable {
 
     await this.storage.clear();
     await this.storage.store(reconciled);
-    await this.assignmentChange.fire({
+    this.assignmentChange.fire({
       added: [],
       removed: removed.map((s) => ({ server: s, userInitiated: false })),
       changed: [],
@@ -602,7 +602,3 @@ function colabProxyFetch(
 function isRequest(info: RequestInfo): info is Request {
   return typeof info !== "string" && !("href" in info);
 }
-
-export const TEST_ONLY = {
-  UNKNOWN_REMOTE_SERVER_NAME,
-};
--- a/src/jupyter/assignments.unit.test.ts
+++ b/src/jupyter/assignments.unit.test.ts
@@ -33,11 +33,7 @@ import { TestEventEmitter } from "../test/helpers/events";
 import { ServerStorageFake } from "../test/helpers/server-storage";
 import { newVsCodeStub, VsCodeStub } from "../test/helpers/vscode";
 import { isUUID } from "../utils/uuid";
-import {
-  AssignmentChangeEvent,
-  AssignmentManager,
-  TEST_ONLY,
-} from "./assignments";
+import { AssignmentChangeEvent, AssignmentManager } from "./assignments";
 import {
   ColabAssignedServer,
   ColabServerDescriptor,
@@ -521,7 +517,7 @@ describe("AssignmentManager", () => {
           accelerator: "",
         },
         {
-          label: TEST_ONLY.UNKNOWN_REMOTE_SERVER_NAME,
+          label: "Untitled",
           endpoint: endpointWithoutSession,
           variant: Variant.DEFAULT,
           accelerator: "",
@@ -564,13 +560,13 @@ describe("AssignmentManager", () => {
             accelerator: "",
           },
           {
-            label: TEST_ONLY.UNKNOWN_REMOTE_SERVER_NAME,
+            label: "Untitled",
             endpoint: endpointWithoutName,
             variant: Variant.DEFAULT,
             accelerator: "",
           },
           {
-            label: TEST_ONLY.UNKNOWN_REMOTE_SERVER_NAME,
+            label: "Untitled",
             endpoint: endpointWithoutSession,
             variant: Variant.DEFAULT,
             accelerator: "",
--- a/src/jupyter/servers.ts
+++ b/src/jupyter/servers.ts
@@ -44,8 +44,10 @@ export type ColabAssignedServer = ColabJupyterServer & {
   readonly dateAssigned: Date;
 };
 
-export function isColabAssignedServer(obj: object): obj is ColabAssignedServer {
-  return "connectionInformation" in obj;
+export function isColabAssignedServer(
+  s: ColabAssignedServer | UnownedServer,
+): s is ColabAssignedServer {
+  return "connectionInformation" in s;
 }
 
 export const DEFAULT_CPU_SERVER: ColabServerDescriptor = {
--- a/src/colab/commands/server.unit.test.ts
+++ b/src/colab/commands/server.unit.test.ts
@@ -19,7 +19,7 @@ import {
 } from "../../test/helpers/quick-input";
 import { newVsCodeStub, VsCodeStub } from "../../test/helpers/vscode";
 import { Variant } from "../api";
-import { removeServer, renameServerAlias, TEST_ONLY } from "./server";
+import { removeServer, renameServerAlias } from "./server";
 
 describe("Server Commands", () => {
   let vsCodeStub: VsCodeStub;
@@ -199,18 +199,18 @@ describe("Server Commands", () => {
           {
             label: defaultServer.label,
             value: defaultServer,
-            description: TEST_ONLY.ServerCategory.VS_CODE,
+            description: "VS Code Server",
           },
           {
             label: additionalVsCodeServer.label,
             value: additionalVsCodeServer,
-            description: TEST_ONLY.ServerCategory.VS_CODE,
+            description: "VS Code Server",
           },
           { label: "", kind: -1 },
           {
             label: nonVsCodeServer.label,
             value: nonVsCodeServer,
-            description: TEST_ONLY.ServerCategory.COLAB_WEB,
+            description: "Colab Web Server",
           },
         ]);
       });
@@ -229,12 +229,12 @@ describe("Server Commands", () => {
           {
             label: defaultServer.label,
             value: defaultServer,
-            description: TEST_ONLY.ServerCategory.VS_CODE,
+            description: "VS Code Server",
           },
           {
             label: additionalVsCodeServer.label,
             value: additionalVsCodeServer,
-            description: TEST_ONLY.ServerCategory.VS_CODE,
+            description: "VS Code Server",
           },
         ]);
       });
@@ -257,7 +257,7 @@ describe("Server Commands", () => {
           {
             label: nonVsCodeServer.label,
             value: nonVsCodeServer,
-            description: TEST_ONLY.ServerCategory.COLAB_WEB,
+            description: "Colab Web Server",
           },
         ]);
       });


src/jupyter/servers.ts
```diff
@@ -48,3 +53,12 @@ export const DEFAULT_CPU_SERVER: ColabServerDescriptor = {
...
   variant: Variant.DEFAULT,
 };
+
+export function isInstanceOfColabAssignedServer(
+  obj: object,
+): obj is ColabAssignedServer {
+  // `connectionInformation` currently only exists in `ColabAssignedServer` but
+  // not in `ColabRemoteServer`, hence using it to identify if an object is
+  // `ColabAssignedServer`.
+  return "connectionInformation" in (obj as ColabAssignedServer);
```
kevineger: - Nit: code is self describing, these kind of comments can go stale.
- Nit: unecessary cast

```suggestion
  return "connectionInformation" in obj;
```

src/jupyter/servers.ts
```diff
@@ -48,3 +53,12 @@ export const DEFAULT_CPU_SERVER: ColabServerDescriptor = {
   label: "Colab CPU",
   variant: Variant.DEFAULT,
 };
+
+export function isInstanceOfColabAssignedServer(
```
kevineger: Nit: type guards are typically just `isFoo`.

```suggestion
export function isColabAssignedServer(
```

src/jupyter/assignments.ts
```diff
@@ -186,6 +189,34 @@ export class AssignmentManager implements vscode.Disposable {
     }));
   }
 
+  /**
+   * Retrieves the list of servers that have been created remotely (outside VS
+   * Code).
+   */
+  async getRemoteServers(signal?: AbortSignal): Promise<ColabRemoteServer[]> {
```
kevineger: Sorry to be pedantic, but the word _Remote_ is tripping me up a bit. Technically all Colab servers we deal with in the extension are "remote".

Could we rename this here, and elsewhere (e.g. `server.ts`) to `getUnownedServers` (`UnownedServer`), or something else that reflects these are Colab servers that aren't managed by the extension? Other potential choices: `getExternalServers`/`ExternalServer`, `getNonVsCodeServers`/`NonVsCodeServer`.

src/jupyter/assignments.ts
```diff
@@ -186,6 +189,34 @@ export class AssignmentManager implements vscode.Disposable {
...
+    const stored = new Set(
+      (await this.storage.list()).map((server) => server.endpoint),
+    );
+    return Promise.all(
+      (await this.client.listAssignments(signal))
+        .filter((assignment) => !stored.has(assignment.endpoint))
+        .map(async (assignment) => {
+          // For any remote servers created in Colab web UI, assuming there is
+          // only one session per assignment.
```
kevineger, line 194: I think this assumption is reasonable. I think the only time `sessions[0]` won't exist is the slim race condition where the server has been created but not quite connected to.

src/jupyter/assignments.ts
```diff
@@ -353,29 +384,37 @@ export class AssignmentManager implements vscode.Disposable {
   /**
    * Unassigns the given server.
    *
-   * Deletes all kernel sessions for the specified server before
-   * unassigning. Only unassigns if all session deletions succeed.
+   * For VS Code initiated servers, deletes all kernel sessions for the
```
```suggestion
   * For servers created in VS Code, deletes all kernel sessions for the
```

src/jupyter/assignments.ts
```diff
@@ -567,3 +598,7 @@ function colabProxyFetch(
 function isRequest(info: RequestInfo): info is Request {
   return typeof info !== "string" && !("href" in info);
 }
+
+export const TEST_ONLY = {
+  UNKNOWN_REMOTE_SERVER_NAME,
+};
```
kevineger: Nit: I'd much rather redefine `"Untitled"` in test than have exported _test only_ members in prod code.

src/jupyter/assignments.unit.test.ts
```diff
@@ -477,6 +482,112 @@ describe("AssignmentManager", () => {
...
+      ]);
+      colabClientStub.listSessions.withArgs(endpointWithoutName).resolves([
+        {
+          ...defaultSession,
+          name: "",
+        },
+      ]);
+      colabClientStub.listSessions
+        .withArgs(endpointWithoutSession)
+        .resolves([]);
```
kevineger: Rather than define all of the test variants in the `describe` and setup each of their matching stubs in a `beforeEach`, could we instead have the test tell the full story? It's okay if in some we need to do "duplicate" setup, to keep the tests fully descriptive (internal guidance go/tott/643).

Right now just reading the test body, it doesn't tell the full story, which forces readers to bounce between the setup blocks here and the test code.

src/jupyter/assignments.unit.test.ts
```diff
@@ -737,7 +829,7 @@ describe("AssignmentManager", () => {
       sinon.assert.notCalled(assignmentChangeListener);
     });
 
-    describe("when the server exists", () => {
+    describe("when VS Code server exists", () => {
```
kevineger: Nit

```suggestion
    describe("when the server created in VS Code exists", () => {
```

src/jupyter/assignments.unit.test.ts
```diff
@@ -737,7 +829,7 @@ describe("AssignmentManager", () => {
       sinon.assert.notCalled(assignmentChangeListener);
     });
```
kevineger, line 863: Should we have a similar test to `"does nothing when the server does not exist"`, but for a non-VS-Code server?

src/jupyter/assignments.unit.test.ts
```diff
@@ -815,6 +907,23 @@ describe("AssignmentManager", () => {
         );
       });
     });
+
+    describe("when a remote Colab server exists", () => {
+      it("unassigns server", async () => {
```
kevineger: Nit: full sentence + consistency with above.

```suggestion
      it("unassigns the server", async () => {
```

src/colab/client.ts
```diff
@@ -276,28 +276,37 @@ export class ColabClient {
...
    *
-   * @param server - The server to list sessions for.
+   * @param serverOrEndpoint - The server or assignment endpoint to list
+   *   sessions for.
    * @param signal - Optional {@link AbortSignal} to cancel the request.
    * @returns The list of sessions.
    */
   async listSessions(
-    server: ColabAssignedServer,
+    serverOrEndpoint: ColabAssignedServer | string,
```
kevineger, line 287: Totally your call, just something to consider:

I generally try and avoid `fooOrBar` parameters when the param represents two very different things (e.g. cases that are not like `url: URL | string`). That leaves overloading or separate functions as alternatives.

I think overloading is unecessary. It'll give a slightly better autocomplete hint, but requires 3 near identical lines just to shave off the `|` in lints.

Splitting the function could make the functional difference a bit more clear for callers, e.g.:

- `listServerSessions`
- `listEndpointSessions`

Again, your call. What you have is also fine, just a minor nit to consider.

src/colab/client.ts
```diff
@@ -519,3 +528,7 @@ class ColabRequestError extends Error {
     this.responseBody = responseBody;
   }
 }
+
+export const TEST_ONLY = {
```
kevineger: Same feedback w.r.t. exporting just for tests.

src/colab/client.unit.test.ts
```diff
@@ -590,6 +591,63 @@ describe("ColabClient", () => {
...
+              [COLAB_TUNNEL_HEADER.key]: COLAB_TUNNEL_HEADER.value,
+            },
+            withAuthUser: false,
+          }),
+        )
+        .resolves(
+          new Response(
+            withXSSI(
+              JSON.stringify([
+                {
```
kevineger: Consider putting the `listSessions` tests in a `describe` and defining a `const mockSession` once.

src/colab/commands/server.ts
```diff
@@ -67,19 +68,39 @@ export async function removeServer(
   assignmentManager: AssignmentManager,
   withBackButton?: boolean,
 ) {
-  const servers = await assignmentManager.getAssignedServers();
-  if (servers.length === 0) {
+  const vsCodeServers = await assignmentManager.getAssignedServers();
+  const colabRemoteServers = await assignmentManager.getRemoteServers();
```
kevineger: Here, we're doing a redundant call to list all servers. `assignmentManager.getAssignedServers` lists them from the API, and then filters based on what's stored. `assignmentManager.getRemoteServers` will then list them again.

To avoid doubling calls and expensive calls, we should probably make `getServers` method which does this in one pass. This can be achieved cleanly with:

```
export interface AllServers {
  extension: ColabAssignedServer[];
  external: ColabRemoteServer[];
}

// In AssignmentManager:

getServers(
  from: "extension",
  signal?: AbortSignal,
): Promise<ColabAssignedServer[]>;
getServers(
  from: "external",
  signal?: AbortSignal,
): Promise<ColabRemoteServer[]>;
getServers(from: "all", signal?: AbortSignal): Promise<AllServers>;
getServers(
  from: "extension" | "external" | "all",
  signal?: AbortSignal,
): Promise<ColabAssignedServer[] | ColabRemoteServer[] | AllServers> {
  // TODO
}
```

src/jupyter/assignments.ts
hjjackyang: Going with "unowned" to avoid renaming the existing "assigned" server nomenclature.

src/jupyter/assignments.ts
hjjackyang: Ack! The slim chance of no session is also handled by this code gracefully.

src/colab/client.ts
hjjackyang: I actually initially went with overloading, but VS Code (or the linter) suggested me to remove unnecessary overloading and just do it this way.

That being said, we can split the method signature as you suggested, but I feel that's a bit of overkill, so prefer to keep this as is for now.

src/colab/client.ts
hjjackyang: Wouldn't this be better than hardcoding the same value(s) in tests over and over again?

http://go/typescript-testing#export-visibility

src/jupyter/assignments.unit.test.ts
hjjackyang: I don't feel it's necessary in this case because when this notification shows up, it means there will always be a server (whether it is a VSCode or non-VSCode server).

src/colab/commands/server.ts
hjjackyang: As discussed offline, I changed `getUnownedServers` to `getAllServers` to avoid the double calls for now. Please review again. Thanks!

src/jupyter/assignments.ts
hjjackyang: Replied in the other `TEST_ONLY` related comment.

src/colab/client.ts
kevineger: I'd prefer if we spelled it out explicitly in tests for two reasons:

1. Keeps production code clean and free of "test only" exports.
1. Focuses tests strictly on the public API.
1. `path: "/tun/m/foo/keep-alive/"` reads a lot nicer than `path: \`${TEST_ONLY.TUN_ENDPOINT}/foo/keep-alive/"` at a glance.

---

Also as is, the `TEST_ONLY.TUN_ENDPOINT` is only used in the new tests added, with all others spelling out the `"/tun/m"` string.

src/jupyter/servers.ts
```diff
@@ -44,7 +44,25 @@ export type ColabAssignedServer = ColabJupyterServer & {
   readonly dateAssigned: Date;
 };
 
+export function isColabAssignedServer(obj: object): obj is ColabAssignedServer {
```
kevineger: Narrow `object` to what you need to discriminate.

As is, this type guard would pass for an object that was missing all the other properties of a `ColabAssignedServer` or with `"connectionInformation"` that didn't match. E.g.:

```json
{
    "connectionInformation": {
        "not actually connection info": 123456789,
        "foo": "bar"
    }
}
```

```suggestion
export function isColabAssignedServer(s: ColabAssignedServer | UnownedServer): obj is ColabAssignedServer {
```

hjjackyang:
@kevineger - Friendly reminder to give this PR a second pass :)

src/jupyter/assignments.unit.test.ts
kevineger: In terms of the remove notification sure, but the behaviour of this unit likely shouldn't assume that's the only use case.

src/jupyter/assignments.ts
```diff
@@ -415,6 +451,34 @@ export class AssignmentManager implements vscode.Disposable {
     return `${labelBase} (${placeholderIdx.toString()})`;
   }
 
+  private async reconcileStoredServers(
+    storedServers: ColabAssignedServer[],
```
kevineger, line 441: If we don't refactor (which would make lifting `reconcileStoredServers` unecessary), shouldn't we read `storedServers` from the member, `this.storage.list`?

src/jupyter/assignments.ts
```diff
@@ -186,6 +170,50 @@ export class AssignmentManager implements vscode.Disposable {
     }));
   }
 
+  /**
+   * Retrieves the list of all servers that are assigned both in and outside VS
+   * Code.
+   */
+  async getAllServers(signal?: AbortSignal): Promise<AllServers> {
```
kevineger, line 177: Removing the duplicate call to the API is fixed, but the current factoring duplicates the work of creating a set of stored servers and iterating it to create the two disjoint collections (stored vs. unowned).

I do think overloading here makes a ton of sense, to both clean up the API and reduce the duplicate work.

```ts
async getServers(from: "extension"): Promise<ColabAssignedServer>;
async getServers(from: "external"): Promise<UnownedServer>;
async getServers(from: "all"): Promise<AllServers>;
async getServers(
  from: "extension" | "external" | "all",
): Promise<ColabAssignedServer | UnownedServer | AllServers> {
  // TODO
}
```

I won't block on this change, but if we don't tackle it here I'll probably follow up with it.

Of course this can just as easily be done with individual methods, but I think we should push the actual logic into a central place to remove/reduce the duplicate work.

src/jupyter/assignments.unit.test.ts
kevineger: Re-opening, mind addressing?

kevineger (approved):
Thanks Jack! Please consider addressing comments. If not, I can take a pass once this is in.

src/colab/client.ts
hjjackyang: That's fair. Removed all `TEST_ONLY` exports.

src/jupyter/servers.ts
hjjackyang: Done!

src/jupyter/assignments.unit.test.ts
hjjackyang: Wait, I did address this and moved data arrange into each individual test. Which part do you feel I missed?

src/jupyter/assignments.unit.test.ts
hjjackyang: Perhaps I misunderstood this block. This block / unit specifically targets `notifyMaxAssignmentsExceeded()` by throwing a `TooManyAssignmentsError` in `beforeEach`. And the updated implementation of `notifyMaxAssignmentsExceeded()` does not care whether there are servers or not (it doesn't check it anymore), so `"does nothing when the server does not exist"` won't apply here.

src/jupyter/assignments.ts
hjjackyang: Don't we want to reconcile the stored servers when we call `getAllServers` (which is the existing behavior of `getAssignedServers`)?

src/jupyter/assignments.ts
hjjackyang: Ack. I'll merge this PR and do this refactoring in a follow-up PR.

src/jupyter/assignments.unit.test.ts
hjjackyang: As discussed offline, I feel the current tests have the right amount of abstraction, with some basic test objects created outside test body, and choosing which one(s) to use/stub inside test body. Resolving.

src/jupyter/assignments.ts
hjjackyang: As discussed offline, there is an early return in `reconcileAssignedServers` if `stored.length === 0`. If I move the `this.storage.list` call into this refactored `reconcileStoredServers` method, the outer `reconcileAssignedServers` method will make unnecessary call to `listAssignments` even if there are no stored servers. As a result, the current way is probably more network efficient.

PR #277, opened 2025-11-27
feat: include in auth events if there's a valid active session

kevineger:
This is used in downstream changes to simplify eventing based on the presence of a valid auth session.


src/auth/auth-provider.ts
```diff
@@ -170,6 +180,7 @@ export class GoogleAuthProvider
       added: [],
       removed: [],
       changed: [this.session],
+      hasValidSession: !!this.session,
```
amtoney524: nit/optional: would it be worth explicitly setting to true / false in create / remove functions?

kevineger: Not hard feelings either way, changed!

PR #279, opened 2025-11-27
refactor: simplify async toggling

kevineger:
I wasn't happy with `AsyncToggleable`. There were a few things I didn't like, but the fact that it is generic (`<T extends Disposable>`) highlights the biggest issue. It conflated resource lifecycle with execution lifecycle.

By making it singly-responsible for execution lifecycle, the code reads more easily and yields flexibility to callers who can control when and how their resources are managed. Gone is the implicit indirection.

This is achieved by adding `LatestCancelable` which simply looks after running one async task at a time by cancelling the in-flight one. I feel that this pattern feels natural when you look at the call-sites, e.g.: `this.runner.run("my", "params")`. When ran from a sync context, as is often the case from sync event listeners, the `void` makes this _runner_ more obvious -  we're fire-and-forgetting these async _runs_.

I added a flexible test helper / spy (`ControllableAsyncToggle`), used to make writing `AsyncToggle`-based unit tests a bit less painful.

### DIFF
--- a/src/test/helpers/async.ts
+++ b/src/test/helpers/async.ts
@@ -1,0 +1,164 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { AsyncToggle } from "../../common/toggleable";
+
+/**
+ * A simple Deferred promise helper.
+ */
+export class Deferred<T> {
+  readonly promise: Promise<T>;
+  resolve: (value: T | PromiseLike<T>) => void;
+  reject: (reason?: unknown) => void;
+
+  constructor() {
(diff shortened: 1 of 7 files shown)


src/test/helpers/async.ts
```diff
@@ -0,0 +1,164 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { AsyncToggle } from "../../common/toggleable";
+
+/**
+ * A simple Deferred promise helper.
+ */
+export class Deferred<T> {
```
amtoney524, line 12: super useful, thanks!

kevineger: No sweat!

PR #286, opened 2025-12-03
build: add some helpful Mocha test snippets

kevineger:
https://code.visualstudio.com/docs/editing/userdefinedsnippets

### DIFF
--- a/.vscode/mocha-test.code-snippets
+++ b/.vscode/mocha-test.code-snippets
@@ -1,0 +1,26 @@
+{
+  "beforeEach": {
+    "scope": "typescript",
+    "prefix": ["beforeEach", "be"],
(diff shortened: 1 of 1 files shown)


hjjackyang (approved):
Could you add some description to PR explaining how these snippets can be used?

Also, out of curiosity, why did we choose Mocha over Jasmine for unit test framework?

kevineger:
> Could you add some description to PR explaining how these snippets can be used?
> 
> Also, out of curiosity, why did we choose Mocha over Jasmine for unit test framework?

Sure!

Mocha's better integrated into VS Code for the extension-tester ([docs](https://code.visualstudio.com/api/working-with-extensions/testing-extension)) and is also more prevalent across VS Code extensions and [in general](https://www.npmjs.com/package/mocha).

PR #292, opened 2025-12-04
refactor: `getAssignedServers()` and `getAllServers()` to `getServers()` overloads

hjjackyang:
This is a follow-up to https://github.com/googlecolab/colab-vscode/pull/274.

### DIFF
--- a/src/jupyter/assignments.unit.test.ts
+++ b/src/jupyter/assignments.unit.test.ts
@@ -429,6 +429,19 @@ describe("AssignmentManager", () => {
           expect(stripFetches(servers)).to.deep.equal(storedServers);
         });
 
+        it("reconciles assigned servers before returning", async () => {
+          colabClientStub.listAssignments.resolves([defaultAssignment]);
+          const noLongerAssignedServer = {
+            ...defaultServer,
+            endpoint: "no-longer-assigned",
+          };
+          await serverStorage.store([defaultServer, noLongerAssignedServer]);
+
+          const results = await assignmentManager.getServers("extension");
+
+          expect(stripFetches(results)).to.deep.equal([defaultServer]);
+        });
+
         it("includes a fetch implementation that attaches Colab connection info", async () => {
           const servers = await assignmentManager.getServers("extension");
           assert.lengthOf(servers, 1);


src/jupyter/assignments.ts
```diff
@@ -152,66 +152,85 @@ export class AssignmentManager implements vscode.Disposable {
   }
 
   /**
-   * Retrieves the list of servers that have been assigned in VS Code.
+   * Retrieves the list of servers that have been assigned in VS Code extension.
```
kevineger: Nit: grammar

```suggestion
   * Retrieves the list of servers that have been assigned in the VS Code
   * extension.
```

kevineger: Apply "**the** VS Code extension" below too (in both other doc-blocks).

Though I think simply "VS Code" or "the extension" should suffice.

src/jupyter/assignments.ts
```diff
@@ -152,66 +152,85 @@ export class AssignmentManager implements vscode.Disposable {
...
+    signal?: AbortSignal,
+  ): Promise<ColabAssignedServer[] | UnownedServer[] | AllServers> {
     let storedServers = await this.storage.list();
-    if (storedServers.length > 0) {
-      storedServers = await this.reconcileStoredServers(
-        storedServers,
-        allAssignments,
-      );
+    if (from === "extension" && storedServers.length === 0) {
+      return storedServers;
```
kevineger, line 187: This introduces a bug where stored servers are never reconciled, no?

I believe in **every** case we should be reconciling. I think we can:

- get rid of the private `reconcileStoredServers` 
- revert the public `reconcileAssignedServers` to what it was before #274
- call it at the beginning of this method, regardless of what `from` is

Beyond the potential bug I mentioned, reconciling in all cases will prevent us from future potential pains where servers become accessible or transferable between the extension and external (where currently, we skip reconciling).

kevineger: This also highlights that there's a missing test case (which predates this change) to validate calls to `getAssignedServers` (now `getServers("extension")`) are reconciled. Mind adding one?

src/colab/keep-alive.unit.test.ts
```diff
@@ -273,7 +288,10 @@ describe("ServerKeepAliveController", () => {
       >;
 
       beforeEach(() => {
-        assignmentStub.getAssignedServers.resolves([defaultServer]);
+        // Type assertion needed due to overloading on getServers
+        (assignmentStub.getServers as sinon.SinonStub)
+          .withArgs("extension", sinon.match.any)
```
kevineger: The `signal` matcher (`sinon.match.any`) can be omitted here. Please apply globally.

```suggestion
          .withArgs("extension")
```

src/jupyter/assignments.ts

src/colab/keep-alive.unit.test.ts
hjjackyang: Done!

src/jupyter/assignments.ts
hjjackyang: > This also highlights that there's a missing test case (which predates this change) to validate calls to `getAssignedServers` (now `getServers("extension")`) are reconciled. Mind adding one?

Yes, I can add this.


> This introduces a bug where stored servers are never reconciled, no?

This actually aligns with the original public `reconcileAssignedServers` implementation where stored servers are not reconciled if there is none.

See:
https://github.com/googlecolab/colab-vscode/blob/09f56df337ea1eacbca53aa3edcfbb2031a1f75c/src/jupyter/assignments.ts#L134

Could you please elaborate more on what you think the bug was? Currently, it only skips reconciling servers if we request for servers from `extension` and there is no stored server to reconcile.

kevineger: Ah whoops, I missed the `&& storedServers.length === 0` on L186. Sorry! Monday morning 😴

hjjackyang: No worries! I also just added the test case to cover server reconcile when calling `getServers` from `extension`.

PR #294, opened 2025-12-05
feat: allow high-ram machine shapes for pro users

mounikasomisetty:
This PR introduces the ability to select a machine shape as part of the server picker flow. 

Changes include:
1. Add `shape` as a parameter in the client to create/get assignments with a specified shape
2. Pass the `SubscriptionTier` via the provider to the `AssignmentManager` method to allow it populate the available descriptors based on tier. For non-pro users, the functionality has not changed. 
3. Update the `ServerPicker` flow to account for the new step to select a machine shape

Here's a [screencast](https://screencast.googleplex.com/cast/NjU3ODg3MjIzMjQ0MzkwNHxhMDllZjMxZC1kMg) of things working for me as a non-pro user and here's [one](https://screencast.googleplex.com/cast/NDk4MDc4MTg3NjI0ODU3Nnw5MTU5NTAyMS03MQ) with things working with an OTA as a pro user


src/colab/api.ts
```diff
@@ -391,3 +391,40 @@ export function variantToMachineType(variant: Variant): string {
...
+      return "Standard";
+  }
+}
+
+const HIGHMEM_ONLY_ACCELERATORS: Set<string> = new Set<string>([
+  "L4",
+  "V28",
+  "V5E1",
+  "V6E1",
+]);
```
kevineger, line 415: Could you file a bug for these to be returned from the API? Unlikely to happen in the near future, but would like to not maintain disjoint lists.

src/colab/api.ts
```diff
@@ -391,3 +391,40 @@ export function variantToMachineType(variant: Variant): string {
...
+ *
+ * @param accelerator - The accelerator to check.
+ * @returns Whether the accelerator requires a high-memory machine shape.
+ */
+export function isHighMemOnlyAccelerator(accelerator?: string): boolean {
+  if (!accelerator) {
+    return false;
+  }
+
+  return HIGHMEM_ONLY_ACCELERATORS.has(accelerator);
```
kevineger: Nit: can optionally be simplified as:

```suggestion
  return (
    accelerator !== undefined && HIGHMEM_ONLY_ACCELERATORS.has(accelerator)
  );
```

src/colab/client.ts
```diff
@@ -528,3 +538,12 @@ class ColabRequestError extends Error {
     this.responseBody = responseBody;
   }
 }
+
+function mapShapeToURLParam(shape: Shape): string {
+  switch (shape) {
+    case Shape.HIGHMEM:
+      return "hm";
+    default:
+      return "";
```
kevineger: Wouldn't this set the URL param to `shape=` (no value)? Shouldn't we omit the param if shape != high mem?

kevineger: Ah, I see that we only call this when `shape !== Shape.STANDARD`. May instead make sense to either: 

1. Change the block in `buildAssignUrl` to only set if `shape` is high-mem.
2. Change this function to `buildShapeParam(shape: Shape): URLSearchParams | undefined`.

I'm a bit partial to the latter, since it'll guard us if we ever add more shapes (still has the `switch` and we can remove `default` or check exhaustively).

Your call! Was just a bit confused since there's a dead path here given that the logic is split.

src/colab/server-picker.ts
```diff
@@ -94,16 +114,23 @@ export class ServerPicker {
     // Skip prompting for an accelerator for the default variant (CPU).
     if (state.variant === Variant.DEFAULT) {
       state.accelerator = "NONE";
-      return (input: MultiStepInput) => this.promptForAlias(input, state);
+      return (input: MultiStepInput) =>
+        this.promptForMachineShape(input, state, shapesByAccelerators);
```
kevineger: If I'm tracing this right, won't this prompt free users who only have default machine shapes to click `Standard` on each assignment? If that's right, could we instead only add the step when there's an option to choose between (default/high-mem)?

src/jupyter/assignments.unit.test.ts
```diff
@@ -143,45 +143,74 @@ describe("AssignmentManager", () => {
...
-          accelerator: "V6E1",
-        },
+        defaultGpuT4Descriptor,
+        defaultGpuA100Descriptor,
+        defaultTpuV5E1Descriptor,
+        defaultTpuV6E1Descriptor,
+      ]);
+    });
+
+    it("returns the default CPU and the eligible servers for pro users", async () => {
```
kevineger: Nice work on test refactor + addition!

src/colab/api.ts
mounikasomisetty: Done -  b/467101307

src/colab/client.ts
mounikasomisetty: Ah gotcha, I had option #1 before but I changed to this - I tweaked it to make it more readable!

src/colab/server-picker.ts
mounikasomisetty: Done

src/jupyter/assignments.unit.test.ts
mounikasomisetty: Thanks!

src/colab/server-picker.ts
```diff
@@ -126,23 +163,56 @@ export class ServerPicker {
...
+      buttons: [input.vs.QuickInputButtons.Back],
+    });
+    state.shape = pick.value;
+    if (!isShapeDefined(state)) {
+      return;
+    }
 
-    return (input: MultiStepInput) => this.promptForAlias(input, state);
+    return (input: MultiStepInput) =>
+      this.promptForAlias(input, state, /** totalSteps */ step + 1);
```
amtoney524: super nit: `/* totalSteps= */` to follow style guide for parameter name comments (ditto below!)

mounikasomisetty: Done!

mounikasomisetty:
Here's an updated [screencast](go/scrcast/NjQ4MDY3OTIxMzcyNzc0NHxhYThjNTdjYS00NA) for a pro user and [one](go/scrcast/NTEwOTY2NDM2NTY3NDQ5NnwyMTdhNDI5ZS00Yw) for a free user

src/colab/server-picker.ts
```diff
@@ -94,16 +114,33 @@ export class ServerPicker {
...
+      const shapePicks = getShapePicksFromAccelerator(
+        shapesByAccelerators,
+        state.accelerator,
+      );
+      if (shapePicks.length <= 1) {
+        state.shape =
+          shapePicks.length === 1 ? shapePicks[0].value : Shape.STANDARD;
+        return (input: MultiStepInput) =>
+          this.promptForAlias(input, state, /** totalSteps= */ 2);
+      }
```
kevineger, line 125: Is it possible to consolidate / refactor the duplicate logic between here and in `promptForAccelerator`? Only difference is the `totalSteps` value.

src/colab/server-picker.ts
```diff
@@ -126,23 +163,56 @@ export class ServerPicker {
...
+  private async promptForMachineShape(
+    input: MultiStepInput,
+    state: PartialServerWith<'variant'>,
+    items: ShapePick[],
+  ) {
+    const step = state.accelerator && state.accelerator !== 'NONE' ? 3 : 2;
+    const pick = await input.showQuickPick({
+      title: 'Select a machine shape',
+      step,
+      totalSteps: step + 1,
```
kevineger: Nit: Consider defining `const totalSteps = step + 1` below the definition of `step` to be used here and when calling `promptForAlias` (then we don't need the inline param comment).

src/colab/server-picker.ts
```diff
@@ -178,10 +249,35 @@ function isAcceleratorDefined(
...
+interface ShapePick extends QuickPickItem {
+  value: Shape;
+}
+
+function getShapePicksFromAccelerator(
+  shapesByAccelerators: Map<string, Set<Shape>>,
+  accelerator: string,
+): ShapePick[] {
+  const shapes = shapesByAccelerators.get(accelerator) ?? new Set();
+  const items: ShapePick[] = [];
```
kevineger: Nit: no need to new up the empty `Set`.

```suggestion
  const items: ShapePick[] = [];
  const shapes = shapesByAccelerators.get(accelerator);
  if (!shapes) {
    return items;
  }
```

src/jupyter/assignments.ts
```diff
@@ -105,6 +108,7 @@ export class AssignmentManager implements vscode.Disposable {
   // TODO: Consider communicating which machines are available, but not to the
   // user at their tier (in the "ineligible" list).
   async getAvailableServerDescriptors(
+    subscriptionTier = SubscriptionTier.NONE,
```
kevineger: Should we remove the default? Looks like the only caller is `ColabJupyterServerProvider.assignServer` which always passes it.

Nit: please update the doc-block.

src/jupyter/provider.unit.test.ts
```diff
@@ -486,6 +486,8 @@ describe('ColabJupyterServerProvider', () => {
         });
 
         it('completes assigning a server', async () => {
+          colabClientStub.getSubscriptionTier.resolves(SubscriptionTier.PRO);
+
```
kevineger, line 489: Nit: remove newline in favour of clearly separating arrange-act-assert blocks.

```suggestion
          colabClientStub.getSubscriptionTier.resolves(SubscriptionTier.PRO);
```

kevineger (approved):
**@mounikasomisetty such great work on this, thanks so much for taking it on!** 🥳 

Sorry I was slow to get back to you on this one, been slammed these last couple days!

src/colab/server-picker.ts
mounikasomisetty: I made an attempt to move some more logic over but I had an issue not being able to click back when not calling the next prompt step in the method directly (if that makes sense)

Issue #256, opened 2025-11-21
(Now supported!): Support mounting drive (`google.colab.drive.mount`)


ruzbeh13:
### Feature Request: Add Google Drive Integration in Colab for VS Code

First, thank you for providing the official Colab extension for VS Code — it’s extremely useful and a great workflow improvement.

### Problem
Currently, the extension does not provide a built-in way to access Google Drive when running a Colab kernel in VS Code.

In the classic Google Colab environment, users can mount Google Drive via:

```python
from google.colab import drive
drive.mount('/content/drive')
````

However, this API is not available in the VS Code extension environment because the `google.colab` package is not included on the remote kernel, and the extension does not expose a corresponding UI for mounting or browsing Drive.

### Requested Feature

Please add **native Google Drive integration** to the Colab for VS Code extension, similar to:

* A “Colab FS” panel showing:

  * `Session FS`
  * `Drive` (Google Drive)
* A button to **Sign in with Google** to grant Drive access
* Automatic mounting of Drive to a consistent path (e.g., `/drive/MyDrive/`)
* Full read/write access from Python code running on the Colab kernel

### Why this is important

* Many Colab workflows rely on files stored in Google Drive.
* Users migrating from classic Colab to VS Code expect the same file accessibility.
* It significantly improves data handling for ML training, image datasets, notebooks, and large file workflows.
* Without this feature, users have no straightforward way to work with Drive in VS Code.

### Additional notes

Even an initial minimal implementation (browse-only or read-only access) would greatly improve usability.
Thank you for considering this request!

AI-Cortex:
I also ran into the same issue. Looking forward to seeing Drive integration added in Colab for VS Code — this would be extremely useful for workflows that rely on Google Drive.

kevineger:
Thanks for the well structured feature request @ruzbeh13. We've been using https://github.com/googlecolab/colab-vscode/issues/215 as the canonical issue for this, and other Colab tools that aren't yet supported. But given how popular the request to add Drive support is, I'll leave this issue up as the canonical _Drive_ one, in hopes more folks ➕  it instead of filing a new issue.

> Looking forward to seeing Drive integration 

Us too! We're starting to work on it but it'll take a while to untangle some of our existing support which relies on functionality unique to the Colab web client.

saserbest:
✚

GinjaFam:
+++

aykanugur:
+++

jen6:
++++

[...]

PR #297, opened 2025-12-06
feat: show warning notification on `drive.mount`

hjjackyang:
* **Why**? While we work on implementing `drive.mount` for VS Code extension, this is a stop-gap solution to point users to our [workaround doc](https://github.com/googlecolab/colab-vscode/wiki/Known-Issues-and-Workarounds#drivemount), and also to let them know there is already an [open issue](https://github.com/googlecolab/colab-vscode/issues/256), so they don't file another one.
* [Screencast](https://screencast.googleplex.com/cast/NTI2MDg3MjA0NDc3MzM3Nnw5ZTg5MmFiOS0wNg)
* [Bug reference](https://b.corp.google.com/issues/461876404)
* Notes:
   * I initially experimented with faking an `execute_reply` message and displaying the warning as a kernel execution error. However, that didn't work out as how I expected, hence the current solution to show as a VS Code warning notification.
   * For code organization, I put this new warning logic in a separate file, so it's easier to 1) unit test, and 2) deprecate after we implement `drive.mount` for VS Code.

Preview of the warning notification message:
Drive mount warning

### DIFF
--- a/src/jupyter/drive-mount-warning.unit.test.ts
+++ b/src/jupyter/drive-mount-warning.unit.test.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { expect } from 'chai';
 import sinon from 'sinon';
 import { Uri } from 'vscode';
 import { newVsCodeStub, VsCodeStub } from '../test/helpers/vscode';
@@ -23,27 +24,34 @@ describe('warnOnDriveMount', () => {
     });
 
     it('shows warning notification', async () => {
-      const showWarningMessageStub = vsCodeStub.window
-        .showWarningMessage as sinon.SinonStub;
-      showWarningMessageStub.resolves(undefined);
+      const warningShown = new Promise<void>((resolve) => {
+        (vsCodeStub.window.showWarningMessage as sinon.SinonStub).callsFake(
+          (message: string) => {
+            expect(message).to.match(/drive.mount is not currently supported/);
+            resolve();
+            return Promise.resolve(undefined);
+          },
+        );
+      });
 
       warnOnDriveMount(vsCodeStub.asVsCode(), rawDriveMountMessage);
-      await flush();
 
-      sinon.assert.calledOnceWithMatch(
-        showWarningMessageStub,
-        /drive.mount is not currently supported/,
-      );
+      await expect(warningShown).to.eventually.be.fulfilled;
     });
 
     it('presents an action to view workaround', async () => {
-      (vsCodeStub.window.showWarningMessage as sinon.SinonStub).resolves(
-        'Workaround',
-      );
+      const warningShown = new Promise<void>((resolve) => {
+        (vsCodeStub.window.showWarningMessage as sinon.SinonStub).callsFake(
+          () => {
+            resolve();
+            return Promise.resolve('Workaround');
+          },
+        );
+      });
 
       warnOnDriveMount(vsCodeStub.asVsCode(), rawDriveMountMessage);
-      await flush();
 
+      await expect(warningShown).to.eventually.be.fulfilled;
       sinon.assert.calledOnceWithMatch(
         vsCodeStub.env.openExternal,
         sinon.match(function (url: Uri) {
@@ -56,13 +64,18 @@ describe('warnOnDriveMount', () => {
     });
 
     it('presents an action to view issue', async () => {
-      (vsCodeStub.window.showWarningMessage as sinon.SinonStub).resolves(
-        'GitHub Issue',
-      );
+      const warningShown = new Promise<void>((resolve) => {
+        (vsCodeStub.window.showWarningMessage as sinon.SinonStub).callsFake(
+          () => {
+            resolve();
+            return Promise.resolve('GitHub Issue');
+          },
+        );
+      });
 
       warnOnDriveMount(vsCodeStub.asVsCode(), rawDriveMountMessage);
-      await flush();
 
+      await expect(warningShown).to.eventually.be.fulfilled;
       sinon.assert.calledOnceWithMatch(
         vsCodeStub.env.openExternal,
         sinon.match(function (url: Uri) {


package.json
```diff
@@ -195,10 +195,12 @@
   "dependencies": {
     "glob": "^11.1.0",
     "google-auth-library": "^9.15.0",
+    "isomorphic-ws": "^4.0.1",
```
kevineger: Do we actually need this?

As I understand this package exists for isomorphism between `ws` in Node and `WebSocket` in browsers.

Our extension is a standard extension, which always runs on the VS Code server - https://code.visualstudio.com/docs/remote/vscode-server#_architecture. We don't publish a "Web Extension" - https://code.visualstudio.com/api/extension-guides/web-extensions.

Since we're never running in the browser, I don't _think_ we need compatibility with `WebSocket`. In other words, I think we can purely use `ws`. Sorry, this may require a bit of homework / research.

src/jupyter/assignments.ts
```diff
@@ -217,6 +219,10 @@ export class AssignmentManager implements vscode.Disposable {
         connectionInformation: {
           ...server.connectionInformation,
           fetch: colabProxyFetch(server.connectionInformation.token),
+          WebSocket: colabProxyWebSocket(
+            this.vs,
+            server.connectionInformation.token,
+          ),
         },
       }));
```
kevineger: Totally optional nit, this may read easier.

```ts
.map((server) => {
  const c = server.connectionInformation;
  return {
    ...server,
    connectionInformation: {
      ...c,
      fetch: colabProxyFetch(c.token),
      WebSocket: colabProxyWebSocket(this.vs, c.token),
    },
  };
});
```

src/jupyter/assignments.ts
```diff
@@ -651,3 +658,74 @@ function colabProxyFetch(
...
+    };
+  };
+
+  return class OurWebSocket extends WebSocketIsomorphic {
+    constructor(
+      url: string | URL,
+      protocols?: string | string[],
+      options?: WebSocketIsomorphic.ClientOptions,
+    ) {
+      super(url, protocols, addOurHeaders(options));
```
kevineger: _Our_ -> _Colab_

```suggestion
  // These custom headers are required for Colab's proxy WebSocket to work.
  const colabHeaders: Record<string, string> = {};
  colabHeaders[COLAB_RUNTIME_PROXY_TOKEN_HEADER.key] = token;
  colabHeaders[COLAB_CLIENT_AGENT_HEADER.key] = COLAB_CLIENT_AGENT_HEADER.value;

  const addColabHeaders = (
    options?: WebSocketIsomorphic.ClientOptions,
  ): WebSocketIsomorphic.ClientOptions => {
    options ??= {};
    options.headers ??= {};
    const headers: Record<string, string> = {
      ...options.headers,
      ...colabHeaders,
    };
    return {
      ...options,
      headers,
    };
  };

  return class ColabWebSocket extends WebSocketIsomorphic {
    constructor(
      url: string | URL,
      protocols?: string | string[],
      options?: WebSocketIsomorphic.ClientOptions,
    ) {
      super(url, protocols, addColabHeaders(options));
```

src/jupyter/assignments.ts
```diff
@@ -651,3 +658,74 @@ function colabProxyFetch(
...
+    options ??= {};
+    options.headers ??= {};
+    const headers: Record<string, string> = {
+      ...options.headers,
+      ...colabHeaders,
+    };
+    return {
+      ...options,
+      headers,
+    };
```
kevineger: Totally optional nit, single line for readability.

```suggestion
    return { ...options, headers };
```

src/jupyter/assignments.ts
```diff
@@ -651,3 +658,74 @@ function colabProxyFetch(
...
+    override send(
+      data: string,
+      options: {
+        mask?: boolean;
+        binary?: boolean;
+        compress?: boolean;
+        fin?: boolean;
+      },
+      cb?: (err?: Error) => void,
+    ): void;
```
kevineger: I'm pretty sure we don't need all of these overloads re-defined. Overriding only the implementation signature is sufficient.

The only reason I think we'd want to override the overloads is if we handled `OurWebSocket` directly in code and wanted the simplified call signatures. But considering we don't, just the impl. sig should be enough.

src/jupyter/assignments.ts
```diff
@@ -651,3 +658,74 @@ function colabProxyFetch(
...
+            compress?: boolean;
+            fin?: boolean;
+          }
+        | ((err?: Error) => void)
+        | undefined,
+      cb?: (err?: Error) => void,
+    ) {
+      void warnOnDriveMount(vs, data);
+
+      // This is required to override all `send` overload signatures:
```
kevineger: Nit: given the signature, if you didn't have this it'd complain no? Optionally consider removing comment.

_Argument of type [...] is not assignable to parameter of type [...]_

src/jupyter/drive-mount-warning.ts
```diff
@@ -0,0 +1,54 @@
...
+  };
+  if (
+    parsedJupyterMessage.header.msg_type === 'execute_request' &&
+    DRIVE_MOUNT_PATTERN.exec(parsedJupyterMessage.content.code)
+  ) {
+    await notifyDriveMountUnsupported(vs);
+  }
+}
+
+async function notifyDriveMountUnsupported(vs: typeof vscode): Promise<void> {
```
kevineger: Considering this function renders the warning message and opens external links, we'd never care to wait for the result. As such, it probably makes more sense to have this be a sync method to simplify callsites. As is, the `void` call requires readers to click in a figure out what async work is being done, to ensure sync mechanism aren't needed.

In other words, consider making this a synchronous function and chaining the selected action logic as a thenable.

src/jupyter/drive-mount-warning.ts
```diff
@@ -0,0 +1,54 @@
...
+    case DriveMountUnsupportedAction.VIEW_WORKAROUND:
+      vs.env.openExternal(vs.Uri.parse(DRIVE_MOUNT_WIKI_LINK));
+      break;
+    case DriveMountUnsupportedAction.VIEW_ISSUE:
+      vs.env.openExternal(vs.Uri.parse(DRIVE_MOUNT_ISSUE_LINK));
+      break;
+  }
+}
+
+const DRIVE_MOUNT_PATTERN = /drive\.mount\(.*\)/;
```
kevineger: Hopefully people don't write other code which looks like this. I think it'll be slim enough of a case not to be an issue. But unfortunately we (a) won't catch all with this (e.g. consider an aliased import) and (b) we might catch non-drive mount calls (rare...).

Should be fine!

---

We can be _slightly_ more precise with the regex tho.

```suggestion
const DRIVE_MOUNT_PATTERN = /drive\.mount\(.+\)/;
```

src/jupyter/drive-mount-warning.ts
```diff
@@ -0,0 +1,54 @@
...
+    parsedJupyterMessage.header.msg_type === 'execute_request' &&
+    DRIVE_MOUNT_PATTERN.exec(parsedJupyterMessage.content.code)
+  ) {
+    await notifyDriveMountUnsupported(vs);
+  }
+}
+
+async function notifyDriveMountUnsupported(vs: typeof vscode): Promise<void> {
+  const selectedAction = await vs.window.showWarningMessage(
+    `drive.mount() is not supported by Colab VS Code extension at the moment, and we are actively working on supporting it. Please see [our wiki](${DRIVE_MOUNT_WIKI_LINK}) for workaround and [this issue](${DRIVE_MOUNT_ISSUE_LINK}) for progress.`,
```
kevineger: Grammar + brevity

```suggestion
    `drive.mount is not currently supported in the extension. We're working on it! See the [wiki](${DRIVE_MOUNT_WIKI_LINK}) for workarounds and track this [issue](${DRIVE_MOUNT_ISSUE_LINK}) for progress.`,
```

src/jupyter/drive-mount-warning.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const DRIVE_MOUNT_PATTERN = /drive\.mount\(.*\)/;
+const DRIVE_MOUNT_ISSUE_LINK =
+  'https://github.com/googlecolab/colab-vscode/issues/256';
+const DRIVE_MOUNT_WIKI_LINK =
+  'https://github.com/googlecolab/colab-vscode/wiki/Known-Issues-and-Workarounds#drivemount';
+
+enum DriveMountUnsupportedAction {
+  VIEW_ISSUE = 'View Issue',
+  VIEW_WORKAROUND = 'View Workaround',
```
kevineger: Nit: _View_ may be redundant.

```suggestion
  VIEW_ISSUE = 'GitHub Issue',
  VIEW_WORKAROUND = 'Workaround',
```

src/jupyter/drive-mount-warning.ts
```diff
@@ -0,0 +1,54 @@
...
+ * an execute request containing `drive.mount()`.
+ */
+export async function warnOnDriveMount(
+  vs: typeof vscode,
+  rawJupyterMessage: string,
+): Promise<void> {
+  const parsedJupyterMessage = JSON.parse(rawJupyterMessage) as {
+    header: { msg_type: string };
+    content: { code: string };
+  };
```
kevineger: Isn't this `as` cast and raw access unsafe? Won't it cause unhandled errors? Shouldn't we use a type-guard, or better yet a Zod definition with safe parsing?

src/jupyter/drive-mount-warning.unit.test.ts
```diff
@@ -0,0 +1,95 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import sinon from 'sinon';
+import { Uri } from 'vscode';
+import { newVsCodeStub, VsCodeStub } from '../test/helpers/vscode';
+import { warnOnDriveMount } from './drive-mount-warning';
+
+describe('warnOnDriveMount', () => {
```
kevineger, line 13: Related to cast / type-guard / parsing, add a test for empty data, or other data. We want to make sure we don't throw in that case. Should no-op.

src/jupyter/assignments.unit.test.ts
```diff
@@ -1365,14 +1391,19 @@ describe('AssignmentManager', () => {
...
+  const { fetch: _f, WebSocket: _ws, ...c } = server.connectionInformation;
   return {
     ...server,
     connectionInformation: c,
   };
 }
 
-function stripFetches(servers: ColabAssignedServer[]): ColabAssignedServer[] {
-  return servers.map(stripFetch);
+function stripFetchesAndWebSockets(
```
kevineger: Nit: could we name this `stripNetworkOverrides`?

src/jupyter/assignments.unit.test.ts
```diff
@@ -496,6 +504,13 @@ describe('AssignmentManager', () => {
             }),
           });
         });
+
+        it('includes a custom WebSocket implementation', async () => {
+          const servers = await assignmentManager.getServers('extension');
+          assert.lengthOf(servers, 1);
+          const server = servers[0];
+          assert.isDefined(server.connectionInformation.WebSocket);
```
kevineger, line 504: Similar to what's done above, can we guard that this attaches the headers we require in order to use this websocket?

src/jupyter/assignments.unit.test.ts
```diff
@@ -863,6 +879,10 @@ describe('AssignmentManager', () => {
           }),
         });
       });
+
+      it('includes a custom WebSocket implementation', () => {
+        assert.isDefined(assignedServer.connectionInformation.WebSocket);
```
kevineger, line 875: Same here.

kevineger (changes requested):
Sorry it took me so long to review this, crazy week. Largely looking great!

src/jupyter/assignments.ts
kevineger: Some other questions:

- Can `type SuperSend = WebSocketIsomorphic['send'];` be used to simplify code as well as make the code a bit more future proof?
- What about `...args` spreading?

package.json
hjjackyang: Do we foresee supporting Web Extensions in the future? Perhaps it's better to be a bit future proof, and at the same time align with Jupyter extension implementation (which also uses `isomorphic-ws`)?

src/jupyter/assignments.ts
hjjackyang: Done

src/jupyter/assignments.ts
hjjackyang: Done

src/jupyter/drive-mount-warning.ts
hjjackyang: Done

src/jupyter/drive-mount-warning.unit.test.ts
hjjackyang: You are right! New tests do fail and I added the proper type guard.

src/jupyter/assignments.unit.test.ts
hjjackyang: Done. Renamed to `stripNetworkOverride` for a single server and `stripNetworkOverrides` for multiple servers.

src/jupyter/assignments.ts
hjjackyang: Done

src/jupyter/assignments.ts
hjjackyang: Looks like I'm able to override just the implementation signature without the various overloads.

Re `type SuperSend = WebSocketIsomorphic['send']`, I'm not familiar with how to use it to simplify the code. Would be great if you can suggest an example. Thanks!

src/jupyter/drive-mount-warning.ts
hjjackyang: Done

src/jupyter/drive-mount-warning.ts
hjjackyang: Done

src/jupyter/drive-mount-warning.ts
hjjackyang: Done

src/jupyter/drive-mount-warning.ts
hjjackyang: Done

src/jupyter/assignments.ts
hjjackyang: Done

src/jupyter/assignments.unit.test.ts
hjjackyang: I spent a couple of hours on this, but unable to make it work unfortunately. I wonder if it's actually worth testing this, considering our e2e tests will cover and ensure our custom WebSocket works as expected.

hjjackyang: As discussed offline, I tried `sinon.stub(WebSocketIsomorphic.prototype, 'send')` and it didn't work either, so I went with the discussed solution of extracting out `colabProxyWebSocket` function and passing in `typeof WebSocketIsomorphic` for easier stubbing.

Please kindly review again!

src/jupyter/colab-proxy-web-socket.ts
```diff
@@ -0,0 +1,77 @@
...
+import WebSocketIsomorphic from 'isomorphic-ws';
+import vscode from 'vscode';
+import {
+  COLAB_CLIENT_AGENT_HEADER,
+  COLAB_RUNTIME_PROXY_TOKEN_HEADER,
+} from '../colab/headers';
+import { warnOnDriveMount } from './drive-mount-warning';
+
+/**
+ * Returns a `WebSocket` class which extends `WebSocketIsomorphic`, adds Colab's
```
kevineger: Nit: consider `{@link ...}` comments where relevant.

src/jupyter/colab-proxy-web-socket.unit.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      address: string | URL | null,
+      protocols?:
+        | string
+        | string[]
+        | WebSocketIsomorphic.ClientOptions
+        | ClientRequestArgs,
+      options?: WebSocketIsomorphic.ClientOptions | ClientRequestArgs,
+    ) {
+      super(null); // Avoid real WS connection
+      if (address !== null) {
```
kevineger: Remove so that `TestWebSocket` _always_ validates the headers are present?

src/jupyter/drive-mount-warning.ts
```diff
@@ -10,45 +10,71 @@ import vscode from 'vscode';
...
   );
-  switch (selectedAction) {
-    case DriveMountUnsupportedAction.VIEW_WORKAROUND:
-      vs.env.openExternal(vs.Uri.parse(DRIVE_MOUNT_WIKI_LINK));
-      break;
-    case DriveMountUnsupportedAction.VIEW_ISSUE:
-      vs.env.openExternal(vs.Uri.parse(DRIVE_MOUNT_ISSUE_LINK));
-      break;
-  }
 }
```
kevineger, line 33: Nit: did you consider Zod?

This may be more readable and robust:

```ts
export const ExecuteRequestSchema = z.object({
  header: z.object({
    msg_type: z.literal("execute_request"),
  }),
  content: z.object({
    code: z.string(),
  }),
});
```

```ts
export function isExecuteRequest(
  message: unknown,
): message is JupyterExecuteRequestMessage {
  return ExecuteRequestSchema.safeParse(message).success;
}
```

src/jupyter/drive-mount-warning.unit.test.ts
```diff
@@ -23,20 +23,26 @@ describe('warnOnDriveMount', () => {
     });
 
     it('shows warning notification', async () => {
-      await warnOnDriveMount(vsCodeStub.asVsCode(), rawDriveMountMessage);
+      const showWarningMessageStub = vsCodeStub.window
+        .showWarningMessage as sinon.SinonStub;
+      showWarningMessageStub.resolves(undefined);
+
+      warnOnDriveMount(vsCodeStub.asVsCode(), rawDriveMountMessage);
+      await flush();
```
kevineger: In the cases where we expect "does nothing", the best we can do is `flush` then assert it wasn't called.

But where we're expecting the message, why not use proper async/await mechanisms? E.g. take a look at [this](https://github.com/googlecolab/colab-vscode/blob/84049083b68e7e3c78bd4c3ba6501455e99279ce/src/auth/flows/loopback.unit.test.ts#L148-L153https://github.com/googlecolab/colab-vscode/blob/84049083b68e7e3c78bd4c3ba6501455e99279ce/src/auth/flows/loopback.unit.test.ts#L148-L153) for reference.

src/jupyter/assignments.ts
```diff
@@ -212,13 +213,17 @@ export class AssignmentManager implements vscode.Disposable {
...
-        },
-      }));
+      ).map((server) => {
+        const c = server.connectionInformation;
+        return {
+          ...server,
+          connectionInformation: {
+            ...c,
+            fetch: colabProxyFetch(c.token),
+            WebSocket: colabProxyWebSocket(this.vs, c.token),
```
kevineger, line 223: This is another situation where it'd be great if we had experiment config. I get a little worried / anxious going 0 to 100% for overriding the `WebSocket` impl. Not actionable, just mentioning.

package.json
kevineger: It's very unlikely we'd ever ship as a non-Universal extension, `Web` nor platform specific.

src/jupyter/assignments.ts
kevineger: You could simplify the member signature by referencing the super's sig like that.

package.json
hjjackyang: Done

src/jupyter/assignments.ts
hjjackyang: Please excuse my lack of TS knowledge. I'm still not sure what the syntax of override would look like with `type SuperSend = WebSocketIsomorphic['send']`. Could you provide an example snippet?

src/jupyter/colab-proxy-web-socket.ts
hjjackyang: Done

src/jupyter/colab-proxy-web-socket.unit.test.ts
hjjackyang: Done

src/jupyter/drive-mount-warning.ts
hjjackyang: Done

src/jupyter/drive-mount-warning.unit.test.ts
hjjackyang: What's the proper async/await mechanism here when `warnOnDriveMount` is a sync function and there is no promise to be awaited?

src/jupyter/assignments.ts
kevineger: I was wondering if you could use the base member's signature typing to reduce definition clutter here. I don't know if you can or not.

But my other point is that I'm pretty sure you only need to `override` the implementing signature.

src/jupyter/drive-mount-warning.unit.test.ts
kevineger: Something like the reference I linked. E.g.:

```ts
const warningShown = new Promise<void>((r) => {
  showWarningMessageStub.callsFake(() r.resolve());
  return Promise.resolve(true);
});

warnOnDriveMount(vsCodeStub.asVsCode(), rawDriveMountMessage);

await expect(warningShown).to.be.eventually.fulfilled;
```

src/jupyter/assignments.ts
hjjackyang: SG! I've already updated it to only override the implementing signature.

src/jupyter/drive-mount-warning.unit.test.ts
hjjackyang: Done

Issue #280, opened 2025-11-27
Connecting to GPU and TPU is bugged


TariqNier:
here is what i do to use it: 
> Select Kernel 
> Pick Colab 
> Pick GPU
> Pick T4 
> Im prompted to enter an alias, i press enter to continue to ignore 
> It says "Detecting Kernel" , here it should connect 
> It goes back to saying "Select Kernel" like i did absolutely nothing
but the CPU option works normally

![Image](https://github.com/user-attachments/assets/2a4eb6a7-7a54-45da-a3e2-86771bb5702f)

codeNinja62:
is your vscode up-to-date?

TariqNier:
> is your vscode up-to-date?

yea i just checked. i even tried cursor ide it has the same exact issue

Priyansh4444:
Same issue here, but unsure if this is an issue with me going past the Free limit's quota.

czhou578:
Yes. It doesn't clearly say on the UI whether I have exceeded the Free limit's quota.

kevineger:
We warn when you're unable to assign another server (limit to 3) and need to delete one. We also have warnings when you eclipse your paid and free quota. These all should show as notifications. While it's possible there's some bug there, I'd like to rule out if you're hitting some other issue.

To best help, could you please provide the info as requested in the [bug template](https://github.com/googlecolab/colab-vscode/issues/new?template=bug_report.md): all the environment info, consistent repro steps and the `Colab` & `Jupyter` logs. If you can, bump the verbosity up for both (in VS Code settings) before repro'ing and collecting them.

Hansehart:
I have the issue too. I guess I reached the limit. However I get no notification and I worked before. I am on WSL2, Ubuntu 24.04, steps you see on the video.

VSC:
Version: 1.106.3 (user setup)
Commit: bf9252a2fb45be6893dd8870c0bf37e2e1766d61
Date: 2025-11-25T22:28:18.024Z
Electron: 37.7.0
ElectronBuildId: 12781156
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26100

I dont see any logs from the extensions

kevineger:
@Hansehart you don't see any logs in `Output > Colab` nor `Output > Jupyter`?

Hansehart:
[...]

PR #303, opened 2025-12-08
refactor: log assignServer() error for debugging

hjjackyang:
* **Why**? This is to help debug https://github.com/googlecolab/colab-vscode/issues/280 when we ask users to provide logs.
* Also logged the random UUID, so we can potentially use it to query our server logs.

### DIFF
--- a/src/jupyter/assignments.ts
+++ b/src/jupyter/assignments.ts
@@ -276,7 +276,7 @@ export class AssignmentManager implements vscode.Disposable {
         signal,
       ));
     } catch (error) {
-      log.error(`Failed assigning server ${id}`, error);
+      log.trace(`Failed assigning server ${id}`, error);
       // TODO: Consider listing assignments to check if there are too many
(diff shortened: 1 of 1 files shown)


src/jupyter/assignments.ts
```diff
@@ -275,6 +276,7 @@ export class AssignmentManager implements vscode.Disposable {
         signal,
       ));
     } catch (error) {
+      log.error(`Failed assigning server ${id}`, error);
```
kevineger: If placing here, can you downgrade this to a warning (or debug, or trace)? It's a user error if they've exceeded quota or are deny-listed and I've tried to reserve `log.error` for unanticipated issues with the extension.

That or simply move this above the `throw error` below.

hjjackyang: Changed to `log.trace` because I tried moving it to above `throw error` and it still logs those user errors.

PR #307, opened 2025-12-10
feat: implement the `ProxiedJupyterClient`

kevineger:
This client is a thin wrapper of the APIs the generated `JupyterClient` exposes. 

- Binds the RP token (auth) on each call (`Middleware`).
- Lazily constructs the underlying API objects
- Offers two factory functions (`static`):
  - `withStaticConnection`: creates a client with the server's current connection information. Useful for one-off queries.
  - `withRefreshingConnection`: creates a client that uses the server's updated connection information (new token as it changes). Useful for persistent / long-lived clients.

A downstream change will leverage this client extensively as we use the `ContentsApi` to enable the file system provider.


src/jupyter/client/index.ts
```diff
@@ -0,0 +1,183 @@
...
+    this.clientConfig = new Configuration({
+      basePath: basePath.toString(),
+      headers: {
+        [COLAB_CLIENT_AGENT_HEADER.key]: COLAB_CLIENT_AGENT_HEADER.value,
+      },
+      middleware: [new AddProxyToken(getProxyToken)],
+    });
+  }
+
+  get config() {
```
hjjackyang, line 71: Internally, Google styles discourage usage of getters and setters. Shall we consider exposing these as `get-` methods instead?

http://go/tsjs-style#classes-getters-and-setters

src/jupyter/client/index.unit.test.ts
```diff
@@ -0,0 +1,209 @@
+/**
```
hjjackyang, line 1: Given the tests in this file only test logic written in your `index.ts`, any reason we don't name the file `index.unit.test.ts` to make it more easily correlated with `index.ts`?

hjjackyang (changes requested):

src/jupyter/client/index.ts
kevineger: Internally our style guide cautions against the use of getters and setters because they are not optimized well by the JSCompiler (closure compiler). Getters and setters look like properties but act like functions, which makes it very hard for the closure compiler to analyze side effects and optimize safely. That coupled with aggressive property renaming often forces the compiler to give up on optimizing the object entirely (and produce "bloated code").

Here, we're building with `esbuild` which does not perform the aggresive property-renaming optimizations that make getters sub-optimal in Closure. In V8 (which is the underlying engine running the extension code), getters are optimized and use broadly.

I'd prefer to keep these, they read more nicely and aren't a problem like they are internally.

Here's a snippet of the relevant produced JS (before being minimized):

```js
var ProxiedJupyterClient = class _ProxiedJupyterClient {
  configApi;
  contentsApi;
  identityApi;
  kernelsApi;
  kernelspecsApi;
  sessionsApi;
  statusApi;
  terminalsApi;
  clientConfig;
  constructor(basePath, getProxyToken) {
    this.clientConfig = new Configuration({
      basePath: basePath.toString(),
      headers: {
        [COLAB_CLIENT_AGENT_HEADER.key]: COLAB_CLIENT_AGENT_HEADER.value
      },
      middleware: [new AddProxyToken(getProxyToken)]
    });
  }
  get config() {
    return this.configApi ??= new ConfigApi(this.clientConfig);
  }
  get contents() {
    return this.contentsApi ??= new ContentsApi(this.clientConfig);
  }
  get identity() {
    return this.identityApi ??= new IdentityApi(this.clientConfig);
  }
  get kernels() {
    return this.kernelsApi ??= new KernelsApi(this.clientConfig);
  }
  get kernelspecs() {
    return this.kernelspecsApi ??= new KernelspecsApi(this.clientConfig);
  }
  get sessions() {
    return this.sessionsApi ??= new SessionsApi(this.clientConfig);
  }
  get status() {
    return this.statusApi ??= new StatusApi(this.clientConfig);
  }
  get terminals() {
    return this.terminalsApi ??= new TerminalsApi(this.clientConfig);
  }
  /**
   * Initializes a {@link ProxiedJupyterClient} with static connection
   * information corresponding to the provided server. In other words, the proxy
   * token is used on all requests and is not refreshed.
   *
   * @param server - The Colab server to connect to.
   * @returns a {@link ProxiedJupyterClient} to the specified server.
   */
  static withStaticConnection(server) {
    return new _ProxiedJupyterClient(
      server.connectionInformation.baseUrl,
      () => Promise.resolve(server.connectionInformation.token)
    );
  }
  /**
   * Initializes a {@link ProxiedJupyterClient} with static connection
   * information corresponding to the provided server. In other words, the proxy
   * token is used on all requests and is not refreshed.
   *
   * Callers should call `dispose` when the proxied connection no longer needs
   * refreshing.
   *
   * If the server is removed while the client is still alive, it will be
   * disposed. In other words, the last known token will be used but subsequent
   * calls will likely error (404).
   *
   * @param server - The Colab server to connect to.
   * @returns a {@link ProxiedJupyterClient} to the specified server.
   */
  static withRefreshingConnection(server, changes) {
    return new RefreshingClient(server, changes);
  }
};
```

src/jupyter/client/index.unit.test.ts
kevineger: Done

src/jupyter/client/index.ts
hjjackyang: Thanks for clarifying! LGTM!

PR #308, opened 2025-12-10
refactor: use the generated jupyter client

kevineger:
This effectively migrates all `/api` requests to the generated Jupyter client.

Added `src/test/helpers/jupyter.ts` to facilitate stubbing.

Since we still need to be able to `listSession` on the `tun/m` endpoint (to get the non-VS servers) we still need to Zod definition. Added `z.ZodType` annotations to ensure consistency.

### DIFF
--- a/src/colab/client.ts
+++ b/src/colab/client.ts
@@ -9,7 +9,8 @@ import * as https from 'https';
 import fetch, { Request, RequestInit, Headers } from 'node-fetch';
 import { z } from 'zod';
 import { traceMethod } from '../common/logging/decorators';
-import { ColabAssignedServer } from '../jupyter/servers';
+import { JupyterClient } from '../jupyter/client';
+import { Session } from '../jupyter/client/generated';
 import { uuidToWebSafeBase64 } from '../utils/uuid';
 import {
   Assignment,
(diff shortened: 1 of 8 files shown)


src/colab/client.ts
```diff
@@ -251,61 +248,22 @@ export class ColabClient {
...
-      headers = { [COLAB_TUNNEL_HEADER.key]: COLAB_TUNNEL_HEADER.value };
-    } else {
-      const connectionInfo = serverOrEndpoint.connectionInformation;
-      url = new URL('api/sessions', connectionInfo.baseUrl.toString());
-      headers = {
-        [COLAB_RUNTIME_PROXY_TOKEN_HEADER.key]: connectionInfo.token,
-      };
-    }
+    const url = new URL(
+      `${TUN_ENDPOINT}/${endpoint}/api/sessions`,
```
hjjackyang, line 262: Nit: I wonder if this can be supported by the generated Open API client somehow too? (Not something to be addressed in this PR.)

hjjackyang (approved):
Such a nice cleanup!

src/colab/client.ts
kevineger: Since this is a tunneled call, I'd like to leave it on the `ColabClient`. Uses a different root URL and auth scheme (OAuth token vs. RP token).

PR #314, opened 2025-12-11
fix: properly set command icons across uses

kevineger:
https://github.com/googlecolab/colab-vscode/pull/312 introduced a bug where the label, including `$(trash)`, was used as the quick pick title:

image

This PR fixes that bug and restructures how icons are specified and used for commands across the codebase.

### DIFF
--- a/src/colab/commands/utils.ts
+++ b/src/colab/commands/utils.ts
@@ -10,12 +10,12 @@ import { Command } from './constants';
 /**
  * Build a label that's prefixed with the provided icon.
  *
- * This is needed in cases where we want the same styling as a quick pick, but
+ * This is needed in cases where we want the same styling as a Quick Pick, but
  * can only provide a simple label (as a string).
  *
  * @param icon - The VS Code icon to prepend the label with.
  * @param label - The label to follow the icon.
- * @returns A string combining the icon and label in the same way quick picks
+ * @returns A string combining the icon and label in the same way Quick Picks
  * do.
  */
 export function buildIconLabel(command: Command) {
@@ -24,7 +24,7 @@ export function buildIconLabel(command: Command) {
 }
 
 /**
- * Strips the icon from the command's labeled, if it is prefixed with one.
+ * Strips the icon from the command's label, if it is prefixed with one.
  *
  * This is needed in cases where we need to handle commands whose labels were
  * styled with the icon but we need to match against the command's true label


src/colab/commands/utils.ts
```diff
@@ -0,0 +1,55 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import vscode, { ThemeIcon } from 'vscode';
+import { Command } from './constants';
+
+/**
+ * Build a label that's prefixed with the provided icon.
+ *
+ * This is needed in cases where we want the same styling as a quick pick, but
```
amtoney524: super nit: "Quick Pick" throughout to match https://code.visualstudio.com/api/ux-guidelines/quick-picks?

src/colab/commands/utils.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @returns A string combining the icon and label in the same way quick picks
+ * do.
+ */
+export function buildIconLabel(command: Command) {
+  const iconPart = command.icon ? `$(${command.icon})  ` : '';
+  return `${iconPart}${command.label}`;
+}
+
+/**
+ * Strips the icon from the command's labeled, if it is prefixed with one.
```
amtoney524: sp: label? (also thanks for the nice comment!)

src/test/helpers/theme.ts
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * A test double for {@link ThemeIcon}.
+ */
+export class TestThemeIcon implements ThemeIcon {
+  constructor(
+    readonly id: string,
+    readonly color?: ThemeColor,
+  ) {}
+
+  static readonly File = new TestThemeIcon('file');
```
amtoney524, line 18: are these needed?

amtoney524 (approved):
looks great!

src/test/helpers/theme.ts
kevineger: Yeah, without them: `Type 'typeof TestThemeIcon' is missing the following properties from type 'typeof ThemeIcon': File, Folder` in `src/test/helpers/vscode.ts`.

src/colab/commands/utils.ts
kevineger: Thanks! Fixed.

src/colab/commands/utils.ts
kevineger: Done

PR #315, opened 2025-12-11
test: add e2e test cases to verify df display and plot

hjjackyang:
**Why**? This is to add more Kernel related test coverage to increase confidence of our custom `WebSocket` override in https://github.com/googlecolab/colab-vscode/pull/297.

### DIFF
--- a/src/test/extension.e2e.test.ts
+++ b/src/test/extension.e2e.test.ts
@@ -25,7 +25,7 @@ import {
 import { CONFIG } from '../colab-config';
 
 const ELEMENT_WAIT_MS = 10000;
-const CELL_EXECUTION_WAIT_MS = 45000;
+const CELL_EXECUTION_WAIT_MS = 30000;
 
 describe('Colab Extension', function () {
   dotenv.config();


src/test/extension.e2e.test.ts
```diff
@@ -24,7 +25,7 @@ import {
 import { CONFIG } from '../colab-config';
 
 const ELEMENT_WAIT_MS = 10000;
-const CELL_EXECUTION_WAIT_MS = 30000;
+const CELL_EXECUTION_WAIT_MS = 45000;
```
kevineger: Is this needed? Looking at the code we really shouldn't need to bump this. Please revert. If you've found it is needed - I'd rather get to the bottom of why.

hjjackyang: Reverted!

amtoney524 (approved):
thanks for the additional coverage! since this is also used as our probe, we'll want to keep an eye out for any flakiness increase. (in the future, we could consider maintaining a small probe test independent of a larger e2e test suite!)

hjjackyang:
> We can totally improve on this over time, but the added functionality may be better off as a standalone test(s)? @amtoney524 could you take a look and review this PR too? I'll approve since this expands test coverage, but I think we should work towards splitting things up (in this PR or a follow-up).
> 
> If not separating here, maybe the auth + server connection could happen in a `beforeAll` and the `it`-s can verify the disjoint cases?
> 
> Related, one thing I'd like us to do is launch the test in a _test workspace_ (VS Code workspace) where the test file (or better yet, file**s**) is already authored and we just connect to Colab + execute. That would remove the complexity and flake of authoring the cells. Can take a look at the `jupyter-vscode` repo for reference.

Thanks for the feedback! I'm going to merge this PR as is to unblock my `WebSocket` override PR, and we can refactor the e2e test structure in a future PR if needed.

PR #316, opened 2025-12-11
fix: catch access errors during token refresh and remove session

mounikasomisetty:
When refreshing the access token, if certain `GaxiosErrors` occur, the session should be removed from storage correctly whether the AuthProvider is initialized or not. Added tests to verify this behavior.

This specific scenario is hard to reproduce manually so there is no screencast attached.

### DIFF
--- a/src/auth/auth-provider.ts
+++ b/src/auth/auth-provider.ts
@@ -218,7 +218,20 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
     if (scopes && !matchesRequiredScopes(scopes)) {
       return [];
     }
-    await this.refreshSessionIfNeeded();
+    try {
+      await this.refreshSessionIfNeeded();
+    } catch (err: unknown) {
+      const { shouldClearSession, reason } =
+        this.shouldClearSessionOnRefreshError(err);
+      if (shouldClearSession) {
+        log.warn(`${reason}. Clearing session.`, err);
+        if (this.session?.id) {
+          await this.removeSession(this.session.id);
+        }
+        return [];
+      }
+      log.error('Unable to refresh access token', err);
+    }
     if (options.account && this.session?.account != options.account) {
       return [];
     }
@@ -370,19 +383,7 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
     if (expiryDateMs && expiryDateMs > Date.now() + REFRESH_MARGIN_MS) {
       return;
     }
-    try {
-      await this.oAuth2Client.refreshAccessToken();
-    } catch (err: unknown) {
-      const { shouldClearSession, reason } =
-        this.shouldClearSessionOnRefreshError(err);
-      if (shouldClearSession) {
-        log.warn(`${reason}. Clearing session.`, err);
-        await this.removeSession(this.session.id);
-        return;
-      }
-      log.error('Unable to refresh access token', err);
-      throw err;
-    }
+    await this.oAuth2Client.refreshAccessToken();
     const accessToken = this.oAuth2Client.credentials.access_token;
     if (!accessToken) {
       throw new Error('Failed to refresh Google OAuth token.');
--- a/src/auth/auth-provider.unit.test.ts
+++ b/src/auth/auth-provider.unit.test.ts
@@ -566,13 +566,13 @@ describe('GoogleAuthProvider', () => {
         });
       }
 
-      it('re-throws non handled errors when refreshing the access token', async () => {
+      it('ignores unhandled errors when refreshing the access token', async () => {
         refreshAccessTokenStub.throws(new Error('🤮'));
         fakeClock.tick(HOUR_MS * 2);
 
         const sessions = authProvider.getSessions(undefined, {});
 
-        await expect(sessions).to.eventually.be.rejectedWith(/🤮/);
+        await expect(sessions).to.eventually.deep.equal([DEFAULT_AUTH_SESSION]);
       });
     });
   });


src/auth/auth-provider.ts
```diff
@@ -364,15 +371,13 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
     if (expiryDateMs && expiryDateMs > Date.now() + REFRESH_MARGIN_MS) {
       return;
     }
-    await this.oAuth2Client.refreshAccessToken();
-    const accessToken = this.oAuth2Client.credentials.access_token;
-    if (!accessToken) {
-      throw new Error('Failed to refresh Google OAuth token.');
+    const accessToken = await this.refreshSession(this.session.id);
```
kevineger: If this is called and an _invalid grant_ error is thrown, then the `accessToken` will be returned as `''` which would not update the session and therefore not refresh the session.

That's because `this.initialize()` will be called, but will early return due to:

```ts
if (this.isInitialized) {
  return;
}
```

What we want is the re-trigger the auth-flow so when we detect these token issues during refresh, the user is prompted to sign in again.

The solution is likely to refactor the code a bit. Hint: `this.initialize()` should only ever be called from within `initialize` if we're recursing.

If the above wasn't clear, happy to pair on this.

kevineger (changes requested):

src/auth/auth-provider.ts
mounikasomisetty: I thought that's what this code did, as when the refresh session fails, the session is cleared and then `getSessions` returns `[]` which would cause `createSession` to re-trigger the OAuth flow.  Let me know if I'm misunderstanding.

I made some changes to make the code more readable, I think the helper method just made it harder to read. Also removing it helped keep the recursion within the `initialize` method.

src/auth/auth-provider.ts
```diff
@@ -141,23 +141,8 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
...
-
-      if (
-        err instanceof GaxiosError &&
-        err.status === 400 &&
-        err.message.includes('invalid_grant')
-      ) {
-        reason = 'OAuth app access to Colab was revoked';
-        shouldClearSession = true;
-      } else if (err instanceof GaxiosError && err.status === 401) {
-        // This should only ever be the case when developer building from source
```
amtoney524, line 155: nit: worth retaining this comment somewhere?

src/auth/auth-provider.ts
```diff
@@ -364,11 +369,24 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
     if (expiryDateMs && expiryDateMs > Date.now() + REFRESH_MARGIN_MS) {
       return;
     }
-    await this.oAuth2Client.refreshAccessToken();
+    try {
+      await this.oAuth2Client.refreshAccessToken();
+    } catch (err: unknown) {
```
amtoney524: (not blocking) wondering if it's worth refactoring into a helper with an option to reinitialize since the error handling's pretty similar here and in `initialize`?

mounikasomisetty: I moved that out of the helper so that the recursion was more evident for reinitializing. Ideas on what you'd like to see?

amtoney524: got it! I was thinking something like a `refreshAccessToken` method that exercised the `try/catch` and returns whether or not the token's refreshed. in looking at this again, I see that there's also a subtle difference in removing the session (e.g., `this.storage.removeSession` vs. `this.removeSession`) between callers so maybe not worth it!

src/auth/auth-provider.ts
```diff
@@ -364,11 +370,24 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
...
+      const { shouldClearSession, reason } =
+        this.shouldClearSessionOnRefreshError(err);
+      if (shouldClearSession) {
+        log.warn(`${reason}. Clearing session.`, err);
+        await this.removeSession(this.session.id);
+        return;
```
kevineger: Thanks so much for refactoring, the flow is now much more clear.

My only qualm is that `refreshSessionIfNeeded(): Promise<void>` now has _side effects_.

A reader would expect it to either refresh the session if it needs it, or do nothing. But now when it detects it can't refresh it removes the session (the side effect).

Could we instead leave `refreshSessionIfNeeded` to throw the error and have the caller, `async getSessions`, handle the `shouldClearSessionOnRefreshError` logic? Seeing the session removed in that method would be more clear.

Also, right now throwing the error in `refreshSessionIfNeeded` would bubble up and be thrown from `getSessions`. This would be an _unhandled exception_. We can't have the extension throwing unhandled exceptions (they show up in the extension view as _erroring_). If that happens, it's a programming error.

---

TL;DR, I'd recommend the following:

```ts
private async refreshSessionIfNeeded(): Promise<void> {
  if (!this.session) {
    return;
  }
  const expiryDateMs = this.oAuth2Client.credentials.expiry_date;
  if (expiryDateMs && expiryDateMs > Date.now() + REFRESH_MARGIN_MS) {
    return;
  }
  await this.oAuth2Client.refreshAccessToken();
  const accessToken = this.oAuth2Client.credentials.access_token;
  if (!accessToken) {
    throw new Error('Failed to refresh Google OAuth token.');
  }

  this.session = {
    ...this.session,
    accessToken,
  };
}
```

```
async getSessions(
  scopes: readonly string[] | undefined,
  options: AuthenticationProviderSessionOptions,
): Promise<AuthenticationSession[]> {
  this.assertReady();
  if (scopes && !matchesRequiredScopes(scopes)) {
    return [];
  }
  try {
    await this.refreshSessionIfNeeded();
  } catch (err: unknown) {
    const { shouldClearSession, reason } =
      this.shouldClearSessionOnRefreshError(err);
    if (shouldClearSession) {
      log.warn(`${reason}. Clearing session.`, err);
      await this.removeSession(this.session.id);
      return;
    }
    log.error('Unable to refresh access token', err);
  }
  
  if (options.account && this.session?.account != options.account) {
    return [];
  }
  return this.session ? [this.session] : [];
}
```

src/auth/auth-provider.ts
```diff
@@ -411,6 +430,18 @@ function matchesRequiredScopes(scopes: readonly string[]): boolean {
   );
 }
```
kevineger, line 433: Nice extraction for clarity, thanks 😎

kevineger (approved):

src/auth/auth-provider.ts
mounikasomisetty: Done with a few minute changes

PR #320, opened 2025-12-18
feat: implement new `propagateDriveCredentials` call in client.ts

hjjackyang:
* **Why**? This API call will be needed later to propagate auth for Drive mount.
* **Note**: There are still some open questions to be addressed in http://go/colab-vscode-drive-mount. However, this API call will be needed regardless.

### DIFF
--- a/src/colab/client.ts
+++ b/src/colab/client.ts
@@ -279,39 +279,35 @@ export class ColabClient {
   }
 
   /**
-   * Propagates credentials to the backend.
+   * Propagates Drive credentials to the backend.
    *
    * @param endpoint - The assignment endpoint to propagate credentials to.
-   * @param authType - If `dfs_persistent`, backend will record the credentials
-   *   and avoid future requests.
-   * @param fileId - This notebook's file ID.
-   * @param renderDataToken - A URL-safe, tamper-proof token containing a record
-   *   of FE rendering data, in the form of an ARI `AuditRenderData` proto
-   *   message, that can be used to reconstruct the text that was shown to users
-   *   during consent moments.
-   * @param dryRun - If true, check if credentials are already propagated to the
-   *   backend and/or obtain an OAuth redirect URL.
+   * @param params - Parameters for credentials propagation API.
    * @param signal - Optional {@link AbortSignal} to cancel the request.
    * @returns Whether propagation is successful. If not, an OAuth redirect URL
    *   is returned to obtain the credentials.
    */
-  async propagateCredentials(
+  async propagateDriveCredentials(
     endpoint: string,
-    authType: 'dfs_ephemeral' | 'dfs_persistent',
-    fileId: string,
-    renderDataToken: string,
-    dryRun = true,
+    params: {
+      authType: 'dfs_ephemeral';
+      // This notebook's file ID.
+      fileId: string;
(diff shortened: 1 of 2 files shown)


src/colab/client.ts
```diff
@@ -275,6 +278,63 @@ export class ColabClient {
...
+   * @param signal - Optional {@link AbortSignal} to cancel the request.
+   * @returns Whether propagation is successful. If not, an OAuth redirect URL
+   *   is returned to obtain the credentials.
+   */
+  async propagateCredentials(
+    endpoint: string,
+    authType: 'dfs_ephemeral' | 'dfs_persistent',
+    fileId: string,
+    renderDataToken: string,
+    dryRun = true,
```
kevineger: Since there's a lot of fields, and some of them are strings, what about putting these in an `interface` so they're named at call sites?

src/colab/client.ts
```diff
@@ -275,6 +278,63 @@ export class ColabClient {
...
+   *   during consent moments.
+   * @param dryRun - If true, check if credentials are already propagated to the
+   *   backend and/or obtain an OAuth redirect URL.
+   * @param signal - Optional {@link AbortSignal} to cancel the request.
+   * @returns Whether propagation is successful. If not, an OAuth redirect URL
+   *   is returned to obtain the credentials.
+   */
+  async propagateCredentials(
+    endpoint: string,
+    authType: 'dfs_ephemeral' | 'dfs_persistent',
```
kevineger: As discussed offline, we may have issues with `dfs_persistent` since we don't have an easy way to get the needed `renderDataToken`. Totally up to you, but maybe we start with only `dfs_ephemeral` and don't offer persistent?

src/colab/client.ts
```diff
@@ -275,6 +278,63 @@ export class ColabClient {
...
+   *   of FE rendering data, in the form of an ARI `AuditRenderData` proto
+   *   message, that can be used to reconstruct the text that was shown to users
+   *   during consent moments.
+   * @param dryRun - If true, check if credentials are already propagated to the
+   *   backend and/or obtain an OAuth redirect URL.
+   * @param signal - Optional {@link AbortSignal} to cancel the request.
+   * @returns Whether propagation is successful. If not, an OAuth redirect URL
+   *   is returned to obtain the credentials.
+   */
+  async propagateCredentials(
```
kevineger: Nit: include _drive_ in method name.

```suggestion
  async propagateDriveCredentials(
```

src/colab/client.ts
hjjackyang: Done. Moved them into a single `params` interface.

src/colab/client.ts
hjjackyang: Done. I initially named it more generically because the backend API supports more than just Drive credentials. However, since the current use case in VS Code is Drive only, we can name the method more specifically, and we can always rename it in the future if we expand the use case.

src/colab/client.ts
hjjackyang: Done. I removed `dfs_persistent` and also `renderDataToken` parameter.

hjjackyang:
> LG!

Thanks for the quick review, @kevineger!

(Now I can go on my vacation with a peace of mind 👍🏼 )

PR #321, opened 2025-12-18
chore: migrate Jupyter definition to Open API 3

kevineger:
Jupyter servers expose the API with Swagger 2. Version 2 has one notable limitation that's needed for downstream PRs, support for `oneOf`.

The Jupyter Contents API returns "one of" a few types for the model's `contents` property. It's description/comment:

> The content, if requested (otherwise null). Will be an array if type is 'directory'.

With the models/API generated with version 2, our only option is to manually modify the generated code, or write complex post-generation swaps which are brittle or break the type system.

I started with breaking the type system and hated it. So I auto-migrated the API definition to 3 using https://converter.swagger.io/ (official supported tool) and re-generated. Tested E2E with my virtual filesystem (downstream PRs) and it's working E2E as expected.

---

I've updated `src/jupyter/client/README.md` to reflect this.


src/jupyter/client/api.yaml
hjjackyang, line 1: Just for my own understanding, is this yaml file auto-generated? When we add new APIs in the future, do we manually update/maintain this file?

kevineger: Should be all answered in `src/jupyter/client/README.md` but let me know if you have any questions!

PR #323, opened 2025-12-18
feat: converters between Jupyter and VS files

kevineger:
These are used in a downstream PR which adds a Jupyter Server `contents/` file system that can be mounted to VS Code.


amtoney524 (approved):
looks great!

PR #324, opened 2025-12-18
feat: Jupyter contents file system

kevineger:
This implements a Colab virtual file-system over the Jupyter contents API, across Colab servers. With this PR, nothing is wired up to call `mount`. That will come in downstream PRs which expose the command entrypoints.

This effectively makes `colab://` a file-type and leverages the `authority` as the disambiguating multi-plexing key. This creates the necessary glue between the many endpoints of the Jupyter Contents API and VS Code. The file-scheme is a standard approach for "remote" or VFS-s (e.g. ssh).

https://github.com/user-attachments/assets/9b0543a9-81bb-46e6-a13c-0de30774ceda

---

These change identified a minor fix needed to `TestUri.from`.

### DIFF
--- a/src/jupyter/contents/file-system.ts
+++ b/src/jupyter/contents/file-system.ts
@@ -110,9 +110,7 @@ export class ContentsFileSystemProvider
       log.info(`Server is already mounted: "${server.label}"`, existingFolder);
       return;
     }
-    const lastIdx = this.vs.workspace.workspaceFolders
-      ? this.vs.workspace.workspaceFolders.length
-      : 0;
+    const lastIdx = this.vs.workspace.workspaceFolders?.length ?? 0;
     // TODO: Consider adding a mechanism to listen for the next workspace folder
     // change that corresponds to the addition, and render then dismiss a
     // notification.
@@ -146,7 +144,7 @@ export class ContentsFileSystemProvider
     },
   ): Disposable {
     this.guardDisposed();
-    this.skipVsCodeFiles(uri);
+    this.throwForVsCodeFile(uri);
     return {
       dispose: () => {
         // No-op
@@ -165,10 +163,10 @@ export class ContentsFileSystemProvider
   @traceMethod
   async stat(uri: Uri): Promise<FileStat> {
     this.guardDisposed();
-    this.skipVsCodeFiles(uri);
+    this.throwForVsCodeFile(uri);
     const path = this.uriToPath(uri);
     try {
-      const client = await this.client(uri);
+      const client = await this.getOrCreateClient(uri);
       const content = await client.get({ path, content: 0 });
       return toFileStat(this.vs, content);
     } catch (error) {
@@ -187,10 +185,10 @@ export class ContentsFileSystemProvider
   @traceMethod
   async readDirectory(uri: Uri): Promise<[string, FileType][]> {
     this.guardDisposed();
-    this.skipVsCodeFiles(uri);
+    this.throwForVsCodeFile(uri);
     const path = this.uriToPath(uri);
     try {
-      const client = await this.client(uri);
+      const client = await this.getOrCreateClient(uri);
       const dir = await client.get({
         path,
         type: ContentsGetTypeEnum.Directory,
@@ -222,10 +220,10 @@ export class ContentsFileSystemProvider
   @traceMethod
   async createDirectory(uri: Uri): Promise<void> {
     this.guardDisposed();
-    this.skipVsCodeFiles(uri);
+    this.throwForVsCodeFile(uri);
     const path = this.uriToPath(uri);
     try {
-      const client = await this.client(uri);
+      const client = await this.getOrCreateClient(uri);
       await client.save({
         path,
         model: {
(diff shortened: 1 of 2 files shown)


src/jupyter/contents/file-system.ts
```diff
@@ -0,0 +1,521 @@
...
+      path: '/',
+    });
+    const existingFolder = this.vs.workspace.getWorkspaceFolder(uri);
+    if (existingFolder) {
+      log.info(`Server is already mounted: "${server.label}"`, existingFolder);
+      return;
+    }
+    const lastIdx = this.vs.workspace.workspaceFolders
+      ? this.vs.workspace.workspaceFolders.length
+      : 0;
```
hjjackyang: This can be simplified as `this.vs.workspace.workspaceFolders?.length ?? 0`.

src/jupyter/contents/file-system.ts
```diff
@@ -0,0 +1,521 @@
...
+      readonly overwrite: boolean;
+    },
+  ): Promise<void> {
+    this.guardDisposed();
+    this.skipVsCodeFiles(uri);
+    const client = await this.client(uri);
+    const path = this.uriToPath(uri);
+    try {
+      let exists = false;
+      if (!(options.create && options.overwrite)) {
```
hjjackyang: Nit: Do we need this outer `if` check?

Let's say if user sets both `create: true` and `overwrite: true`, we will skip the `fileExists` check. In this case, if the file already exists, we will still emit `this.vs.FileChangeType.Created` event type regardless. Not sure how critical for the event type to be accurate in this case.

src/jupyter/contents/file-system.ts
```diff
@@ -0,0 +1,521 @@
...
+      if (newUriExists) {
+        throw this.vs.FileSystemError.FileExists(newUri);
+      }
+    }
+
+    try {
+      await client.rename({ path: oldPath, rename: { path: newPath } });
+      this.changeEmitter.fire([
+        { type: this.vs.FileChangeType.Deleted, uri: oldUri },
+        { type: this.vs.FileChangeType.Created, uri: newUri },
```
hjjackyang: Nit: Similar to the other comment, should we conditionally set this event type to `Changed` vs `Created` depending on the existence of `newUri` file?

src/jupyter/contents/file-system.ts
```diff
@@ -0,0 +1,521 @@
...
+
+  private guardDisposed() {
+    if (this.isDisposed) {
+      throw new Error(
+        'Cannot use ContentsFileSystemProvider after it has been disposed',
+      );
+    }
+  }
+
+  private async client(endpoint: string | Uri): Promise<ContentsApi> {
```
hjjackyang: Nit: `getClient` or `getOrCreateClient` so method name is prefixed with a verb?

src/jupyter/contents/file-system.ts
```diff
@@ -0,0 +1,521 @@
...
+  private async fileExists(
+    client: ContentsApi,
+    path: string,
+  ): Promise<boolean> {
+    try {
+      await client.get({ path, content: 0 });
+      return true;
+    } catch (error) {
+      // Throw errors that aren't a 404.
+      if (!(error instanceof ResponseError) || error.response.status !== 404) {
```
hjjackyang: Prefer to invert this condition with positive boolean to improve readability.

```ts
if (error instanceof ResponseError && error.response.status === 404) {
  return false;
}
throw error;
```

http://go/tott/764

src/jupyter/contents/file-system.ts
```diff
@@ -0,0 +1,521 @@
...
+    }
+    throw error;
+  }
+
+  /**
+   * Throw a {@link FileSystemError.FileNotFound} error for VS Code files we
+   * know the server doesn't have, but VS Code looks for. Avoids the unnecessary
+   * round-trip.
+   */
+  private skipVsCodeFiles(uri: Uri) {
```
hjjackyang: Nit: Given this method throws an error instead of just gracefully "skip", should we rename it to `throwIfVsCodeFile` to be more explicit?

hjjackyang (approved):
LGTM overall with a few minor / readability comments!

src/jupyter/contents/file-system.ts
kevineger: Good point, done!

src/jupyter/contents/file-system.ts
kevineger: Good point, done.

src/jupyter/contents/file-system.ts
kevineger: Done.

src/jupyter/contents/file-system.ts
kevineger: Done.

src/jupyter/contents/file-system.ts
kevineger: Great point, fixed. Both work, but `Changed` is more accurate.

src/jupyter/contents/file-system.ts
kevineger: Nice eye, while not critical (the tree would full-update the file with Created) that would be better.

kevineger (commented):
Thanks for the review!

PR #332, opened 2025-12-22
feat: mount server quick pick

kevineger:
This still is not wired up (called anywhere), but when called will mount the user-selected server. When there's no server, it does nothing. When there's one server, it mounts it directly. When there's multiple servers, it shows a quick pick.

### DIFF
--- a/src/colab/commands/constants.ts
+++ b/src/colab/commands/constants.ts
@@ -60,6 +60,14 @@ export const OPEN_COLAB_WEB: Command = {
   description: 'Open Colab web.',
 };
 
+/** Command to mount a server's file-system. */
+export const MOUNT_SERVER: RegisteredCommand = {
+  id: 'colab.mountServer',
+  label: 'Mount Server to Workspace',
+  icon: 'remote',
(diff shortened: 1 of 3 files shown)


src/colab/commands/constants.ts
```diff
@@ -60,6 +60,14 @@ export const OPEN_COLAB_WEB: Command = {
   description: 'Open Colab web.',
 };
 
+/** Command to mount a server's file-system. */
+export const MOUNT_SERVER: RegisteredCommand = {
+  id: 'colab.mountServer',
+  label: 'Mount Server to Workspace',
+  icon: 'remote',
+  description: 'Reloads VS Code if a Workspace is not already open.',
```
amtoney524, line 68: (just curious) where's the reload triggered?

kevineger: When calling `updateWorkspaceFolders` (VS Code `workspace` API). Details on exactly when a reload will be triggered are [here](https://code.visualstudio.com/api/references/vscode-api#workspace) in the interface docs.

image

PR #334, opened 2025-12-23
feat: support uploading files and folders to Colab

kevineger:
_A (moving) picture's worth a thousand words_...

https://github.com/user-attachments/assets/faea26f3-9a6a-4c2f-9f77-aff949de5da5

As seen in the demo, this enables users to upload a single file, multiple files, directories and nested directories/files to Colab servers.

Why `5b_importexport@30` for the `group`?

When connected remotely, this places _Upload to Colab_ in the same group, just after [Download...](https://github.com/microsoft/vscode/blob/c561232a0a8f96c73b7bd9dc2039b00804284a44/src/vs/workbench/contrib/files/browser/fileActions.contribution.ts#L568C10-L568C25) and [Upload...](https://github.com/microsoft/vscode/blob/c561232a0a8f96c73b7bd9dc2039b00804284a44/src/vs/workbench/contrib/files/browser/fileActions.contribution.ts#L585).

image

### DIFF
--- a/src/colab/commands/files.ts
+++ b/src/colab/commands/files.ts
@@ -24,8 +24,9 @@ import { UPLOAD } from './constants';
  * assignments.
  * @param uri - The primary URI of the file or folder to upload, e.g. the URI of
  * the file which was right-clicked in the file explorer.
- * @param uris - An optional array of URIs to upload, e.g. the other files or
- * folders which were multi-selected.
+ * @param uris - An optional array of all files and folders which should be
+ * uploaded. When provided, the primary URI, provided with `uri`, is assumed to
+ * be in this list.
  * @returns A promise that resolves when the upload process is complete.
  */
 export async function upload(
--- a/src/colab/commands/files.unit.test.ts
+++ b/src/colab/commands/files.unit.test.ts
@@ -101,7 +101,12 @@ describe('File Commands', () => {
     });
 
     it('prompts user to select a server when multiple are assigned', async () => {
-      const otherServer = { ...DEFAULT_SERVER, id: randomUUID(), label: 'bar' };
+      const otherServer = {
+        ...DEFAULT_SERVER,
+        id: randomUUID(),
+        endpoint: 'm-s-bar',
+        label: 'bar',
+      };
       (assignmentManagerStub.getServers as sinon.SinonStub)
         .withArgs('extension')
         .resolves([DEFAULT_SERVER, otherServer]);
@@ -116,7 +121,7 @@ describe('File Commands', () => {
 
       sinon.assert.calledWith(
         vsCodeStub.workspace.fs.writeFile,
-        TestUri.parse('colab://m-s-foo/my-file.txt'),
+        TestUri.parse('colab://m-s-bar/my-file.txt'),
         fileContent,
       );
       sinon.assert.calledWith(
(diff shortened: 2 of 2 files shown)


MeetInCode:
i created below code snippet for temporary use 

```python
# Import deps
import ipywidgets as widgets
from IPython.display import display
import os

# Upload widget
train_uploader = widgets.FileUpload(accept='.xlsx', multiple=True)

# Create an output widget to display status
output = widgets.Output()

def save_files(change):
    """Callback function to save uploaded files"""
    with output:
        output.clear_output()
        if train_uploader.value:
            for filename, file_info in train_uploader.value.items():
                with open(filename, "wb") as f:
                    f.write(file_info['content'])
                print(f"Saved: {filename}")
        else:
            print("No files uploaded yet. Please upload files first.")

# Button to trigger save
save_button = widgets.Button(
    description='Save Uploaded Files',
    button_style='success',
    tooltip='Click after uploading files to save them',
    icon='save'
)
save_button.on_click(lambda _: save_files(None))

# Observe the upload widget for changes
train_uploader.observe(save_files, names='value')

# Display widgets
display(train_uploader)
display(save_button)
display(output)

print("Upload files above. They will be automatically saved when uploaded.")
```

src/colab/commands/files.ts
```diff
@@ -0,0 +1,192 @@
...
+ * files to be uploaded for progress tracking, and performs the upload
+ * recursively. It displays a progress notification during the operation and
+ * emits a final summary message upon completion.
+ *
+ * @param vs - The VS Code module.
+ * @param assignmentManager - The manager responsible for handling server
+ * assignments.
+ * @param uri - The primary URI of the file or folder to upload, e.g. the URI of
+ * the file which was right-clicked in the file explorer.
+ * @param uris - An optional array of URIs to upload, e.g. the other files or
```
amtoney524: if present, is it assumed that this will also contain `uri`? (looking at logic on L42)

src/colab/commands/files.unit.test.ts
```diff
@@ -0,0 +1,344 @@
...
+        value: otherServer,
+      } as QuickPickItem);
+      const fileContent = new Uint8Array([4, 5, 6]);
+      vsCodeStub.workspace.fs.readFile.resolves(fileContent);
+
+      await upload(vsCodeStub.asVsCode(), assignmentManagerStub, fileUri);
+
+      sinon.assert.calledWith(
+        vsCodeStub.workspace.fs.writeFile,
+        TestUri.parse('colab://m-s-foo/my-file.txt'),
```
amtoney524: nit: would it be worth differentiating this value for other server?

src/colab/commands/files.unit.test.ts
```diff
@@ -0,0 +1,344 @@
...
+        vsCodeStub.window.showInformationMessage,
+        'Successfully uploaded 1 item to foo',
+      );
+      sinon.assert.calledWith(
+        vsCodeStub.window.showErrorMessage,
+        'Failed to upload 1 item. Check logs for details.',
+      );
+    });
+
+    it('reports total failure', async () => {
```
amtoney524, line 278: nice handling of error cases! just curious, is the user able to attempt to upload an empty dir? what happens then?

kevineger: We allow it! A dir is _technically_ a file in Linux / Unix. "Everything is a file".

https://en.wikipedia.org/wiki/Everything_is_a_file

src/colab/commands/files.unit.test.ts
kevineger: Yes it absolutely would, fantastic catch. Fixed.

src/colab/commands/files.ts
kevineger: It is, yes. I'll make that explicit in the docs.

PR #340, opened 2025-12-26
feat: Auto-configure Plotly renderer for Colab compatibility

### LINKED ISSUE #339: Interactive Plotly Figures Fail to Render with ipywidget Error
## Description
When executing cells that generate interactive Plotly visualizations in a VS Code notebook connected to a Colab runtime, the figures fail to render and display a JavaScript error instead. This completely breaks interactive exploratory data analysis workflows that depend on Plotly's hover, zoom, and interactive features.

**Error Message:**
```
Error rendering output item using 'jupyter-ipywidget-renderer': i is not a function
```

## Environment
- **VS Code version:** 1.96.0
- **Colab extension version:** 0.0.16 (latest as of Dec 2025)
- **Jupyter extension version:** 2024.11.0
- **OS:** Windows 11
- **Python version:** 3.10.12 (Colab default)
- **Plotly version:** 5.24.1

## Reproduction Steps

1. Connect VS Code to a Colab runtime (CPU or GPU)
2. Create a new notebook cell with this minimal example:
```python
import plotly.graph_objects as go

fig = go.Figure(data=go.Scatter(x=[1, 2, 3], y=[4, 5, 6]))
fig.show()
```
3. Execute the cell
4. Observe the error instead of the interactive plot

## Expected Behavior
The Plotly figure should render as an interactive visualization with hover tooltips, zoom controls, and pan functionality, just like it does in:
- Native Google Colab web interface
- Local VS Code with local Jupyter kernels
- JupyterLab

## Actual Behavior
The cell output shows:
```
Error rendering output item using 'jupyter-ipywidget-renderer': i is not a function
```

The figure does not render at all.

## Workarounds Attempted (None Successful)
I've tried multiple approaches to fix this, all unsuccessful:

**1. Changing Plotly renderers:**
```python
import plotly.io as pio
pio.renderers.default = 'colab'  # No effect
pio.renderers.default = 'iframe'  # No effect
pio.renderers.default = 'plotly_mimetype'  # ✓ WORKS!
```

**2. Installing ipywidgets explicitly:**
```python
!pip install ipywidgets==8.1.5
```

**3. Using static renderers (works but defeats purpose):**
```python
pio.renderers.default = 'png'  # Works but produces static images only
```

**4. Saving to HTML and displaying:**
```python
fig.write_html('temp.html')
from IPython.display import IFrame
IFrame('temp.html', width=800, height=600)
```
This displays the chart but requires manual file downloads and breaks the notebook workflow.

## Impact
This creates **unnecessary friction** for data science and ML workflows because:
- Users expect Plotly to "just work" like it does in native Colab
- The default renderer fails with a cryptic error message
- Users must discover the non-obvious `plotly_mimetype` workaround
- Every notebook needs manual renderer configuration at the top
- Breaks compatibility with notebooks written for native Colab (they use default renderer)

**User Experience Issue:**
When migrating from native Colab to VS Code + Colab extension, users hit this error immediately and may assume Plotly doesn't work at all, rather than discovering the workaround.

## Additional Context
- This works perfectly in the native Colab web interface
- This works with local VS Code + local Jupyter kernels
- Other visualizations (matplotlib, seaborn) work fine with the Colab extension
- The issue appears to be specific to widget-based interactive outputs (Plotly, ipywidgets)

## Suggested Fix

Since `plotly_mimetype` works, the extension should either:

**Option 1 (Preferred):** Auto-configure the correct renderer
- Detect when connected to Colab runtime
- Automatically set `plotly.io.renderers.default = 'plotly_mimetype'`
- Make Plotly work "out of the box" like in native Colab

**Option 2:** Fix the default renderer
- Investigate why the default ipywidget renderer fails
- Ensure compatibility with Colab's widget implementation
- This would benefit other interactive libraries too (ipywidgets, altair, etc.)

**Option 3 (Minimum):** Documentation
- Add a troubleshooting section to the README
- Document the `plotly_mimetype` workaround
- Include example setup cell for notebooks

## Related Issues
- #215 mentions `google.colab` library incompatibilities (userdata, drive.mount)
- Similar widget rendering issues might affect other interactive libraries (ipywidgets, altair, bokeh)

ashioyajotham:
Fixes #339: Interactive Plotly figures fail to render with ipywidget error.

Automatically sets the Plotly renderer to 'plotly_mimetype' on first code execution per kernel session, enabling interactive Plotly charts to render correctly in VS Code when connected to a Colab runtime.

Implementation:
- Created plotly-config.ts module to inject Plotly config code
- Modified colab-proxy-web-socket.ts to intercept execute_request messages
- Uses try/except to gracefully handle cases where Plotly is not installed
- Only configures on first execute_request per session (idempotent)
- Follows existing pattern from drive-mount-warning.ts

### DIFF
--- a/project-words.txt
+++ b/project-words.txt
@@ -21,9 +21,12 @@ milli
 mocharc
 openapitools
 pkce
+plotly
+Plotly
 Pyright
 quickpick
 refreshable
+sess
 toggleable
 tombstoned
 toolsai
--- a/src/jupyter/colab-proxy-web-socket.ts
+++ b/src/jupyter/colab-proxy-web-socket.ts
@@ -11,11 +11,12 @@ import {
   COLAB_RUNTIME_PROXY_TOKEN_HEADER,
 } from '../colab/headers';
 import { warnOnDriveMount } from './drive-mount-warning';
+import { injectPlotlyConfig } from './plotly-config';
 
 /**
  * Returns a class which extends {@link WebSocket}, adds Colab's custom headers,
- * and intercepts {@link WebSocket.send} to warn users when on `drive.mount`
- * execution.
+ * intercepts {@link WebSocket.send} to warn users when on `drive.mount`
+ * execution, and auto-configures Plotly renderer for Colab compatibility.
  */
 export function colabProxyWebSocket(
   vs: typeof vscode,
@@ -67,11 +68,14 @@ export function colabProxyWebSocket(
     ) {
       warnOnDriveMount(vs, data);
 
+      // Auto-configure Plotly renderer for Colab compatibility
+      const modifiedData = injectPlotlyConfig(data);
+
       if (options === undefined || typeof options === 'function') {
         cb = options;
         options = {};
       }
-      super.send(data, options, cb);
+      super.send(modifiedData, options, cb);
     }
   };
 }
--- a/src/jupyter/colab-proxy-web-socket.unit.test.ts
+++ b/src/jupyter/colab-proxy-web-socket.unit.test.ts
@@ -9,6 +9,7 @@ import { expect } from 'chai';
 import WebSocket from 'ws';
 import { newVsCodeStub, VsCodeStub } from '../test/helpers/vscode';
 import { colabProxyWebSocket } from './colab-proxy-web-socket';
+import { resetConfiguredSessions } from './plotly-config';
 
 describe('colabProxyWebSocket', () => {
   const testToken = 'test-token';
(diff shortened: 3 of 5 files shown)


kevineger:
Just as a paper trail, ongoing discussion on the ideal approach in the linked issue https://github.com/googlecolab/colab-vscode/issues/339. But thanks for the initiative here!

kevineger:
Closing per discussion and proposed solution (PR sent to Plotly) as discussed in https://github.com/googlecolab/colab-vscode/issues/339.

ashioyajotham:
Thanks, Kevin, will follow up on it!

Issue #328, opened 2025-12-21
Binary data (ArrayBuffer) not handled in v0.1.4


Hansehart:
## Description

ipywidgets.FileUpload widget fails when uploading files. The Colab extension attempts to parse an ArrayBuffer as JSON, causing a SyntaxError. This was working in v0.1.3.

## Environment

VS Code version: 1.107.0
Colab extension version: 0.1.4
Jupyter extension version: 2025.9.1
OS (Linux | Mac | Windows): Windows (WSL2 with Ubuntu dev container)

## Repro Steps

```python
# Select local training file
import ipywidgets as widgets
from IPython.display import display, FileLink

train_uploader = widgets.FileUpload(accept='.csv', multiple=True)
display(train_uploader)

# Write from local memory into remote Colab environment
for filename, file_info in train_uploader.value.items():
    with open(filename, "wb") as f:
        f.write(file_info['content'])
    print(f"Saved: {filename}")
```

## Expected Behaviour
File uploads successfully and train_uploader.value contains the uploaded file data, allowing it to be saved to the Colab filesystem.

## Actual Behaviour
Upload fails silently. The Output panel shows a SyntaxError where JSON.parse() is called on an ArrayBuffer object ("[object ArrayBuffer]" is not valid JSON).

## Additional context

Downgrading to v0.1.3 solves the issue — this is a regression in v0.1.4
The error originates in the Colab extension's G7 function which doesn't handle binary ArrayBuffer data before parsing as JSON
A secondary error in Jupyter (cellIndex undefined) also appears, likely caused by the Colab error leaving the widget in a bad state

## Logs

Logs from `Jupyter` in the `Output` panel (`View` > `Output`, change the drop-down the upper-right of the `Output` panel to `Jupyter`).

```txt
2025-12-21 14:14:47.934 [error] TypeError: Cannot read properties of undefined (reading 'cellIndex')
    at /home/codespace/.vscode-server/extensions/ms-toolsai.jupyter-2025.9.1-linux-x64/dist/extension.node.js:309:28637

2025-12-21 14:14:52.046 [error] SyntaxError: Unexpected token 'o', "[object ArrayBuffer]" is not valid JSON
    at JSON.parse (<anonymous>)
    at G7 (/home/codespace/.vscode-server/extensions/google.colab-0.1.4/out/extension.js:72:36473)
    at t.send (/home/codespace/.vscode-server/extensions/google.colab-0.1.4/out/extension.js:72:37506)
[...]

PR #348, opened 2026-01-05
fix: ArrayBuffer handling in Colab proxy WebSocket

hjjackyang:
This PR fixed the ArrayBuffer handling (raised in https://github.com/googlecolab/colab-vscode/issues/328) in the following way:
* Aligning the `override send` signature with `ws` and taking in `data` parameter as `BufferLike` type.
* Only triggering `warnOnDriveMount` if `typeof data === 'string'`.

As enhancements / refactoring, I also:
1) Consolidated `drive-mount-warning.ts` into `colab-proxy-web-socket.ts` to improve testability of the new behavior(s).
2) Additionally handled string data that is a malformed JSON message.

Demo of the fixed proxy WebSocket with `drive.mount` warning and upload widget:

[Drive mount warning after fixing ArrayBuffer handling.webm](https://github.com/user-attachments/assets/63567105-84a2-4443-b39a-c9a5717ca6fb)

### DIFF
--- a/src/jupyter/colab-proxy-web-socket.unit.test.ts
+++ b/src/jupyter/colab-proxy-web-socket.unit.test.ts
@@ -148,9 +148,7 @@ describe('colabProxyWebSocket', () => {
       testWebSocket.send(rawJupyterMessage, {});
       await flush();
 
-      sinon.assert.notCalled(
-        vsCodeStub.window.showWarningMessage as sinon.SinonStub,
-      );
+      sinon.assert.notCalled(vsCodeStub.window.showWarningMessage);
     });
 
     it('does not show warning notification if not executing drive.mount()', async () => {
@@ -168,9 +166,7 @@ describe('colabProxyWebSocket', () => {
       testWebSocket.send(rawJupyterMessage, {});
       await flush();
 
-      sinon.assert.notCalled(
-        vsCodeStub.window.showWarningMessage as sinon.SinonStub,
-      );
+      sinon.assert.notCalled(vsCodeStub.window.showWarningMessage);
     });
 
     it('does not show warning notification if message is empty', async () => {
@@ -184,9 +180,7 @@ describe('colabProxyWebSocket', () => {
       testWebSocket.send('', {});
       await flush();
 
-      sinon.assert.notCalled(
-        vsCodeStub.window.showWarningMessage as sinon.SinonStub,
-      );
+      sinon.assert.notCalled(vsCodeStub.window.showWarningMessage);
     });
 
     it('does not show warning notification if message is not Jupyter message format', async () => {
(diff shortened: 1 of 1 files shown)


src/jupyter/colab-proxy-web-socket.unit.test.ts
```diff
@@ -18,27 +20,225 @@ describe('colabProxyWebSocket', () => {
...
+      },
+      { name: 'options only', protocols: {}, options: undefined },
+      { name: 'single protocol only', protocols: '', options: undefined },
+      { name: 'protocols only', protocols: [], options: undefined },
+      { name: 'single protocol and options', protocols: '', options: {} },
+      { name: 'protocols and options', protocols: [], options: {} },
+    ];
+
+    tests.forEach(({ name, protocols, options }) => {
+      it(`adds Colab headers to WebSocket with ${name}`, () => {
```
kevineger, line 38: Not new in this PR, but this test doesn't actually validate the headers are added? It's just guarding that construction doesn't throw.

src/jupyter/colab-proxy-web-socket.unit.test.ts
```diff
@@ -18,27 +20,225 @@ describe('colabProxyWebSocket', () => {
...
+        testToken,
+        TestWebSocket,
+      );
+      const testWebSocket = new wsc('ws://example.com/socket');
+
+      testWebSocket.send(rawJupyterMessage, {});
+      await flush();
+
+      sinon.assert.notCalled(
+        vsCodeStub.window.showWarningMessage as sinon.SinonStub,
```
kevineger: Nit: cast is unecessary. It's only needed when calling `callsFake` due to method overloading. Remove throughout.

src/jupyter/drive-mount-warning.ts
```diff
@@ -1,77 +0,0 @@
-/**
```
kevineger, line 1: Mind providing a bit more detail on why you don't want to keep the separation? Personally I preferred them separate as you originally had.

kevineger (approved):
Thanks for circling back and fixing this 🙂

src/jupyter/colab-proxy-web-socket.unit.test.ts
hjjackyang: This is a bit hacky, but the headers are validated in `TestWebSocket` constructor.

src/jupyter/colab-proxy-web-socket.unit.test.ts
hjjackyang: Done!

src/jupyter/drive-mount-warning.ts
hjjackyang: Sure, basically I had a hard time stubbing out `warnOnDriveMount` function in `colab-proxy-web-socket.unit.test.ts`, and I don't want to assert if vscode notification was shown or not in `colab-proxy-web-socket.unit.test.ts` because that was an implementation detail in `drive-mount-warning.ts`. So I ended up deciding to consolidate 2 files, so I can directly assert whether vscode notification is shown for the ArrayBuffer test case.

PR #349, opened 2026-01-05
chore: fix license header check

kevineger:
The [company guidance](http://go/copyright#the-year) is as follows:

- For existing files (initially published in 2025) that you modify:
  - You should retain the earliest copyright year (2025).
  - The guidelines state: "Use the year the file was first published... don't change this if you edit, move or copy the file."
  - While you can optionally add the current year (e.g., Copyright 2025-2026 Google LLC), this is explicitly stated as "not necessary nor encouraged."
  - Therefore, if a file was first published in 2025, its copyright notice should remain Copyright 2025 Google LLC, even if you modify it in 2026.
- For new files:
  - For new files, the copyright year should be the year the file is first published externally.
  - If you add a new file and publish it to GitHub in 2026, its copyright notice should be Copyright 2026 Google LLC.

In summary, the key is to use the single year of first external publication for each file and generally avoid updating it or using year ranges unless there's a specific, encouraged reason to do so, which is not the case for simple modifications.

Drew influence from https://github.com/google-gemini/gemini-cli/pull/15780.


amtoney524 (approved):
nice!

PR #362, opened 2026-01-17
feat: Colab server tree data provider

kevineger:
Adds a `TreeDataProvider` for Colab servers, to effectively "mount" the `/content` directory as a tree-view.

### DIFF
--- a/src/colab/server-browser/server-tree.ts
+++ b/src/colab/server-browser/server-tree.ts
@@ -151,17 +151,19 @@ export class ServerTreeProvider
 
       // Sort: Directories first, then alphabetical by name.
       entries.sort((a, b) => {
-        if (a[1] !== b[1]) {
-          return b[1] === FileType.Directory ? 1 : -1;
+        const [aName, aType] = a;
+        const [bName, bType] = b;
+        if (aType !== bType) {
+          return bType === FileType.Directory ? 1 : -1;
         }
-        return a[0].localeCompare(b[0]);
+        return aName.localeCompare(bName);
       });
 
       return entries.map(([name, type]) => {
         const itemUri = Uri.joinPath(uri, name);
         const uriString = itemUri.toString();
         const existing = this.serverItemsByUri.get(uriString);
-        if (existing && existing.type === type) {
+        if (existing?.type === type) {
           return existing;
         }
 
@@ -186,12 +188,11 @@ export class ServerTreeProvider
   }
 
   private guardDisposed() {
-    if (!this.isDisposed) {
-      return;
+    if (this.isDisposed) {
+      throw new Error(
+        'ServerTreeProvider cannot be used after it has been disposed.',
+      );
     }
-    throw new Error(
-      'ServerTreeProvider cannot be used after it has been disposed.',
-    );
   }
 }
 
--- a/src/colab/server-browser/server-tree.vscode.test.ts
+++ b/src/colab/server-browser/server-tree.vscode.test.ts
@@ -118,18 +118,21 @@ describe('ServerTreeProvider', () => {
         (assignmentStub.getServers as sinon.SinonStub).returns([]);
       });
 
-      it('returns no items with unauthorized', async () => {
-        await expect(tree.getChildren(undefined)).to.eventually.deep.equal([]);
+      const authStates = [AuthState.SIGNED_OUT, AuthState.SIGNED_IN] as const;
 
-        sinon.assert.notCalled(fsStub.readDirectory);
-      });
+      authStates.forEach((authState) => {
+        const state =
+          authState === AuthState.SIGNED_IN ? 'authorized' : 'unauthorized';
 
-      it('returns no items while authorized', async () => {
-        toggleAuth(AuthState.SIGNED_IN);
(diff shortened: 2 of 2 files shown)


src/colab/server-browser/server-tree.ts
```diff
@@ -0,0 +1,203 @@
...
+  private readonly fileListener: Disposable;
+  // VS Code uses referential equality to identify TreeItems, so we need to
+  // cache them to ensure we event with the same instance as returned by
+  // `getChildren`.
+  private serverItemsByUri = new Map<string, ServerItem>();
+  private isAuthorized = false;
+  private isDisposed = false;
+
+  constructor(
+    private readonly assignments: AssignmentManager,
```
hjjackyang, line 51: Nit: Prefer to group params by visibility. For example:
```ts
constructor(
  authChange: Event<AuthChangeEvent>,
  assignmentChange: Event<AssignmentChangeEvent>,
  fileChange: Event<FileChangeEvent[]>,
  private readonly assignments: AssignmentManager,
  private readonly scheme = 'colab',
) {
```

src/colab/server-browser/server-tree.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+  }
+
+  private async getServerItems(uri: Uri): Promise<ServerItem[]> {
+    try {
+      const entries = await workspace.fs.readDirectory(uri);
+
+      // Sort: Directories first, then alphabetical by name.
+      entries.sort((a, b) => {
+        if (a[1] !== b[1]) {
```
hjjackyang: Nit: For better readability, could we assign `a[0]`, `a[1]`, `b[0]`, `b[1]` to local variables, so it's easier for readers to know what they are?

For example:
```ts
const aFileType = a[1];
const bFileType = b[1];
if (aFileType !== bFileType) {
...
```

src/colab/server-browser/server-tree.ts
```diff
@@ -0,0 +1,203 @@
...
+          return b[1] === FileType.Directory ? 1 : -1;
+        }
+        return a[0].localeCompare(b[0]);
+      });
+
+      return entries.map(([name, type]) => {
+        const itemUri = Uri.joinPath(uri, name);
+        const uriString = itemUri.toString();
+        const existing = this.serverItemsByUri.get(uriString);
+        if (existing && existing.type === type) {
```
hjjackyang: Nit: Could this be simplified as:
```ts
if (existing?.type === type) {
```

src/colab/server-browser/server-tree.ts
```diff
@@ -0,0 +1,203 @@
...
+    // Also remove any children that might be in the cache.
+    for (const key of this.serverItemsByUri.keys()) {
+      if (key.startsWith(uriString + '/')) {
+        this.serverItemsByUri.delete(key);
+      }
+    }
+  }
+
+  private guardDisposed() {
+    if (!this.isDisposed) {
```
hjjackyang: Nit: Why not simply:
```ts
if (this.isDisposed) {
  throw new Error(...);
}
```

src/colab/server-browser/server-tree.vscode.test.ts
```diff
@@ -0,0 +1,475 @@
...
+    sinon.restore();
+  });
+
+  describe('getChildren', () => {
+    describe('without servers', () => {
+      beforeEach(() => {
+        (assignmentStub.getServers as sinon.SinonStub).returns([]);
+      });
+
+      it('returns no items with unauthorized', async () => {
```
hjjackyang: Nit: I feel perhaps we could combine these 2 tests into one parameterized test:
```ts
const tests = [
  { name: 'unauthorized', state: AuthState.SIGNED_OUT },
  { name: 'authorized', state: AuthState.SIGNED_IN },
];

tests.forEach(({ name, state }) => {
  it(`returns no items while ${name}`, async () => {
    toggleAuth(state);

    await expect(tree.getChildren(undefined)).to.eventually.deep.equal([]);

    sinon.assert.notCalled(fsStub.readDirectory);
  });
});
```
https://mochajs.org/declaring/dynamic-tests

hjjackyang (approved):
Approved with a few nit comments!

src/colab/server-browser/server-tree.ts
kevineger: Done.

src/colab/server-browser/server-tree.ts
kevineger: Done.

src/colab/server-browser/server-tree.ts
kevineger: Done, with destructuring.

src/colab/server-browser/server-tree.ts
kevineger: I'd rather not apply that for constructor params. A ctor is the public API and I prioritize call site clarity over member impl details.

src/colab/server-browser/server-tree.vscode.test.ts
kevineger: Done.

kevineger (commented):
Thanks for the review!

PR #365, opened 2026-01-18
feat: support `drive.mount` MVP

hjjackyang:
This MVP introduces an experiment `colab.driveMounting` config for users to enable it explicitly. Once enabled, `google.colab.drive.mount` library will work in VS Code extension.

---

**Caveat**: This MVP did not make any backend changes, so the OAuth redirect page:
1. Only shows "Please close this tab / window" without a better VS Code specific UI.
2. Does not loop back to VS Code after completed. Users will have to manually switch back to VS Code and click "Continue".

**Immediate next steps**:
1. Add experiment feature to wiki.
2. Update `drive.mount` warning to point to experiment feature wiki.

---

[Demo screencast](https://screencast.googleplex.com/cast/NTMwODQ4NzEyNTk1ODY1NnwxMDkyODE4Yi1mNw)

---

* Related GitHub Issue: https://github.com/googlecolab/colab-vscode/issues/256
* Internal tracking bug: b/461876404

### DIFF
--- a/src/jupyter/assignments.ts
+++ b/src/jupyter/assignments.ts
@@ -552,7 +552,6 @@ export class AssignmentManager implements vscode.Disposable {
       ...colabServer,
       connectionInformation: {
         ...colabServer.connectionInformation,
-        fetch: colabProxyFetch(token),
         WebSocket: colabProxyWebSocket(this.vs, this.client, colabServer),
       },
     };


src/auth/drive.ts
```diff
@@ -0,0 +1,169 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { v4 as uuid } from 'uuid';
+import vscode from 'vscode';
+import WebSocket from 'ws';
+import { ColabClient } from '../colab/client';
+import { log } from '../common/logging';
+
+export async function handleDriveFsAuth(
```
kevineger, line 23: Nit: documented exported functions. There are a couple of non-obvious params (e.g. _what_ `socket`?) and the flow varies a lot depending on the dry run result, etc.

src/auth/drive.ts
```diff
@@ -0,0 +1,169 @@
...
+  vs: typeof vscode,
+  socket: WebSocket,
+  client: ColabClient,
+  endpoint: string,
+  requestMessageId: number,
+) {
+  const fileId = vs.window.activeNotebookEditor?.notebook.uri.path ?? '';
+  const dryRunResult = await client.propagateDriveCredentials(endpoint, {
+    authType: 'dfs_ephemeral',
+    fileId,
```
kevineger: For Colab web, isn't this the notebook file ID? Those derive notebook hash's. In VS Code, we `randomUUID()` those.

How is the `file_id` form param used in the backend on the `/credentials-propagation` call?

I think it's unlikely we want to use the active NB editor file ID here. Given that you could, it seems like any random string can be used, since these file IDs mean nothing to our backend. A user can have many notebook editors connected to the same, or different, Colab servers. In addition, the _active_ editor may not be the one which is mounting (e.g. you're running a few cells and you switch editor tabs).

src/auth/drive.ts
```diff
@@ -0,0 +1,169 @@
...
+    dryRun: true,
+  });
+
+  if (dryRunResult.success) {
```
kevineger, line 35: Nit: consider lifting the functionality on each of these big `if` blocks to their own functions so the overall flow is more easily understood at a glance.

src/auth/drive.ts
```diff
@@ -0,0 +1,169 @@
...
+      fileId,
+      requestMessageId,
+    );
+    return;
+  }
+
+  if (dryRunResult.unauthorizedRedirectUri) {
+    const yes = 'Connect to Google Drive';
+    const consent = await vs.window.showInformationMessage(
+      'Permit this notebook to access your Google Drive files?',
```
kevineger: It isn't the `notebook` which will now have access, it's any notebook connected to that runtime. E.g. with the current impl you can mount in one notebook, create another, connect it to the runtime and the `!ls /content/drive` successfully.

Should we update to `Permit ${server.label} to access your Google Drive files?`.

(Also can we double check the verbiage we use on web and be consistent?)

src/auth/drive.ts
```diff
@@ -0,0 +1,169 @@
...
+      yes,
+    );
+    if (consent === yes) {
+      await vs.env.openExternal(
+        vs.Uri.parse(dryRunResult.unauthorizedRedirectUri),
+      );
+
+      const ctn = 'Continue';
+      const selection = await vs.window.showInformationMessage(
+        'Please complete the authorization in your browser. If done, click "Continue".',
```
kevineger: Nit: maybe be clear that if they _continue_ before completing, it won't work.

```suggestion
        'Please complete the authorization in your browser. Only once done, click "Continue".',
```

src/auth/drive.ts
hjjackyang: Done. Let me know if you'd like to adjust anything to make it more clear.

src/auth/drive.ts
hjjackyang: Turns out `file_id` is only used if backend needs to record the consent, which is *not* applicable to `dfs_ephemeral` auth:

http://google3/research/colab/datalab/tunnel/managedtunnel.go;l=4060;rcl=858724559

I just removed the `file_id` from our call stacks altogether.

src/auth/drive.ts
hjjackyang: Done. Adjusted all verbiage to refer to "Colab server" instead of "notebook".

> Also can we double check the verbiage we use on web and be consistent?

I actually copied my original verbiage directly from web consent dialog:
http://google3/research/colab/frontend/external/notebook.ts;l=3033-3037;rcl=859153058

src/auth/drive.ts
hjjackyang: Done!

src/auth/drive.ts
hjjackyang: Done!

src/auth/drive.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    if (success) {
+      sendDriveFsAuthReply(socket, requestMessageId);
+    } else {
+      sendDriveFsAuthReply(
+        socket,
+        requestMessageId,
+        /* err= */ 'Credentials propagation unsuccessful',
+      );
+    }
```
kevineger: Consider inverting the condition and leaning on consistent error reporting in the `catch`. As is, we don't log of this one failure. Also makes the code a bit easier to read at a glance.

```suggestion
    if (!success) {
      throw new Error('Credentials propagation unsuccessful');
    }
    sendDriveFsAuthReply(socket, requestMessageId);
```

src/auth/drive.unit.test.ts
```diff
@@ -0,0 +1,305 @@
...
+  beforeEach(() => {
+    vsCodeStub = newVsCodeStub();
+
+    colabClientStub = sinon.createStubInstance(ColabClient);
+    colabClientStub.propagateDriveCredentials
+      .withArgs(testServer.endpoint, {
+        dryRun: false,
+        authType: 'dfs_ephemeral',
+      })
+      .resolves({ success: true, unauthorizedRedirectUri: undefined });
```
kevineger: Nit: Consider moving this to a `beforeEach` in a `describe('with existing authorization'` block.

src/auth/drive.unit.test.ts
```diff
@@ -0,0 +1,305 @@
...
+    });
+
+    it('propagates credentials and sends reply if continued', async () => {
+      (vsCodeStub.window.showInformationMessage as sinon.SinonStub)
+        .withArgs(
+          sinon.match(
+            `Permit "${testServer.label}" to access your Google Drive files`,
+          ),
+        )
+        .resolves('Connect to Google Drive');
```
kevineger: Nit: optionally consider grouping in a `describe` `with Drive connection initialized` and have it:

- setup the `showInformationMessage` to resolve `'Connect to Google Drive'`.
- kick off `handleDriveFsAuth`.

src/auth/drive.ts
```diff
@@ -0,0 +1,192 @@
...
+ * @param requestMessageId - The Colab auth request message ID to correlate the
+ *   reply message to
+ */
+export async function handleDriveFsAuth(
+  vs: typeof vscode,
+  socket: WebSocket,
+  client: ColabClient,
+  server: ColabAssignedServer,
+  requestMessageId: number,
+) {
```
kevineger: Nit: annotate the return type.

src/auth/drive.ts
```diff
@@ -0,0 +1,192 @@
...
+      );
+    } else {
+      sendDriveFsAuthReply(
+        socket,
+        requestMessageId,
+        /* err= */ 'User cancelled Google Drive authorization',
+      );
+    }
+  }
+}
```
kevineger, line 56: If `dryRunResult.success` is false and `dryRunResult.unauthorizedRedirectUri` is falsy, the function terminates without calling `sendDriveFsAuthReply`. This will leave the kernel waiting indefinitely for an `input_reply`.

src/auth/drive.ts
```diff
@@ -0,0 +1,192 @@
...
+  const replyMsgId = uuid();
+  const replyMsgType = 'input_reply';
+  const replyMessage: ColabInputReplyMessage = {
+    msg_id: replyMsgId,
+    msg_type: replyMsgType,
+    header: {
+      msg_id: replyMsgId,
+      msg_type: replyMsgType,
+      username: 'username',
+      session: uuid(),
```
kevineger: I could be mistaken, but I think a session ID should remain stable for the duration of a kernel session to allow the frontend/kernel to group related messages. If that's right, do we have access to it? Can we set it?

src/auth/drive.ts
```diff
@@ -0,0 +1,192 @@
...
+) {
+  const replyMsgId = uuid();
+  const replyMsgType = 'input_reply';
+  const replyMessage: ColabInputReplyMessage = {
+    msg_id: replyMsgId,
+    msg_type: replyMsgType,
+    header: {
+      msg_id: replyMsgId,
+      msg_type: replyMsgType,
+      username: 'username',
```
kevineger: Is this required? Is it the right value? I suspect we don't look at it in the kernel.

src/auth/drive.ts
```diff
@@ -0,0 +1,192 @@
...
+ *   reply message to
+ */
+export async function handleDriveFsAuth(
+  vs: typeof vscode,
+  socket: WebSocket,
+  client: ColabClient,
+  server: ColabAssignedServer,
+  requestMessageId: number,
+) {
+  const dryRunResult = await client.propagateDriveCredentials(server.endpoint, {
```
kevineger, line 29: `propagateCredentialsAndSendReply` has a try/catch which'll ensure a reply is sent (even on error). But here if this call fails, an unhandled rejection will mean no error reply is sent and the kernel hangs, right?

src/auth/drive.unit.test.ts
```diff
@@ -0,0 +1,305 @@
...
+        sinon.match(
+          'Please complete the authorization in your browser. Only once done, click "Continue"',
+        ),
+      );
+    });
+
+    it('sends error reply if not consented', async () => {
+      (vsCodeStub.window.showInformationMessage as sinon.SinonStub)
+        .withArgs(
+          sinon.match('Permit this notebook to access your Google Drive files'),
```
kevineger: Matchers by default (for a `Promise<void>`) will return undefined, which is why this is currently passing. But the string here isn't matching what's actually passed (`Permit "${serverLabel}" to access`...).

src/auth/drive.unit.test.ts
```diff
@@ -0,0 +1,305 @@
...
+
+import sinon, { SinonStubbedInstance } from 'sinon';
+import { Uri } from 'vscode';
+import WebSocket from 'ws';
+import { ColabClient } from '../colab/client';
+import { ColabAssignedServer } from '../jupyter/servers';
+import { newVsCodeStub, VsCodeStub } from '../test/helpers/vscode';
+import { handleDriveFsAuth } from './drive';
+
+describe('handleDriveFsAuth', () => {
```
kevineger, line 15: Related to other comment, consider adding a test case where the dry run returns `success: false` without a redirect URI.

src/auth/drive.unit.test.ts
```diff
@@ -0,0 +1,305 @@
...
+          authType: 'dfs_ephemeral',
+        },
+      );
+      sinon.assert.calledOnceWithMatch(
+        webSocketStub.send,
+        sinon.match(function (data: string) {
+          return (
+            data.includes('input_reply') &&
+            data.includes('colab_reply') &&
+            data.includes(`"colab_msg_id":${String(testRequestMessageId)}`) &&
```
kevineger: Rather than do all these `includes` checks, coulid we parse and verify specific fields? Would be a bit more robust over substring detection.

src/jupyter/colab-proxy-web-socket.ts
```diff
@@ -51,14 +57,45 @@ export function colabProxyWebSocket(
       } else {
         super(address, protocols, addColabHeaders(options));
       }
+
+      this.addListener(
+        'message',
+        (data: WebSocket.RawData, isBinary: boolean) => {
+          const driveMountingEnabled = vs.workspace
+            .getConfiguration('colab')
```
kevineger, line 78: Configuration is read from vs.workspace.getConfiguration on every single message received and every message sent. While VS Code caches this, it's slightly inefficient for high-frequency WebSocket traffic.

Could we instead listen to config changes and set a `isSnoopingOnMessagesNeeded` (joke name used for clarity) bool to determine whether or not we unpack and look at the message. Look at usages of `workspace.onDidChangeConfiguration` throughout the project for reference.

src/jupyter/colab-proxy-web-socket.ts
```diff
@@ -137,6 +192,22 @@ const ExecuteRequestSchema = z.object({
...
+    session: z.string(),
+  }),
+  content: z.object({
+    request: z.object({
+      authType: z.literal('dfs_ephemeral'),
+    }),
+  }),
+  metadata: z.object({
+    colab_request_type: z.literal('request_auth'),
+    colab_msg_id: z.number(),
```
kevineger, line 310: This bit me before, so calling it out as something to verify (I'm not sure if it's actually an issue here). Large IDs in JavaScript can sometimes encounter precision issues (requiring a string / conversion).

src/jupyter/colab-proxy-web-socket.unit.test.ts
```diff
@@ -218,16 +243,215 @@ describe('colabProxyWebSocket', () => {
...
+      const testWebSocket = new wsc('ws://example.com/socket');
+
+      testWebSocket.send(rawDriveMountMessage);
+      await flush();
+
+      sinon.assert.notCalled(vsCodeStub.window.showWarningMessage);
+    });
+  });
+
+  describe('message event', () => {
```
kevineger, line 301: Could we add a test case where JSON.parse fails in the message listener?

adityaroxx:
Just wanted to ask that with this solution are we above to mount the drive in vscode or antigravity?

src/auth/drive.unit.test.ts
hjjackyang: This stub for `dryRun: false` is actually common between `describe('with existing authorization'` and `describe('with no existing authorization'` blocks.

kevineger:
> Just wanted to ask that with this solution are we above to mount the drive in vscode or antigravity?

@adityaroxx are you asking if this'll work in both VS Code and Antigravity? If so, yes.

src/jupyter/colab-proxy-web-socket.ts
hjjackyang: I think this should be okay. This ID currently [starts at 0](http://google3/third_party/py/google/colab/_message.py;l=111;rcl=524010549), and gets [incremented by 1](http://google3/third_party/py/google/colab/_message.py;l=126;rcl=524010549) for every new blocking request in the lifetime of a VM.

Given users rarely mount Drive more than once per VM, I don't see this ID gets bigger than 10 (considering other blocking request use cases as well).

src/jupyter/colab-proxy-web-socket.unit.test.ts
hjjackyang: I already have 2 test cases for this:
* `does not trigger handleDriveFsAuth if message is empty`
* `does not trigger handleDriveFsAuth if message is malformed`

src/auth/drive.ts
hjjackyang: Done!

src/auth/drive.unit.test.ts
hjjackyang: Done!

src/auth/drive.ts
hjjackyang: Done!

src/auth/drive.ts
hjjackyang: Good catch!

Though this is unlikely to happen, added proper handling for this.

> This will leave the kernel waiting indefinitely for an `input_reply`.

FYI, there is a 2m timeout in the kernel, so in the worst scenario, kernel won't be waiting indefinitely ;)

src/auth/drive.ts
hjjackyang: > Is this required?

Tested locally without this `username` field and still worked. Removed to keep it clean.

> Is it the right value?

Definitely not! I just blindly copied the value from the `input_reply` message I captured from Colab web.

src/auth/drive.ts
hjjackyang: Another good catch!

Lifted the try/catch block to the outmost `handleDriveFsAuth` function to ensure all error cases are handled.

src/auth/drive.unit.test.ts
hjjackyang: Omitted all unnecessary `resolves(undefined)` stubs and let them fall back to the default behavior.

src/auth/drive.unit.test.ts
hjjackyang: Done!

src/auth/drive.unit.test.ts
hjjackyang: Done!

src/jupyter/colab-proxy-web-socket.ts
hjjackyang: Done.

src/auth/drive.ts
hjjackyang: I'm not sure how to retrieve the correct session ID, but given any session ID would work, I opted to generate a session ID in `ColabWebSocket` for now and maintain it across the lifetime of the WebSocket.

As part of this, I also refactored the code by moving the `sendInputReply` function from `/auth/drive.ts` to `colab-proxy-web-socket.ts` so it can more easily access the session ID.

We can always revisit this if you know the correct session ID to use here.

src/jupyter/assignments.ts
```diff
@@ -545,10 +545,17 @@ export class AssignmentManager implements vscode.Disposable {
...
         fetch: colabProxyFetch(token),
-        WebSocket: colabProxyWebSocket(this.vs, token),
       },
       dateAssigned,
     };
+    return {
+      ...colabServer,
+      connectionInformation: {
+        ...colabServer.connectionInformation,
+        fetch: colabProxyFetch(token),
```
amtoney524: nit: duped with L547?

src/jupyter/colab-proxy-web-socket.ts
```diff
@@ -51,14 +65,74 @@ export function colabProxyWebSocket(
...
+              message = JSON.parse(data) as unknown;
+            } catch (e: unknown) {
+              log.warn('Failed to parse received Jupyter message to JSON:', e);
+              return;
+            }
+
+            if (isColabAuthEphemeralRequest(message)) {
+              log.trace('Colab request message received:', message);
+              handleDriveFsAuthFn(vs, client, server)
+                .then(() => {
```
amtoney524, line 103: nit / optional: would async-await work here? (to align w/internal style guide)

hjjackyang: Unfortunately it won't. If I change this listener function to use async-await, I'm met with a `"Promise returned in function argument where a void return was expected"` error :(

src/jupyter/assignments.ts
hjjackyang: Great catch! Done.

PR #369, opened 2026-01-22
feat: add new file & folder support

kevineger:
These are the first context menu commands for Colab servers. There are undoubtedly improvements we can make to them (e.g. after create a folder, expand it in the viewer), but this change instead adds the functionality in its MVP state. We can iteratively add improvements and more commands! What's important here is the wiring and happy path(s).

https://github.com/user-attachments/assets/704aaf6d-a49a-45bf-9e88-4cbb1a41168c

Currently, the activity bar and these corresponding commands are restricted to dev builds.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -206,13 +206,16 @@ function registerCommands(
     vscode.commands.registerCommand(COLAB_TOOLBAR.id, async () => {
       await notebookToolbar(vscode, assignmentManager);
     }),
-    vscode.commands.registerCommand('colab.newFile', (node: ServerItem) => {
-      void newFile(vscode, node);
-    }),
+    vscode.commands.registerCommand(
+      'colab.newFile',
+      (contextItem: ServerItem) => {
+        void newFile(vscode, contextItem);
+      },
+    ),
     vscode.commands.registerCommand(
       'colab.newFolder',
-      (originItem: ServerItem) => {
-        void newFolder(vscode, originItem);
+      (contextItem: ServerItem) => {
+        void newFolder(vscode, contextItem);
       },
     ),
   ];


src/extension.ts
```diff
@@ -204,6 +206,15 @@ function registerCommands(
     vscode.commands.registerCommand(COLAB_TOOLBAR.id, async () => {
       await notebookToolbar(vscode, assignmentManager);
     }),
+    vscode.commands.registerCommand('colab.newFile', (node: ServerItem) => {
```
amtoney524: nit: does it make sense to standardize naming of the `ServerItem` arg here and below? (e.g., node and originNode vs. item and originItem)

src/colab/server-browser/commands.ts
```diff
@@ -0,0 +1,103 @@
...
+import type { ServerItem } from './server-item';
+
+/**
+ * Creates a new file on the Colab server.
+ *
+ * Prompts the user for a name, validates it, and creates an empty file.
+ * If the name ends with a forward slash, a directory is created instead.
+ * Automatically opens the new file after creation.
+ */
+export async function newFile(vs: typeof vscode, contextItem: ServerItem) {
```
amtoney524, line 17: I was a bit thrown off that `newFile` could also create a folder, thanks for the helpful comment! What are your thoughts on consolidating `newFile` and `newFolder` into something like `newFileOrFolder`?

src/extension.ts
kevineger: For sure! Thanks

src/colab/server-browser/commands.ts
kevineger: Technically _folders_ are files! Since we need to call different `workspace.fs` commands (`createDirectory` or `writeFile`) we wouldn't be able to detect that the goal is to create a folder if we went with `newFileOrFolder` (without adding a disambiguating param, which is equivalent to separate funcs).

kevineger:
> looks great! just a few questions out of curiosity (and sorry if these are addressed in other PRs)
> 
> * what should happen when the file viewer's open and the user connects a different server to the notebook? or switches to a different notebook that's connected to a different server?
> * what should happen if the user executes the `Mount Server to Workspace` command while the file viewer's open?

Thanks!

- It appears in the activity bar (you get both). Switching between notebook tabs doesn't change your focus.
- It works! They play well together actually (updates in one trigger the event to update the other).

src/colab/server-browser/commands.ts
```diff
@@ -0,0 +1,103 @@
...
+    return undefined;
+  }
+}
+
+function validateName(value: string): string | undefined {
+  const trimmed = value.trim();
+  if (trimmed.length === 0 || trimmed === '/') {
+    return 'A name must be provided';
+  }
+  if (value.includes('\\')) {
```
hjjackyang, line 93: Nit: Do we care about / support name containing spaces?

src/colab/server-browser/commands.unit.test.ts
```diff
@@ -0,0 +1,225 @@
...
+      vsStub.workspace.fs.stat.rejects(TestFileSystemError.FileNotFound());
+
+      await newFile(vs, CONTENT_ROOT);
+
+      const validate =
+        vsStub.window.showInputBox.firstCall.args[0]?.validateInput;
+      assert(validate);
+      expect(await validate('')).to.equal('A name must be provided');
+      expect(await validate('   ')).to.equal('A name must be provided');
+      expect(await validate('/')).to.equal('A name must be provided');
```
hjjackyang, line 128: Nit: Edge case `'/ /'`.

During validation, perhaps we should omit all `/`s and then trim to validate?

src/colab/server-browser/commands.ts
kevineger: Certainly a bit more validation we could/should add. At the end of the day, the actual call to the Jupyter contents API will guard anything invalid. This is just for better client handling / UX. Let's improve it over time once the MVP is in.

src/colab/server-browser/commands.unit.test.ts
kevineger: Same as above.

PR #371, opened 2026-01-22
feat: add rename support

kevineger:
https://github.com/user-attachments/assets/568dd3e4-04ca-446d-875c-2575b21419fa

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -230,12 +230,6 @@ function registerCommands(
       },
     ),
     vscode.commands.registerCommand(
-      'colab.download',
-      (contextItem: ServerItem) => {
-        void download(vscode, contextItem);
-      },
-    ),
-    vscode.commands.registerCommand(
       'colab.renameFile',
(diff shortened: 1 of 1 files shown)


src/colab/server-browser/commands.ts
```diff
@@ -104,6 +104,38 @@ export async function download(vs: typeof vscode, item: ServerItem) {
...
+    },
+  });
+
+  if (!newName || newName === oldName) {
+    return;
+  }
+
+  const newUri = vs.Uri.joinPath(destination, newName);
+  try {
+    await vs.workspace.fs.rename(item.uri, newUri, { overwrite: false });
```
amtoney524: from the demo, it looks like renaming a folder can transition it from an expanded to collapsed state - worth adding a todo to maintain expanded states (if possible)?

kevineger: Sure! One of many TODOs we likely want to cover!

PR #379, opened 2026-01-26
feat: add telemetry client

amtoney524:
Adds a client for recording telemetry.

The client is responsible for sending events to Clearcut in batch.

note: This is currently a no-op, as the client is not referenced in the extension.

### DIFF
--- a/src/telemetry/client.ts
+++ b/src/telemetry/client.ts
@@ -55,13 +55,25 @@ export class ClearcutClient implements Disposable {
   // The time when the next flush request is allowed.
   private nextFlush = new Date();
 
+  private isDisposed = false;
+
   dispose() {
+    if (this.isDisposed) {
+      return;
+    }
+    this.isDisposed = true;
     // Flush any remaining events before disposing.
     this.flush(/* force= */ true);
   }
 
   /** Queues a Colab log event for sending to Clearcut. */
   log(event: ColabLogEvent) {
+    if (this.isDisposed) {
+      throw new Error(
+        'ClearcutClient cannot be used after it has been disposed.',
+      );
+    }
+
     const numPendingEvents = this.pendingEvents.length;
     // In theory, we shouldn't exceed MAX_PENDING_EVENTS, but for posterity, we
     // guard against it here.
--- a/src/telemetry/client.unit.test.ts
+++ b/src/telemetry/client.unit.test.ts
@@ -42,21 +42,33 @@ describe('ClearcutClient', () => {
       sinon.assert.calledOnceWithExactly(fetchStub, logRequest([DEFAULT_LOG]));
     });
 
-    it('throws an error when Clearcut responds with a non-200 status', async () => {
-      fetchStub.resolves(new Response('', { status: 500 }));
-      // Since log is sync (fires and forgets), spy on internal error handling
-      const requestSpy = sinon.spy(
-        client as unknown as { issueRequest: () => Promise<void> },
-        'issueRequest',
-      );
+    describe('throws an error', () => {
+      it('when Clearcut responds with a non-200 status', async () => {
+        fetchStub.resolves(new Response('', { status: 500 }));
+        // Since log is sync (fires and forgets), spy on internal error handling
+        const requestSpy = sinon.spy(
+          client as unknown as { issueRequest: () => Promise<void> },
+          'issueRequest',
+        );
 
-      client.log(DEFAULT_LOG);
+        client.log(DEFAULT_LOG);
 
-      let error: Error | undefined;
-      await requestSpy.firstCall.returnValue.catch((e: unknown) => {
-        error = e as Error;
+        let error: Error | undefined;
+        await requestSpy.firstCall.returnValue.catch((e: unknown) => {
+          error = e as Error;
+        });
+        expect(error?.message).to.include('Failed to issue request');
       });
-      expect(error?.message).to.include('Failed to issue request');
+
+      it('when the client is disposed', () => {
+        client.dispose();
+
+        expect(() => {
+          client.log(DEFAULT_LOG);
+        }).to.throw(
+          'ClearcutClient cannot be used after it has been disposed.',
+        );
+      });
     });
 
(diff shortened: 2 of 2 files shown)


src/telemetry/client.ts
```diff
@@ -0,0 +1,107 @@
...
+    this.flush(/* force= */ true);
+  }
+
+  /** Queues a Colab log event for sending to Clearcut. */
+  log(event: ColabLogEvent) {
+    const numPendingEvents = this.pendingEvents.length;
+    // In theory, we shouldn't exceed MAX_PENDING_EVENTS, but for posterity, we
+    // guard against it here.
+    if (numPendingEvents >= MAX_PENDING_EVENTS) {
+      this.pendingEvents.splice(0, numPendingEvents - MAX_PENDING_EVENTS + 1);
```
mounikasomisetty, line 82: *nit* would it make sense to add a comment here to explain why the events are being removed from the front in case of an "overflow"

src/telemetry/client.ts
```diff
@@ -0,0 +1,107 @@
...
+    const events = this.pendingEvents;
+    this.pendingEvents = [];
+    this.nextFlush = new Date(Date.now() + MIN_WAIT_BETWEEN_FLUSHES_MS);
+    void this.issueRequest(events);
+  }
+
+  /** Sends a log request to Clearcut. */
+  private async issueRequest(events: LogEvent[]) {
+    const logRequest: LogRequest = {
+      log_source: 'COLAB_VSCODE',
```
mounikasomisetty: *nit* should this be declared as a constant somewhere?

src/telemetry/client.ts
```diff
@@ -0,0 +1,107 @@
...
+      log_event: events,
+    };
+    const request = new Request('https://play.googleapis.com/log', {
```
mounikasomisetty: *nit* should this be declared as a constant somewhere?

src/telemetry/client.ts
```diff
@@ -0,0 +1,107 @@
...
+      headers: {
+        [CONTENT_TYPE_JSON_HEADER.key]: CONTENT_TYPE_JSON_HEADER.value,
+      },
+    });
+    const response = await fetch(request);
+    // TODO: Rate-limit based on next_request_wait_millis in response.
+    // TODO: Retry on 401 and 5xx.
+    if (!response.ok) {
+      throw new Error(
+        `Failed to issue request ${request.method} ${request.url}: ${response.statusText}`,
```
mounikasomisetty, line 120: *super nit* should this be "Failed to issue ${request.method} request"

src/telemetry/client.ts
```diff
@@ -0,0 +1,107 @@
...
+interface LogRequest {
+  log_source: 'COLAB_VSCODE';
+  log_event: LogEvent[];
+}
+
+// Maximum number of pending events before flushing. When exceeded, events will
+// be dropped from the front of the queue.
+export const MAX_PENDING_EVENTS = 1000;
+// Minimum wait time between flushes in milliseconds.
+export const MIN_WAIT_BETWEEN_FLUSHES_MS = 10 * 1000;
```
hjjackyang: Nit: Do these need to be exported?

> Generally minimize the exported API surface of modules.

http://go/tsjs-style#export-visibility

If you are exporting for tests, consider either http://go/typescript-testing#export-visibility or just hardcoding them in tests.

src/telemetry/client.ts
```diff
@@ -0,0 +1,107 @@
...
+  // The time when the next flush request is allowed.
+  private nextFlush = new Date();
+
+  dispose() {
+    // Flush any remaining events before disposing.
+    this.flush(/* force= */ true);
+  }
+
+  /** Queues a Colab log event for sending to Clearcut. */
+  log(event: ColabLogEvent) {
```
hjjackyang, line 70: Should we guard this method with a `disposed` state?

Similar to https://github.com/googlecolab/colab-vscode/blob/main/src/jupyter/colab-proxy-web-socket.ts#L133.

src/telemetry/client.unit.test.ts
```diff
@@ -0,0 +1,181 @@
...
+      expect(error?.message).to.include('Failed to issue request');
+    });
+
+    describe('while waiting between flushes', () => {
+      const firstLog = DEFAULT_LOG;
+
+      beforeEach(() => {
+        // Log an event to trigger the first flush.
+        client.log(firstLog);
+        sinon.assert.calledOnceWithExactly(fetchStub, logRequest([firstLog]));
```
hjjackyang, line 80: Nit: The tests in this block may be easier to read if we `fetchStub.resetHistory()` after this line. That way you can simply assert `fetchStub` not called in L86.

src/telemetry/client.ts
amtoney524: done!

src/telemetry/client.ts
amtoney524: done!

src/telemetry/client.ts
amtoney524: good idea, dropped a comment!

src/telemetry/client.ts
amtoney524: leaving this one as is for consistency with the colab client error :) https://github.com/googlecolab/colab-vscode/blob/4cfb5bb0d61a74adfe6fa3b0b535223fae95e256/src/colab/client.ts#L516

src/telemetry/client.ts
amtoney524: good idea, went with the `TEST_ONLY` export!

src/telemetry/client.unit.test.ts
amtoney524: thanks for the suggestion, done!

src/telemetry/client.ts
amtoney524: sgtm, added a check for `isDisposed`

amtoney524:
Thanks for the detailed reviews!!

PR #383, opened 2026-01-28
feat: use telemetry endpoint response to set next flush time

amtoney524:
This change parses the `next_request_wait_millis` field out of the Clearcut response, allowing us to set the `nextFlush`. In practice, this has been ~15 minutes.

In order to parse the response as json, we specify the `format` query parameter on the endpoint. Now that the request has to terminate before `nextFlush` is set, we introduce an `isDoingFlush` property in order to queue events while a flush is in flight.

### DIFF
--- a/src/telemetry/client.unit.test.ts
+++ b/src/telemetry/client.unit.test.ts
@@ -50,33 +50,29 @@ describe('ClearcutClient', () => {
       sinon.assert.calledOnceWithExactly(fetchStub, logRequest([DEFAULT_LOG]));
     });
 
-    describe('throws an error', () => {
-      it('when Clearcut responds with a non-200 status', async () => {
-        fetchStub.resolves(FETCH_RESPONSE_500);
-        // Since log is sync (fires and forgets), spy on internal error handling
-        const requestSpy = sinon.spy(
-          client as unknown as { issueRequest: () => Promise<void> },
-          'issueRequest',
-        );
+    it('throws an error when Clearcut responds with a non-200 status', async () => {
+      fetchStub.resolves(FETCH_RESPONSE_500);
+      // Since log is sync (fires and forgets), spy on internal error handling
+      const requestSpy = sinon.spy(
+        client as unknown as { issueRequest: () => Promise<void> },
+        'issueRequest',
+      );
 
+      client.log(DEFAULT_LOG);
+
+      let error: Error | undefined;
+      await requestSpy.firstCall.returnValue.catch((e: unknown) => {
+        error = e as Error;
+      });
+      expect(error?.message).to.include('Failed to issue request');
+    });
+
+    it('throws an error when the client is disposed', () => {
+      client.dispose();
+
+      expect(() => {
         client.log(DEFAULT_LOG);
-
-        let error: Error | undefined;
-        await requestSpy.firstCall.returnValue.catch((e: unknown) => {
-          error = e as Error;
-        });
-        expect(error?.message).to.include('Failed to issue request');
-      });
-
-      it('when the client is disposed', () => {
-        client.dispose();
-
-        expect(() => {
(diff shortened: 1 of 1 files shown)


src/telemetry/client.unit.test.ts
```diff
@@ -74,40 +82,90 @@ describe('ClearcutClient', () => {
...
       const firstLog = DEFAULT_LOG;
 
-      beforeEach(() => {
-        // Log an event to trigger the first flush.
-        client.log(firstLog);
-        sinon.assert.calledOnceWithExactly(fetchStub, logRequest([firstLog]));
-        fetchStub.resetHistory();
-      });
+      describe('queues events to send in batch', () => {
+        it('when the flush interval has not passed', async () => {
```
hjjackyang: I find it a bit weird to have `describe` block describing an _action_ and `it` block describing a _condition_. It's normally the other way around. Could we instead just have 2 `it` blocks with no extra `describe` wrapper (since there is no shared setup/teardown in this case anyway)?

For example:
```ts
it('queues events to send in batch when the flush interval has not passed', ...

it('queues events to send in batch when a flush is already pending', ...
```

src/telemetry/client.ts
```diff
@@ -113,19 +134,23 @@ export class ClearcutClient implements Disposable {
...
     const response = await fetch(request);
-    // TODO: Rate-limit based on next_request_wait_millis in response.
     // TODO: Retry on 401 and 5xx.
     if (!response.ok) {
       throw new Error(
         `Failed to issue request ${request.method} ${request.url}: ${response.statusText}`,
       );
     }
+    const { next_request_wait_millis: waitMillis } =
+      (await response.json()) as LogResponse;
```
hjjackyang: Could there be an edge case where the response is 200 OK, but the response JSON cannot be parsed to `LogResponse`?

In such case, I think you might get an error when access `next_request_wait_millis` from the response? If so, it might be worth making this handling more robust and also add a test case for it.

hjjackyang (changes requested):
Left 2 comments that are worth addressing before approval!

src/telemetry/client.unit.test.ts
amtoney524: sounds good to me, updated per the above!

src/telemetry/client.ts
amtoney524: good idea, updated the handling and added the respective test cases - ptal!

src/telemetry/client.unit.test.ts
```diff
@@ -44,7 +52,7 @@ describe('ClearcutClient', () => {
 
     describe('throws an error', () => {
       it('when Clearcut responds with a non-200 status', async () => {
```
hjjackyang: Pre-existing: Missed this in the previous review. Similar to my other comment, let's remove the extra `describe` wrapper here and simply have the following 2 tests:

```ts
it('throws an error when Clearcut responds with a non-200 status', ...

it('throws an error when the client is disposed', ...
```

src/telemetry/client.unit.test.ts
```diff
@@ -135,36 +191,130 @@ describe('ClearcutClient', () => {
...
 
       sinon.assert.notCalled(fetchStub);
     });
 
-    it('forces a flush regardless of the flush interval', () => {
-      // Log an event to trigger the first flush.
-      client.log(DEFAULT_LOG);
-      sinon.assert.calledOnceWithExactly(fetchStub, logRequest([DEFAULT_LOG]));
+    describe('forces a flush', () => {
+      it('when the flush interval has not passed', () => {
```
hjjackyang: Same here.

```ts
it('forces a flush when the flush interval has not passed', ...

it('forces a flush when a flush is already pending', ...
```

amtoney524:
Thanks for the detailed review!

PR #385, opened 2026-01-29
refactor: `colabProxyWebSocket` to capture client session ID from Kernel messages

hjjackyang:
**Why**? Based on https://jupyter-client.readthedocs.io/en/latest/messaging.html#message-header:

> A _client session ID_, in message headers from a client, should be unique among all clients connected to a kernel. When a client reconnects to a kernel, it should use the same client session ID in its message headers. When a client restarts, it should generate a new client session ID.

In my previous implementation, I decided to generate a client session ID for my custom `input_reply` message only. This works for the time being, but may potentially be problematic because the server will not be able to properly identify this message with the client session.

In this PR, I refactored the implementation by intercepting the first message sent to kernel and capturing the client session ID from there. For the current use case, this is safe because my custom `input_reply` message is always sent after a successful `execute_request` message sent by Jupyter extension with a proper client session ID.

---

As part of this PR, I also imported `@jupyterlab/services` package to refactor some of the custom interfaces I created in my initial implementation.

---

[Demo screencast](https://screencast.googleplex.com/cast/NjEwMjE0Mjk0MjE4MzQyNHwyNjc1ODExNS1jYg) to show `drive.mount` still working E2E after this refactoring.

### DIFF
--- a/src/jupyter/colab-proxy-web-socket.ts
+++ b/src/jupyter/colab-proxy-web-socket.ts
@@ -194,8 +194,9 @@ export function colabProxyWebSocket(
           msg_id: uuid(),
           msg_type: 'input_reply',
           session: this.clientSessionId,
-          username: 'username',
           date: new Date().toISOString(),
+          // Hardcoded `username` and `version` to align with Colab web
+          username: 'username',
           version: '5.0',
         },
         content: {


src/jupyter/colab-proxy-web-socket.ts
```diff
@@ -147,52 +162,40 @@ export function colabProxyWebSocket(
...
-        msg_id: replyMsgId,
-        msg_type: replyMsgType,
         header: {
-          msg_id: replyMsgId,
-          msg_type: replyMsgType,
-          session: this.sessionId,
+          msg_id: uuid(),
+          msg_type: 'input_reply',
+          session: this.clientSessionId,
+          username: 'username',
```
amtoney524: (just curious) why not leave as undefined or parse from the kernel message?

src/jupyter/colab-proxy-web-socket.ts
```diff
@@ -232,36 +235,35 @@ export function colabProxyWebSocket(
...
-  msg_id: string;
-  msg_type: 'input_reply';
-  header: {
-    msg_id: string;
-    msg_type: 'input_reply';
-    session: string;
-    version: string;
-  };
+export interface ColabInputReplyMessage
+  extends Omit<KernelMessage.IInputReplyMsg, 'content' | 'parent_header'> {
```
amtoney524, line 244: nice!

package.json
```diff
@@ -296,6 +296,7 @@
     "@types/mocha": "^10.0.10",
     "@types/node": "22.x",
     "@types/node-fetch": "^2.6.12",
+    "@types/react": "^19.2.10",
```
amtoney524, line 299: why is this needed?

src/jupyter/colab-proxy-web-socket.ts
hjjackyang: Since I refactored to use `KernelMessage.IInputReplyMsg` now, I can no longer leave `username` as undefined. Potentially I could set it to the username from the request message, but in reality, the request message is also set to this dummy `'username'` value.

See this gPaste I captured from Colab network traffic: https://paste.googleplex.com/4965681508515840

package.json
hjjackyang: This is needed by `@jupyterlab/services` package, I think. Without it, our `npm run typecheck` fails.

src/jupyter/colab-proxy-web-socket.ts
hjjackyang: I added a comment to clarify why it's hardcoded.

Issue #223, opened 2025-11-15
(Now supported!): Support uploading local files to Colab servers.


JAE-HUN-CHO:
## Is your feature request related to a problem?

I'm always frustrated when I need to manually upload local files (like datasets or helper scripts) to Google Drive or copy-paste their content just to use them in a Colab session launched from VSCode. This breaks my development flow and adds unnecessary friction, especially during rapid prototyping or data exploration.

## Describe the solution you'd like to see

I’d love to see a built-in feature in the VSCode Colab extension that lets me:

- Right-click any local file in the Explorer and select “Upload to Colab Session”
- Or trigger it via Command Palette with: `Colab: Upload File to Runtime`
- The extension automatically uploads the file to the active Colab runtime (e.g., into `/content/`) and optionally prints or copies the destination path to clipboard
- No manual steps — seamless integration between local workspace and remote runtime

This would make switching between local editing and cloud execution feel truly integrated.

## Describe alternatives you've considered

- Manually dragging files into Google Drive and mounting it in Colab → too slow and requires browser context-switching
- Using `%%writefile` magic to recreate files by pasting content → error-prone for large/binary files
- Running a local HTTP server and downloading via `!wget` in Colab → overkill and insecure for simple workflows

None of these provide the smooth, one-click experience that a native extension feature could offer.

## Additional context

This feature would be especially valuable for data scientists and ML engineers who frequently iterate between local code and GPU-backed Colab runtimes. Reducing context switches and manual steps would significantly improve productivity.

kevineger:
@JAE-HUN-CHO thanks for filing this, same with @aribornstein in https://github.com/googlecolab/colab-vscode/issues/210, @Fullerite in https://github.com/googlecolab/colab-vscode/discussions/221 and @ClaudioZandonella in https://github.com/googlecolab/colab-vscode/discussions/212. You folks, along with lots of other people (e.g. [here on X](https://x.com/eger_kevin/status/1989039759180832807)), are all asking for the same thing. I had really hoped to get some level of support for this out before launching, but we opted to release without it in hopes that basic connectivity would scratch the itch lots of folks have. But rest assured, **this is something we really want to add support for**.

@JAE-HUN-CHO, your feature request here really cleanly lays out the potential solutions which we were already considering and working towards, thanks! Therefore, I'll be using this issue as the canonical tracking item.

---

[...]

PR #391, opened 2026-02-02
fix: server path for file upload

hjjackyang:
**Why**? As raised in this [issue comment](https://github.com/googlecolab/colab-vscode/issues/223#issuecomment-3833403895), file upload capability was not working. Upon investigation, this was broken by https://github.com/googlecolab/colab-vscode/pull/359, causing file to be uploaded to server root directory instead of `/content` directory.

---

**Screencasts**:
* [Before](https://screencast.googleplex.com/cast/NjM2NjcxMTY2OTMyNTgyNHwwMjhhNGE4NC01NA)
* [After](https://screencast.googleplex.com/cast/NjAyMDQ3ODI4MzAyMjMzNnwzMzU4MzgxZS0wMQ)

---

Internal tracking bug: b/481100090

### DIFF
--- a/src/colab/commands/files.unit.test.ts
+++ b/src/colab/commands/files.unit.test.ts
@@ -91,7 +91,7 @@ describe('File Commands', () => {
       sinon.assert.notCalled(vsCodeStub.window.showQuickPick);
       sinon.assert.calledWith(
         vsCodeStub.workspace.fs.writeFile,
-        TestUri.parse('colab://m-s-foo/my-file.txt'),
+        TestUri.parse('colab://m-s-foo/content/my-file.txt'),
         fileContent,
       );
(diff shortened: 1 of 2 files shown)


amtoney524 (approved):
nice fix!

src/colab/commands/files.unit.test.ts
```diff
@@ -91,7 +91,7 @@ describe('File Commands', () => {
       sinon.assert.notCalled(vsCodeStub.window.showQuickPick);
       sinon.assert.calledWith(
         vsCodeStub.workspace.fs.writeFile,
-        TestUri.parse('colab://m-s-foo/my-file.txt'),
```
mounikasomisetty, line 94: *nit* maybe we should declare the common part as a const?

Issue #204, opened 2025-11-11
Cloud auth (`google.colab.auth`) is unsupported.


gogasca:
## Description

Can't access GCP Services via Vscode
When I run:

```
from google.colab import auth
auth.authenticate_user()
```

I get a prompt with a link, but can access it or copy it

## Environment

- VS Code version: Latest
- Colab extension version: 
- Jupyter extension version: N/A
- OS (Linux | Mac | Windows): Mac

## Repro Steps

New cell

```
from google.colab import auth
auth.authenticate_user()
```

## Expected Behaviour

Can access link and authenticate

## Actual Behaviour

Unable to access link

## Additional context

<img width="3584" height="2240" alt="Image" src="https://github.com/user-attachments/assets/d7ede81e-3a0c-4619-9ede-81240cbf3964" />

## Logs

Logs from `Colab` in the `Output` panel (`View` > `Output`, change the drop-down the upper-right of the `Output` panel to `Colab`).

```
11:13:29.578 [debug] Handle Execution of Cells 4 for ~/Downloads/BigQuery basics.ipynb
11:13:29.638 [debug] Kernel acknowledged execution of cell 4 @ 1762888409638
11:13:33.802 [warn] Cell completed with errors Iu [Error]: Error fetching credentials
    at n.execute (~/.vscode/extensions/ms-toolsai.jupyter-2025.9.1-darwin-x64/dist/extension.node.js:302:4958)
    at processTicksAndRejections (node:internal/process/task_queues:105:5) {
  ename: 'AuthorizationError',
  evalue: 'Error fetching credentials',
  traceback: [
    '\x1B[0;31m---------------------------------------------------------------------------\x1B[0m',
    '\x1B[0;31mAuthorizationError\x1B[0m                        Traceback (most recent call last)',
    '\x1B[0;32m/tmp/ipython-input-1437431822.py\x1B[0m in \x1B[0;36m<cell line: 0>\x1B[0;34m()\x1B[0m\n' +
      '\x1B[1;32m      1\x1B[0m \x1B[0;32mfrom\x1B[0m \x1B[0mgoogle\x1B[0m\x1B[0;34m.\x1B[0m\x1B[0mcolab\x1B[0m \x1B[0;32mimport\x1B[0m \x1B[0mauth\x1B[0m\x1B[0;34m\x1B[0m\x1B[0;34m\x1B[0m\x1B[0m\n' +
[...]

PR #394, opened 2026-02-03
feat: handle `auth_user_ephemeral` in code

hjjackyang:
This is a prerequisite to properly handle `auth.authenticate_user()` in our extension. Ultimately, auth for `auth.authenticate_user()` behaves very similarly to `drive.mount()`. The main difference is it uses a different auth type `auth_user_ephemeral` and shows some different wording in the consent prompt.

For now, this change is a no-op for `auth.authenticate_user()`. A subsequent change (in a separate PR) will be required to initialize the kernel and set a custom env variable `USE_AUTH_EPHEM` on the server.

---

For `drive.mount()`, this change slightly modifies the error message user sees if user cancels authorization midway. This is to avoid unnecessary conditional statements in the code between `dfs_ephemeral` and `auth_user_ephemeral` auth types.

[Drive mounting error demo](https://screencast.googleplex.com/cast/NTY4NDU1OTE5MzU3MTMyOHwwMzMwOGY1My02Nw)

---

Related GitHub issue: https://github.com/googlecolab/colab-vscode/issues/204
Internal tracking bug: b/480515438

### DIFF
--- a/src/jupyter/colab-proxy-websocket.unit.test.ts
+++ b/src/jupyter/colab-proxy-websocket.unit.test.ts
@@ -140,7 +140,7 @@ describe('colabProxyWebSocket', () => {
     const rawColabRequestMessage = {
       header: { msg_type: 'colab_request' },
       content: {
-        request: { authType: 'dfs_ephemeral' },
+        request: { authType: AuthType.DFS_EPHEMERAL },
       },
       metadata: {
         colab_request_type: 'request_auth',
@@ -234,7 +234,7 @@ describe('colabProxyWebSocket', () => {
     it('does not trigger handleEphemeralAuth if message is not a colab_request', () => {
       const rawMessage = JSON.stringify({
         header: { msg_type: 'execute_reply' },
-        content: { request: { authType: 'dfs_ephemeral' } },
+        content: { request: { authType: AuthType.DFS_EPHEMERAL } },
         metadata: { colab_request_type: 'request_auth', colab_msg_id: 1 },
       });
 
--- a/src/colab/client.unit.test.ts
+++ b/src/colab/client.unit.test.ts
@@ -23,6 +23,7 @@ import {
   Outcome,
   ListedAssignments,
   RuntimeProxyInfo,
+  AuthType,
 } from './api';
 import {
   ColabClient,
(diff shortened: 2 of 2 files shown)


src/jupyter/colab-proxy-websocket.ts
```diff
@@ -211,7 +216,9 @@ function isColabAuthEphemeralRequest(
 interface ColabAuthEphemeralRequestMessage {
   header: { msg_type: 'colab_request' };
   content: {
-    request: { authType: 'dfs_ephemeral' };
+    request: {
+      authType: 'dfs_ephemeral' | 'auth_user_ephemeral';
```
amtoney524: nit: worth defining a type (e.g., `type AuthType = 'dfs_ephemeral' | 'auth_user_ephemeral'`) since these strings are referenced a few times in the PR?

src/jupyter/colab-proxy-websocket.unit.test.ts
```diff
@@ -145,7 +145,8 @@ describe('colabProxyWebSocket', () => {
         colab_request_type: 'request_auth',
         colab_msg_id: testRequestMessageId,
       },
-    });
+    };
+    const supportedAuthTypes = ['dfs_ephemeral', 'auth_user_ephemeral'];
```
amtoney524: super nit: worth moving closer to L171?

src/jupyter/colab-proxy-websocket.ts
hjjackyang: Thanks! Defined a common type in `./colab/api.ts`.

src/jupyter/colab-proxy-websocket.unit.test.ts
hjjackyang: Done!

src/colab/api.ts
```diff
@@ -83,6 +83,9 @@ export enum Shape {
   // VERYHIGHMEM (2) is deprecated.
 }
 
+/** Colab supported auth types. */
+export type AuthType = 'dfs_ephemeral' | 'auth_user_ephemeral';
```
peap: Why not an enum, like several other sets of string values here?

hjjackyang: No particular reason!

I guess I started with a single string literal value (i.e. `dfs_ephemeral`) everywhere and just never thought of changing it to enum.

Now that you brought it up, changing it to enum makes the code so much cleaner! Thanks!

PR #381, opened 2026-01-27
feat: add command to open Colab terminal with WebSocket integration

jen6:
Adds a new command `Colab: Open Colab Terminal` that opens an interactive terminal connected to a Colab runtime via WebSocket. 

#247 

image

Issue #247, opened 2025-11-18
(Now supported!): Support terminal access for Colab


Boyu-Qian:
I don't see a way to write shell command to Colab

amtoney524:
Thanks for reporting @Boyu-Qian , we'll take a closer look! Wondering if the `%%shell` magic might help in the meantime?

kevineger:
While we can't start work on this right away (we're burning down some more pressing bugs and common feature requests), this is likely pretty feasible by leveraging the [Pseudoterminal](https://code.visualstudio.com/api/references/vscode-api#Pseudoterminal) in VS Code, and connecting to the same socket that Colab web uses to support the in-app terinal.

Leaving that note in case anyone else is interested in picking this up 🙂

jen6:
@kevineger Could you please review this PR? I tried it. #381

hjjackyang:
@jen6 - FYI, your PR is now merged into `main` (with more touch-ups after a few more rounds of reviews), and will be included in our next release (date TBD).

Thanks again for doing the heavy lifting!

hjjackyang:
📣 We just released `v0.3.0` with the new experimental Colab terminal feature included!

See https://github.com/googlecolab/colab-vscode/wiki/User-Guide#colab-terminal for more info.

I will keep this issue open to collect feedback / bug 🐛 reports until the feature is more stable and turned on by default.

AusCigarRef:
Terminal text silos after opening a gemini or file view tab on top of the terminal.
Terminal copy and paste copies a lot of whitespace characters and ruins indent when pasting to Colab cell.

kevineger:
@AusCigarRef mind opening a separate issue with screenshots if you believe there's a bug with the extension's terminal support? If I'm following correctly, I don't believe that copy/paste behaviour is unique to this extension. Happy to follow up, just easier done in a standalone issue.

kevineger:
Woops, re-opening until we turn this on by default (maybe next release).

PR #398, opened 2026-02-06
feat: Colab terminal behind experimental config

hjjackyang:
This PR introduces a new `Colab: Open Terminal` command that opens an interactive terminal connected to a Colab runtime via WebSocket. This new command is gated behind an _experimental_ `colab.terminal` config.

[Demo screencast](https://screencast.googleplex.com/cast/NjMzMjU1NzY1MzE4MDQxNnxlZTJlM2NkMS1lZg)

---

Related GitHub issue: https://github.com/googlecolab/colab-vscode/issues/247

Credit to @jen6 for https://github.com/googlecolab/colab-vscode/pull/381.

### DIFF
--- a/src/colab/commands/terminal.unit.test.ts
+++ b/src/colab/commands/terminal.unit.test.ts
@@ -6,11 +6,14 @@
 
 import { randomUUID } from 'crypto';
 import sinon, { SinonStubbedInstance } from 'sinon';
-import { ExtensionTerminalOptions } from 'vscode';
+import { ExtensionTerminalOptions, QuickPick, QuickPickItem } from 'vscode';
 import { Variant } from '../../colab/api';
 import { AssignmentManager } from '../../jupyter/assignments';
 import { ColabAssignedServer } from '../../jupyter/servers';
-import { buildQuickPickStub } from '../../test/helpers/quick-input';
+import {
+  buildQuickPickStub,
+  QuickPickStub,
+} from '../../test/helpers/quick-input';
 import { TestUri } from '../../test/helpers/uri';
 import { newVsCodeStub, VsCodeStub } from '../../test/helpers/vscode';
 import { openTerminal } from './terminal';
@@ -40,6 +43,23 @@ describe('openTerminal command', () => {
   });
 
   describe('Server Selection', () => {
+    it('requests extension source for servers', async () => {
+      const server1 = buildColabAssignedServer({
+        label: 'Server 1',
+        endpoint: 'test-endpoint-1',
+        baseUrl: 'https://server1.example.com',
+        token: 'token1',
+      });
+      (assignmentManager.getServers as sinon.SinonStub).resolves([server1]);
+
+      await openTerminal(vsCodeStub.asVsCode(), assignmentManager);
+
+      sinon.assert.calledWith(
+        assignmentManager.getServers as sinon.SinonStub,
+        'extension',
+      );
+    });
+
     it('shows info message when no servers available', async () => {
       (assignmentManager.getServers as sinon.SinonStub).resolves([]);
 
@@ -79,7 +99,7 @@ describe('openTerminal command', () => {
       );
     });
 
-    it('shows QuickPick with multiple servers', async () => {
+    describe('with multiple servers', () => {
       const server1 = buildColabAssignedServer({
         label: 'Server 1',
         endpoint: 'test-endpoint-1',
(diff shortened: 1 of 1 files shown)


src/colab/commands/terminal.unit.test.ts
```diff
@@ -0,0 +1,224 @@
...
+
+      await openTerminal(vs, assignmentManager);
+
+      sinon.assert.calledOnce(vsCodeStub.window.createTerminal);
+      const options = vsCodeStub.window.createTerminal.firstCall.args[0];
+      expect(options).to.have.property('name', 'Colab Terminal: Server 1');
+    });
+
+    it('shows QuickPick with multiple servers', async () => {
+      const server1: ColabAssignedServer = {
```
prodonjs: *nit* Maybe pull out a helper function to build the assigned servers just to avoid repetition.

src/colab/terminal/colab-pseudoterminal.ts
```diff
@@ -0,0 +1,211 @@
...
+      this.terminalWebSocket.dispose();
+      // Close the terminal on connection failure
+      this.closeEmitter.fire(1);
+    }
+  }
+
+  /**
+   * Closes the terminal and cleans up resources.
+   */
+  close(): void {
```
prodonjs, line 98: Should the closeEmitter fire here like it does if the `open` call fails?

src/colab/terminal/colab-pseudoterminal.ts
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    try {
+      this.terminalWebSocket.send(data);
+    } catch (error: unknown) {
+      log.error('Failed to send input to Colab terminal:', error);
+      const errorMessage =
+        error instanceof Error ? error.message : String(error);
+      this.writeEmitter.fire(
+        `\r\n\x1b[31mError: Failed to send input: ${errorMessage}\x1b[0m\r\n`,
```
prodonjs: Should the formatting codes be put into constants since they are used in a few places? Maybe add a helper that wraps the message in the right codes.

src/colab/terminal/colab-pseudoterminal.unit.test.ts
```diff
@@ -0,0 +1,184 @@
...
+ */
+
+import { expect } from 'chai';
+import sinon from 'sinon';
+import { newVsCodeStub, VsCodeStub } from '../../test/helpers/vscode';
+import { ColabPseudoterminal } from './colab-pseudoterminal';
+import { ColabTerminalWebSocketLike } from './colab-terminal-websocket';
+
+describe('ColabPseudoterminal', () => {
+  let vsCodeStub: VsCodeStub;
```
prodonjs, line 14: Makes me miss `cleanState`!

src/colab/terminal/colab-terminal-websocket.ts
```diff
@@ -0,0 +1,391 @@
...
+        'Colab terminal WebSocket error:',
+        error,
+        this.lastWsUrl ? `url=${this.lastWsUrl}` : '',
+      );
+      if (!this.disposed) {
+        this.onErrorEmitter.fire(error);
+      }
+    });
+
+    this.ws.on('unexpected-response', (_request, response) => {
```
prodonjs: *nit* not sure if there's a convention if just using the `_` if the param is completely unused or not.

src/colab/terminal/colab-terminal-websocket.ts
```diff
@@ -0,0 +1,391 @@
...
+
+      const statusCode = response.statusCode ?? 0;
+      const statusMessage = response.statusMessage ?? '';
+      const url = this.lastWsUrl ?? '(unknown)';
+
+      // Try to read a small amount of the response body for debugging.
+      const chunks: Buffer[] = [];
+      let totalBytes = 0;
+      response.on('data', (chunk: Buffer) => {
+        if (totalBytes >= 2048) {
```
prodonjs: Any significance to 2k? Should we move it to a constant?

src/colab/terminal/colab-terminal-websocket.ts
```diff
@@ -0,0 +1,391 @@
...
+          return;
+        }
+        chunks.push(chunk);
+        totalBytes += chunk.length;
+      });
+      response.on('end', () => {
+        const body = Buffer.concat(chunks).toString('utf8').trim();
+        if (body.length) {
```
prodonjs, line 293: *nit* Since the first 4 args are the same in both cases may use the spread syntax if possible to provide the args?

prodonjs (changes requested):
I'm definitely not all that familiar with the code here so take my review with a grain of salt as I mostly looked for stylistic things and other bits rather than deeply following all of the logic since I am certain you are closer to it than I am. Overall it looks pretty clean so I'd defer to your judgment on how much further to go, particularly since it's flag-gated.

Awesome to see!

src/colab/commands/terminal.unit.test.ts
hjjackyang: Great suggestion! I didn't pay too much attention to tests when I did my initial review.

Done and also refactored a few other places, e.g. using `calledOnceWithMatch`.

src/colab/terminal/colab-pseudoterminal.ts
hjjackyang: Great catch!

It probably won't make much difference in our use case, but good to have for completeness. Also added a test to cover it.

src/colab/terminal/colab-pseudoterminal.ts
hjjackyang: Done!

Refactored ANSI codes into an enum and also extracted formatting logic into a helper method.

src/colab/terminal/colab-pseudoterminal.unit.test.ts
hjjackyang: Me too!

src/colab/terminal/colab-terminal-websocket.ts
hjjackyang: Good catch! Yes, there is.

I think when there are multiple unused params, we can prefix them with `_` so they have different names. In this case, since there is only one unused param, we can directly name it `_`.

src/colab/terminal/colab-terminal-websocket.ts
hjjackyang: I don't think so. Probably just an arbitrary number to keep the output size reasonable. I moved it into a constant.

src/colab/terminal/colab-terminal-websocket.ts
hjjackyang: Done! Let me know if I did it right.

prodonjs (approved):
Looks good!

src/colab/commands/terminal.unit.test.ts
```diff
@@ -0,0 +1,201 @@
...
+      (assignmentManager.getServers as sinon.SinonStub).resolves([server1]);
+      const mockTerminal = { show: sinon.stub() };
+      vsCodeStub.window.createTerminal.returns(mockTerminal as never);
+
+      await openTerminal(vsCodeStub.asVsCode(), assignmentManager);
+
+      sinon.assert.calledOnce(mockTerminal.show);
+    });
+
+    it('requests extension source for servers', async () => {
```
amtoney524: nit: should this be nested in the Server Selection clause (since this populates the quickpick)?

src/colab/commands/terminal.unit.test.ts
```diff
@@ -0,0 +1,201 @@
...
+      sinon.assert.calledOnceWithMatch(
+        vsCodeStub.window.createTerminal,
+        sinon.match(
+          (options: ExtensionTerminalOptions) =>
+            options.name === 'Colab Terminal: Server 1',
+        ),
+      );
+    });
+
+    it('shows QuickPick with multiple servers', async () => {
```
amtoney524: nit: worth adding a case that validates a terminal is created from the selected server when multiple servers are present?

src/colab/terminal/colab-pseudoterminal.ts
```diff
@@ -0,0 +1,275 @@
...
+      // Notify user of disconnection
+      this.writeEmitter.fire(
+        this.format('Connection to Colab terminal closed.', {
+          color: 'warning',
+          withLeadingNewline: true,
+        }),
+      );
+
+      // Close the VS Code terminal
+      this.closeEmitter.fire(0);
```
amtoney524, line 220: super nit: should this be a non-zero exit code if (iiuc) websocket connection is lost?

amtoney524 (approved):
looks great!!

src/colab/terminal/colab-pseudoterminal.ts
hjjackyang: Good question! I just did a quick Google search and AI Overview says:

> A websocket closing and causing a terminal session to terminate is generally considered an expected behavior rather than an "error".

In this case, I'll leave it as is.

src/colab/commands/terminal.unit.test.ts
hjjackyang: Done!

src/colab/commands/terminal.unit.test.ts
hjjackyang: Good call! Done.

PR #401, opened 2026-02-06
feat: add the ability to use experiment flags from the colab web UI

mounikasomisetty:
Add the ability to use experiment flags from the Colab web UI (as long as they are tagged with "vscode"). We'll need to wait till the corresponding endpoint is in production before we merge this change. The only experiment flag declared as part of this commit is `RuntimeVersionNames`, in preparation of adding the runtime version selector. 

Screencast: https://screencast.googleplex.com/cast/NjM0ODM0NDI3NTIzODkxMnxiYjBlZGNjNy0zMg

### DIFF
--- a/src/colab/experiment-state.ts
+++ b/src/colab/experiment-state.ts
@@ -6,7 +6,11 @@
 
 import { log } from '../common/logging';
 import { AsyncToggle } from '../common/toggleable';
-import { ExperimentFlag, ExperimentFlagValue, EXPERIMENT_FLAG_DEFAULT_VALUES } from './api';
+import {
+  ExperimentFlag,
+  ExperimentFlagValue,
+  EXPERIMENT_FLAG_DEFAULT_VALUES,
+} from './api';
 import { ColabClient } from './client';
 
 /** Gets the value of an experiment flag.
@@ -62,4 +66,3 @@ let flags: ReadonlyMap<ExperimentFlag, ExperimentFlagValue> = new Map<
   ExperimentFlag,
   ExperimentFlagValue
 >();
-
--- a/src/colab/experiment-state.unit.test.ts
+++ b/src/colab/experiment-state.unit.test.ts
@@ -37,11 +37,8 @@ describe('ExperimentStateProvider', () => {
   });
 
   it('initializes with default flag values', () => {
-    expect(getFlag(ExperimentFlag.RuntimeVersionNames)).to.deep.equal(
-      [],
-    );
-  }
-);
+    expect(getFlag(ExperimentFlag.RuntimeVersionNames)).to.deep.equal([]);
+  });
 
   it('fetches experiment state with auth when turned on', async () => {
     const experiments = new Map([[ExperimentFlag.RuntimeVersionNames, true]]);


src/colab/client.ts
```diff
@@ -433,15 +463,18 @@ export class ColabClient {
...
   ): Promise<unknown> {
     // The Colab API requires the authuser parameter to be set.
     if (endpoint.hostname === this.colabDomain.hostname) {
       endpoint.searchParams.append('authuser', '0');
     }
-    const token = await this.getAccessToken();
     const requestHeaders = new Headers(init.headers);
     requestHeaders.set(ACCEPT_JSON_HEADER.key, ACCEPT_JSON_HEADER.value);
-    requestHeaders.set(AUTHORIZATION_HEADER.key, `Bearer ${token}`);
+    if (requireAccessToken) {
```
hjjackyang, line 474: Any reason to change this in `getAssignment`? Will `getAssignment` ever work without access token?

src/colab/experiment-state.ts
```diff
@@ -0,0 +1,73 @@
...
+
+function getDefaultValueForExperimentFlag(
+  flag: ExperimentFlag,
+): ExperimentFlagValue {
+  switch (flag) {
+    // Add default values for experiment flags here.
+    default:
+      return [];
+  }
+}
```
hjjackyang: Instead of defining default values for flags in a separate function like this, perhaps we can define the default values in a `Record` and have them defined side by side with their corresponding keys, i.e. `ExperimentFlag` enum?

For example:
```ts
export enum ExperimentFlag {
  RuntimeVersionNames = 'runtime_version_names',
}

export const EXPERIMENT_FLAG_DEFAULT_VALUES: Record<
  ExperimentFlag,
  ExperimentFlagValue
> = {
  [ExperimentFlag.RuntimeVersionNames]: '',
};
```

Just a thought.

src/colab/experiment-state.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Provides experiment state information from the Colab backend.
+ */
+export class ExperimentStateProvider extends AsyncToggle {
+  constructor(private readonly client: ColabClient) {
+    super();
+    // Reset experiment flags.
+    flags = new Map<ExperimentFlag, ExperimentFlagValue>();
+    this.getExperimentState = this.getExperimentState.bind(this);
```
hjjackyang, line 33: Nit: Is this line required?

src/colab/experiment-state.ts
```diff
@@ -0,0 +1,73 @@
...
+      return [];
+  }
+}
+
+/** Gets the value of an experiment flag.
+ *
+ * @param flag - The experiment flag to get.
+ * @returns The value of the experiment flag.
+ */
+export function getFlag(flag: ExperimentFlag): ExperimentFlagValue {
```
hjjackyang: Could we move this exported function to the top of the file (or at least above the non-exported members)?

src/colab/experiment-state.ts
```diff
@@ -0,0 +1,73 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
hjjackyang:
```suggestion
 * Copyright 2026 Google LLC
```

src/colab/experiment-state.unit.test.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
hjjackyang:
```suggestion
 * Copyright 2026 Google LLC
```

src/colab/experiment-state.unit.test.ts
```diff
@@ -0,0 +1,118 @@
...
+describe('ExperimentStateProvider', () => {
+  let colabClientStub: SinonStubbedInstance<ColabClient>;
+  let provider: TestExperimentStateProvider;
+
+  beforeEach(() => {
+    colabClientStub = sinon.createStubInstance(ColabClient);
+    provider = new TestExperimentStateProvider(colabClientStub);
+
+    // Default value of the flag
+    expect(getFlag(ExperimentFlag.RuntimeVersionNames)).to.be.deep.equal([]);
```
hjjackyang: Let's put default value assertion into a standalone test instead of in `forEach`.

src/colab/experiment-state.ts
```diff
@@ -0,0 +1,73 @@
...
+  ): Promise<void> {
+    try {
+      const result = await this.client.getExperimentState(
+        requireAccessToken,
+        signal,
+      );
+      if (result.experiments) {
+        flags = result.experiments;
+      }
+    } catch (e) {
```
hjjackyang: Nit: Catch `e: unknown` to avoid implicit `any` type.

```suggestion
    } catch (e: unknown) {
```

http://go/tsjs-style#any-unknown

src/colab/experiment-state.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { log } from '../common/logging';
+import { AsyncToggle } from '../common/toggleable';
+import { ExperimentFlag, ExperimentFlagValue } from './api';
+import { ColabClient } from './client';
+
+let flags: Map<ExperimentFlag, ExperimentFlagValue> = new Map<
```
hjjackyang: Super nit: Given this `flags` map is never modified in place (only reassigned), suggest to type it as `ReadonlyMap` to reduce mutability.

```suggestion
let flags: ReadonlyMap<ExperimentFlag, ExperimentFlagValue> = new Map<
```

hjjackyang (changes requested):
I only realized that you put the PR back to Draft after I finished my reviews, so I'm sending them anyway. Feel free to only address them after you finalize your PR!

src/colab/experiment-state.ts
mounikasomisetty: Yes, I believe the callers of turnOn/turnOff needs this to be able to access that method

src/colab/experiment-state.ts
mounikasomisetty: Great idea! Done!

src/colab/experiment-state.ts
mounikasomisetty: Done!

src/colab/experiment-state.ts
mounikasomisetty: Done

src/colab/experiment-state.unit.test.ts
mounikasomisetty: Done

src/colab/experiment-state.ts
mounikasomisetty: Done

src/colab/experiment-state.ts
mounikasomisetty: Done

src/colab/experiment-state.unit.test.ts
mounikasomisetty: Done

src/colab/client.ts
mounikasomisetty: I'm not sure what you mean here? Most of our requests do require an access token. I did double check to make sure that I didn't change any other method but lmk if I missed something.

hjjackyang: Oh my bad. The way this diff was presented to me appeared to be part of `getAssignment`, but it was for `issueRequest`. NVM!

PR #402, opened 2026-02-06
feat: add a telemetry entrypoint

amtoney524:
This change adds a telemetry entrypoint, which exports an `initializeTelemetry` method and a `telemetry` object with convenience methods for logging. 

note: this change is a noop as it's not currently being referenced from the application

### DIFF
--- a/src/telemetry/index.ts
+++ b/src/telemetry/index.ts
@@ -62,14 +62,12 @@ export const telemetry = {
       log({
         error_event: { name: e.name, msg: e.message, stack: e.stack ?? '' },
       });
-      return;
-    }
-    if (typeof e === 'string') {
+    } else if (typeof e === 'string') {
       log({ error_event: { name: 'Error', msg: e, stack: '' } });
-      return;
+    } else {
+      const msg = e ? JSON.stringify(e) : String(e);
+      log({ error_event: { name: 'Error', msg, stack: '' } });
     }
-    const msg = e ? JSON.stringify(e) : String(e);
-    log({ error_event: { name: 'Error', msg, stack: '' } });
   },
 };
 


src/jupyter/jupyter-extension.ts
```diff
@@ -11,13 +11,16 @@ import { getPackageInfo } from '../config/package-info';
 
 const JUPYTER_SEMVER_RANGE = '>=2025.0.0';
 
+/** The identifier of the Jupyter extension. */
+export const JUPYTER_EXT_IDENTIFIER = 'ms-toolsai.jupyter';
```
hjjackyang, line 15: Nit: Perhaps also update this test to reference the new variable?

https://github.com/googlecolab/colab-vscode/blob/main/src/jupyter/jupyter-extension.unit.test.ts#L38

src/telemetry/client.ts
```diff
@@ -34,6 +46,11 @@ export interface ColabLogEvent {
   vscode_version: string;
 }
 
+// The Colab log event structure.
+// TODO: Convert to proto definition.
+// TODO: Record events for MVP CUJs.
+export type ColabLogEvent = ColabLogEventBase & ColabEvent;
```
hjjackyang: I personally like to keep top level constructs at the top of the file, so they are more easily discoverable. This also nicely keeps the exported types/interfaces at the top and non-exported ones at the bottom.

```ts
export type ColabLogEvent = ColabLogEventBase & ColabEvent;

export interface ColabLogEventBase {
  ...
}

export type ColabEvent =
  | { activation_event: ColabActivationEvent }
  | { error_event: ColabErrorEvent };

type ColabActivationEvent = Record<string, never>;

interface ColabErrorEvent {
  ...
}
```

Nit: Also, I wonder if we should move these interfaces/types to a separate `model.ts` file?

src/telemetry/index.ts
```diff
@@ -0,0 +1,82 @@
...
+ */
+import assert from 'assert';
+import vscode from 'vscode';
+import { Disposable } from 'vscode';
+import { getPackageInfo } from '../config/package-info';
+import { JUPYTER_EXT_IDENTIFIER } from '../jupyter/jupyter-extension';
+import { ClearcutClient, ColabLogEventBase, ColabEvent } from './client';
+
+// The identifier for the Colab extension.
+const COLAB_EXT_IDENTIFIER = 'google.colab';
```
hjjackyang: Nit: Shall we define this as an exported const in `src/config/constants.ts` instead?

If so, we can also update the following test to reference the new exported const:
https://github.com/googlecolab/colab-vscode/blob/main/src/test/extension.vscode.test.ts#L12

src/telemetry/index.ts
```diff
@@ -0,0 +1,82 @@
...
+function log(event: ColabEvent) {
+  // TODO: Add listener for telemetry setting and return early if opted-out
+  // TODO: Skip logging in integration tests
+  if (!client) {
+    return;
+  }
+  client.log({
+    ...baseLog,
+    ...event,
+    timestamp: new Date().toISOString(),
```
hjjackyang, line 83: Since you always construct `timestamp` separately, consider keeping it out of `ColabLogEventBase`, and making `type ColabLogEvent = ColabLogEventBase & ColabEvent & { timestamp: string }`?

This way you can also avoid `Omit<ColabLogEventBase, 'timestamp'>`.

src/telemetry/index.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * A collection of functions for logging telemetry events.
+ */
+export const telemetry = {
+  logActivation: () => {
+    log({ activation_event: {} });
+  },
+  logError: (e: Error) => {
```
hjjackyang: Nit: Since errors are always caught as `unknown` in a `try/catch`, could we make this interface support `unknown` type instead?

```suggestion
  logError: (e: unknown) => {
```

And handle it properly inside the function of course. For example:
```ts
let event: ColabEvent;
if (e instanceof Error) {
  errorEvent = { error_event: { name: e.name, msg: e.message, stack: e.stack ?? '' } };
} else if (typeof e === 'string') {
  errorEvent = { error_event: { name: 'Error', msg: e, stack: '' } };
} else {
  errorEvent = { error_event: { name: 'Unknown', msg: String(e), stack: '' } };
}
log(event);
```

hjjackyang (changes requested):
Proposed a few non-trivial changes. Please take a look and let me know what you think. Thanks!

src/jupyter/jupyter-extension.ts
amtoney524: good idea, done!

src/telemetry/client.ts
amtoney524: nice suggestion, moved the exported interfaces / types upwards and together. +1 to having a model.ts file, mind if that change is in a follow up PR?

src/telemetry/index.ts
amtoney524: nice, sgtm!

src/telemetry/index.ts
amtoney524: thanks for the pointer, done!

src/telemetry/index.ts
amtoney524: good idea - tweaked this logic a bit to try to stringify the error if we can, otherwise we might log something like `[object Object]` when potentially we could have more detail. also added the corresponding test coverage

hjjackyang: Great, thanks!

src/telemetry/index.ts
```diff
@@ -0,0 +1,87 @@
...
+        error_event: { name: e.name, msg: e.message, stack: e.stack ?? '' },
+      });
+      return;
+    }
+    if (typeof e === 'string') {
+      log({ error_event: { name: 'Error', msg: e, stack: '' } });
+      return;
+    }
+    const msg = e ? JSON.stringify(e) : String(e);
+    log({ error_event: { name: 'Error', msg, stack: '' } });
```
hjjackyang: Nit: I know we generally prefer early return per http://go/tott/733. However, in this particular case, I find it more clear if we use `if/else` per http://go/tott/680 to emphasize the complementary nature of each condition. Thoughts?

```suggestion
    if (e instanceof Error) {
      log({
        error_event: { name: e.name, msg: e.message, stack: e.stack ?? '' },
      });
    } else if (typeof e === 'string') {
      log({ error_event: { name: 'Error', msg: e, stack: '' } });
    } else {
      const msg = e ? JSON.stringify(e) : String(e);
      log({ error_event: { name: 'Error', msg, stack: '' } });
    }
```

amtoney524: oh nice, TIL! sounds good to me, and it's a bit more compact as well - thanks!

amtoney524:
Thanks for the detailed reviews!!

PR #404, opened 2026-02-08
feat: retry logs on 5XX errors

amtoney524:
Retries logging when Clearcut returns a 500 error.

To facilitate a "retry", events that weren't logged are requeued at the front of the queue (based on remaining capacity). Why on 500s? This is the recommended behavior based on the Clearcut HTTP API docs.

Future work: once we're propagating the token, we can also retry on 401. we could also consider adding some exponential backoff on retries if needed

### DIFF
--- a/src/telemetry/client.ts
+++ b/src/telemetry/client.ts
@@ -135,8 +135,12 @@ export class ClearcutClient implements Disposable {
       },
     });
     const response = await fetch(request);
-    // TODO: Retry on 401 and 5xx.
+    // TODO: handle 401 once token is included in request
     if (!response.ok) {
+      if (response.status >= 500) {
+        this.requeue(events); // Retry on next flush
+        return MIN_WAIT_BETWEEN_FLUSHES_MS;
+      }
       throw new Error(
         `Failed to issue request ${request.method} ${request.url}: ${response.statusText}`,
       );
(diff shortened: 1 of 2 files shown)


src/telemetry/client.ts
```diff
@@ -154,6 +158,22 @@ export class ClearcutClient implements Disposable {
...
+    // Queue is full, which means the oldest events should be dropped.
+    if (capacity === 0) {
+      return;
+    }
+    if (capacity >= events.length) {
+      this.pendingEvents.unshift(...events);
+      return;
+    }
+    // Only keep the most recent events within the queue's capacity.
+    this.pendingEvents.unshift(...events.slice(-capacity));
```
hjjackyang, line 175: Nit (unrelated to this PR in particular): Before our capacity management logic gets more complicated, I wonder if we should revisit how this is handled before we start losing events. Luckily we haven't started logging real events yet, so this isn't too urgent to deal with.

One proposal is, if we reach capacity for the current flush, just flush the first X items and leave the remaining items in the queue until the next flush? This way we will not lose any events.

amtoney524: Good point! We could definitely consider something like that, though one constraint with Clearcut is that clients obey the flush interval in their response. With that in mind, I figured dropping events to start may not be too bad since the client logs will only provide a partial view anyways (based on users being able to [opt out](https://code.visualstudio.com/api/extension-guides/telemetry#dos-and-donts) of telemetry). I'll add this as an item to tomorrow's sync to continue the discussion - thanks for raising this!

PR #408, opened 2026-02-10
feat: telemetry is enabled based on VS Code setting

amtoney524:
This change enables telemetry only when the VS Code setting for `Telemetry: Telemetry Level` is set to `all`. This is based on the requirements in VS Code's [Telemetry Extension Authors Guide](https://code.visualstudio.com/api/extension-guides/telemetry#dos-and-donts).

Since the `client.log` function is the trigger for an event being tracked and subsequently logged (when the flush interval's met), it's not called when telemetry's disabled. This blocks logging of additional events while telemetry's disabled, including retries of 5XX requests; it also does not place any new events in the queue.

Below is the corresponding setting in VS Code. Note that there are additional options for `crash` and `error` telemetry; however, those don't fire the `onDidChangeTelemetryEnabled` event with `true` so it seems we're limited to just `all` at the moment. In the future, we could consider adding a Colab-specific telemetry setting that could allow only certain events.

Screenshot 2026-02-10 at 10 12 42 AM

Additionally, we'll want to incorporate a feature flag to use as an off switch, which is a requirement from Clearcut in case of catastrophic issues.

### DIFF
--- a/src/telemetry/telemetry.unit.test.ts
+++ b/src/telemetry/telemetry.unit.test.ts
@@ -83,7 +83,7 @@ describe('Telemetry Module', () => {
     sinon.assert.notCalled(logStub);
   });
 
-  it('enables telemetry when onDidChangeTelemetry fires with true', () => {
+  it('enables telemetry when isTelemetryEnabled is true', () => {
     const logStub = sinon.stub(ClearcutClient.prototype, 'log');
     vs.env.isTelemetryEnabled = false;
     // Maintain a reference to this stub as that's the reference telemetry has.
@@ -101,7 +101,7 @@ describe('Telemetry Module', () => {
     sinon.assert.calledOnce(logStub);
   });
 
-  it('disables telemetry when onDidChangeTelemetry fires with false', () => {
+  it('disables telemetry when isTelemetryEnabled is false', () => {
     const logStub = sinon.stub(ClearcutClient.prototype, 'log');
     // Maintain a reference to this stub as that's the reference telemetry has.
     const vscodeStub = vs.asVsCode();


src/telemetry/index.ts
```diff
@@ -40,12 +42,20 @@ export function initializeTelemetry(vs: typeof vscode): Disposable {
     vscode_version: vs.version,
   };
 
+  isTelemetryEnabled = vs.env.isTelemetryEnabled;
+  const telemetrySettingListener = vs.env.onDidChangeTelemetryEnabled(
```
hjjackyang: I feel the code will be much cleaner if we just check `vs.env.isTelemetryEnabled` at logging time instead of creating a listener and maintaining our own variable.

Looking at the implementation behind `isTelemetryEnabled` on VS Code side, it's just checking a stored `_level` variable when it's called, so performance impact should be minimal:

https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHost.api.impl.ts#L421

https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHostTelemetry.ts#L59

amtoney524: I agree, though wouldn't that increase the signature of each of the logging functions? It seemed like we'd have to plumb the `vs` object through each time (and ensure its accessible at each call site). wdyt?

amtoney524: oh you know what, maybe we can use a closure here, will take a closer look

hjjackyang: > I agree, though wouldn't that increase the signature of each of the logging functions? It seemed like we'd have to plumb the `vs` object through each time (and ensure its accessible at each call site). wdyt?

Oh, I forgot about having to pass `vs` object around. In this case, I'm good with what you have!

(Feel free to either merge as is or spend some time on the closure option)

amtoney524: updated to include the closure, ended up being much simpler though a little bit of overhead in tests :) thanks a lot for the suggestion!

src/telemetry/telemetry.unit.test.ts
```diff
@@ -73,6 +73,51 @@ describe('Telemetry Module', () => {
...
+    const vscodeStub = vs.asVsCode();
+    disposeTelemetry = initializeTelemetry(vscodeStub);
+
+    telemetry.logActivation();
+    sinon.assert.notCalled(logStub);
+    logStub.resetHistory();
+
+    // Required to change read-only property
+    (vscodeStub.env as { isTelemetryEnabled: boolean }).isTelemetryEnabled =
+      true;
```
hjjackyang, line 99: Nit: Does changing `vs.env.isTelemetryEnabled = true` not work here?

Feel free to ack if not.

amtoney524: It doesn't unfortunately :/ I think this is because `.asVsCode()` is returns a new instance and that's the one being referenced in `isTelemetryEnabled()`. 

To be sure this is WAI, I did verify `isTelemetryEnabled()` returns the expected value with the vs code debugger (e.g., change the settings, then log an event)

amtoney524:
Thanks for the detailed review (as always!!)

PR #411, opened 2026-02-10
test: update sinon spy to stub to fix hanging tests

amtoney524:
why? when using a spy, the original method still runs, where some internals probably what caused the tests to hang. we can make the same assertion with a stub, which overrides the method behavior and fixes the issue.

### DIFF
--- a/src/telemetry/telemetry.unit.test.ts
+++ b/src/telemetry/telemetry.unit.test.ts
@@ -75,10 +75,10 @@ describe('Telemetry Module', () => {
 
   describe('logs to Clearcut', () => {
     let baseLog: ColabLogEventBase & { timestamp: string };
-    let logSpy: SinonSpy;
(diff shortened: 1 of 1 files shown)


src/telemetry/telemetry.unit.test.ts
```diff
@@ -78,7 +78,7 @@ describe('Telemetry Module', () => {
     let logSpy: SinonSpy;
 
     beforeEach(() => {
-      logSpy = sinon.spy(ClearcutClient.prototype, 'log');
+      logSpy = sinon.stub(ClearcutClient.prototype, 'log');
```
hjjackyang: Nit: Shall we rename the variable to be `logStub` to be consistent?

amtoney524: nice catch ty!

PR #412, opened 2026-02-10
fix: clear session and retry if issue request encounters a 401

mounikasomisetty:
This should help fix issues where the user revokes access while the extension is still running [screencast](https://screencast.googleplex.com/cast/NTA2MTkwNzk1MzIyMTYzMnxhNDEwMGNlNy0xYg)

### DIFF
--- a/src/colab/client.ts
+++ b/src/colab/client.ts
@@ -473,20 +473,21 @@ export class ColabClient {
 
     let response: Response | undefined;
     let request: Request | undefined;
+    const requestHeaders = new Headers(init.headers);
+    requestHeaders.set(ACCEPT_JSON_HEADER.key, ACCEPT_JSON_HEADER.value);
+    requestHeaders.set(
+      COLAB_CLIENT_AGENT_HEADER.key,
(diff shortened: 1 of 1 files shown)


src/colab/client.ts
```diff
@@ -469,36 +470,51 @@ export class ColabClient {
...
+      requestHeaders.set(
+        COLAB_CLIENT_AGENT_HEADER.key,
+        COLAB_CLIENT_AGENT_HEADER.value,
+      );
+      request = new Request(endpoint, {
+        ...init,
+        headers: requestHeaders,
+        agent: this.httpsAgent,
       });
+      response = await fetch(request);
```
hjjackyang, line 496: Nit: Shall we wrap the for loop starting from this line to avoid rebuilding the same request?

Or is it also critical to retry `this.getAccessToken()`? If so, perhaps we can still avoid rebuilding the rest of the static request headers?

mounikasomisetty: Ah yes it is critical to rety `this.getAccessToken()` because that is what triggers the login/oauth. I moved the static request headers init out of the loop!

PR #413, opened 2026-02-11
fix: error deleting non-empty directories on server

hjjackyang:
**Why**? In the previous implementation, deleting a folder with contents throws a *400 Bad Request* error, and I don't see an option in the [Jupyter contents REST API](https://jupyter-server.readthedocs.io/en/latest/developers/rest-api.html#delete--api-contents-path) to properly handle this. As a workaround, this fix reads the directory and deletes all contents under the directory recursively first.

[Demo screencast](https://screencast.googleplex.com/cast/NTA2OTIyNjkxMzAzODMzNnw5ZGQ0MGQ3NS1kMw)

---

Internal tracking bug: b/479273773

### DIFF
--- a/src/jupyter/contents/file-system.unit.test.ts
+++ b/src/jupyter/contents/file-system.unit.test.ts
@@ -939,62 +939,64 @@ describe('ContentsFileSystemProvider', () => {
       });
 
       for (const recursive of [true, false]) {
-        it(`deletes the file (recursive: ${String(recursive)})`, async () => {
-          await fs.delete(TestUri.parse('colab://m-s-foo/foo.txt'), {
-            recursive,
+        describe(`with recursive ${String(recursive)}`, () => {
+          it('deletes the file', async () => {
+            await fs.delete(TestUri.parse('colab://m-s-foo/foo.txt'), {
+              recursive,
+            });
+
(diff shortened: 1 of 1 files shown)


src/jupyter/contents/file-system.unit.test.ts
```diff
@@ -933,106 +894,109 @@ describe('ContentsFileSystemProvider', () => {
...
+      beforeEach(() => {
+        contentsStub = stubClient('m-s-foo');
+        contentsStub.get
+          .withArgs({ path: '/foo.txt', content: 0 })
+          .resolves(FOO_CONTENT_FILE);
       });
 
-      sinon.assert.calledWith(contentsStub.delete, { path: '/foo.txt' });
-    });
+      for (const recursive of [true, false]) {
```
mounikasomisetty, line 941: *nit* Would you want to wrap the test cases in this for loop in a describe block? That way you could de-dupe the `(recursive: ${String(recursive)})` logic

mounikasomisetty (approved):
LGTM with a small nit for test readability!

src/jupyter/contents/file-system.unit.test.ts
hjjackyang: Great idea! Done.

PR #417, opened 2026-02-11
fix: error reading ipynb file on server

hjjackyang:
**Why**? The previous implementation tries to read file from server using [Jupyter Contents REST API](https://jupyter-server.readthedocs.io/en/latest/developers/rest-api.html#get--api-contents-path) without specifying a `type: 'file'` parameter. For `.ipynb` files, the server will by default handle this request as `type: 'notebook'` and respond with an empty `content` field for some reason. The solution here is to specify the `type: 'file'` parameter and force the server to return the notebook as a `file` type and encoded in `base64` format.

For the above solution (forcing `type: 'file'`) to work safely, I had to make an additional `stat` call to check for `directory` type and return early.

---

**Demo screencasts**:
- [Before](https://screencast.googleplex.com/cast/NTQ1NzYxNjMwOTMyMTcyOHw3MTU4YTM1Zi0yYg)
- [After](https://screencast.googleplex.com/cast/NjYyNTg1MDAzMTA3OTQyNHwwZjBiMGI4Ni0wZQ)

---

Internal tracking bug: b/473601011

### DIFF
--- a/src/jupyter/contents/file-system.unit.test.ts
+++ b/src/jupyter/contents/file-system.unit.test.ts
@@ -624,6 +624,7 @@ describe('ContentsFileSystemProvider', () => {
 
     it('throws file system file is a directory when the contents are for a directory', async () => {
       const contentsStub = stubClient('m-s-foo');
+      // For stat call.
       contentsStub.get
         .withArgs({ path: '/', content: 0 })
         .resolves(FOO_CONTENT_DIR);
@@ -635,9 +636,11 @@ describe('ContentsFileSystemProvider', () => {
 
     it('throws file system file is a directory when the contents not base64 encoded', async () => {
       const contentsStub = stubClient('m-s-foo');
+      // For stat call.
       contentsStub.get
         .withArgs({ path: '/foo.txt', content: 0 })
         .resolves(FOO_CONTENT_FILE);
+      // For read call.
       contentsStub.get
         .withArgs({ path: '/foo.txt', format: 'base64', type: 'file' })
         .resolves({ ...FOO_CONTENT_FILE, format: 'text' });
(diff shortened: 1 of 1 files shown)


src/jupyter/contents/file-system.ts
```diff
@@ -247,15 +248,20 @@ export class ContentsFileSystemProvider
...
+        type: ContentsGetTypeEnum.File,
       });
 
       if (
-        isDirectoryContents(content) ||
-        content.format !== 'base64' ||
+        content.format !== ContentsGetFormatEnum.Base64 ||
         typeof content.content !== 'string'
       ) {
         throw this.vs.FileSystemError.FileIsADirectory(uri);
```
amtoney524, line 267: nit: should this get updated since the directory check throws earlier?

src/jupyter/contents/file-system.unit.test.ts
```diff
@@ -624,7 +624,9 @@ describe('ContentsFileSystemProvider', () => {
 
     it('throws file system file is a directory when the contents are for a directory', async () => {
       const contentsStub = stubClient('m-s-foo');
-      contentsStub.get.resolves(CONTENT_DIR.withoutContents);
+      contentsStub.get
```
amtoney524, line 628: nit: is it worth noting which of these are for `stat` calls or is that obvious?

amtoney524 (approved):
nice find / fix!!

src/jupyter/contents/file-system.unit.test.ts
hjjackyang: Good suggestion! I guess it won't be obvious unless you are very familiar with the `stat` call already!

src/jupyter/contents/file-system.ts
hjjackyang: This check cannot be moved to earlier because the `stat` call does not force a format, so a file can be in `text` or `json` format in the `stat` response.

For now, I left this check here more as a fail-safe, but technically, these check should not fire anymore.

PR #420, opened 2026-02-12
docs: README and extension command updates

hjjackyang:
A couple of documentation updates in this PR:

1. Link our user guide wiki page in README for easier navigation from [VS Marketplace](https://marketplace.visualstudio.com/items?itemName=Google.colab).

2. Fix `package.json` to correctly reflect some commands in VS Code extension documentations. The following commands are currently incorrectly identified to be available in `commandPalette` but are really not:

   Commands before

   After the fix, they will be identified correctly in the extension documentation in VS Code:
   Commands after

   (Note: This does not impact existing commands behavior because these commands do not show up in Command Palette today anyway.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,9 @@ this extension exposes Colab servers directly in VS Code!
 ![Connecting to a new Colab server and executing a code
 cell](./docs/assets/hello-world.gif)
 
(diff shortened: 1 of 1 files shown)


amtoney524 (approved):
nice find/fix!

PR #423, opened 2026-02-12
feat: use flag as telemetry off-switch

amtoney524:
This change checks the value of a feature flag to determine if logging to Clearcut is enabled; if not, events are queued but are not sent to Clearcut (e.g., to minimize telemetry loss if temporarily disabling the flag).

In testing, I encountered a race condition where the flag value may be read before we get the flag response from Colab, resulting in the default value being returned. I observed this when logging an activation event, which ends up on the client queue, until the client's able to successfully flush to Clearcut.

It might be nice to add a promise that resolves when flags are initialized with values from Colab vs. the defaults if we start encountering this behavior in other scenarios.

### DIFF
--- a/src/telemetry/client.unit.test.ts
+++ b/src/telemetry/client.unit.test.ts
@@ -49,6 +49,7 @@ describe('ClearcutClient', () => {
     fetchStub = sinon.stub(fetch, 'default');
 
     // Since we cannot stub the export for getFlag, mock the default
+    // TODO: Replace with flag test helper once available
     enableTelemetryDefault =
       EXPERIMENT_FLAG_DEFAULT_VALUES[ExperimentFlag.EnableTelemetry];
     EXPERIMENT_FLAG_DEFAULT_VALUES[ExperimentFlag.EnableTelemetry] = true;


src/telemetry/client.unit.test.ts
```diff
@@ -34,17 +39,25 @@ const FETCH_RESPONSE_400 = new Response('', { status: 400 });
...
+    // Since we cannot stub the export for getFlag, mock the default
+    enableTelemetryDefault =
+      EXPERIMENT_FLAG_DEFAULT_VALUES[ExperimentFlag.EnableTelemetry];
+    EXPERIMENT_FLAG_DEFAULT_VALUES[ExperimentFlag.EnableTelemetry] = true;
   });
 
   afterEach(() => {
     sinon.restore();
+    EXPERIMENT_FLAG_DEFAULT_VALUES[ExperimentFlag.EnableTelemetry] =
+      enableTelemetryDefault;
```
hjjackyang, line 61: This makes me think if we should wrap the flags retrieval logic in a class like `ExperimentFlagsClient` or something, so it's easier to stub in unit tests.

As experiment flags get used more broadly, one of these days, we may forget to restore the flag value in `afterEach`, causing state leakage.

(Doesn't need to be done in this PR of course.)

amtoney524: yeah, +1! as is, this test is brittle - to your point, maybe we could consider adding a helper similar to what Kevin's done here? https://github.com/googlecolab/colab-vscode/blob/6d968da4dc698f525ae67d2a030e511f9e524636/src/test/helpers/logging.ts#L15

I'll drop a TODO and follow up with a bug internally

amtoney524:
thanks for the review!

src/telemetry/client.unit.test.ts
```diff
@@ -34,17 +39,26 @@ const FETCH_RESPONSE_400 = new Response('', { status: 400 });
...
 
   beforeEach(() => {
     fakeClock = sinon.useFakeTimers({ now: NOW, toFake: [] });
     client = new ClearcutClient();
     fetchStub = sinon.stub(fetch, 'default');
+
+    // Since we cannot stub the export for getFlag, mock the default
+    // TODO: Replace with flag test helper once available
+    enableTelemetryDefault =
+      EXPERIMENT_FLAG_DEFAULT_VALUES[ExperimentFlag.EnableTelemetry];
```
mounikasomisetty, line 54: So this works but I was thinking of adding these functions to `experiment-state.ts` just FYI!

```
/** Sets the value of an experiment flag for testing. */
export function setFlagForTest(flag: ExperimentFlag, value: ExperimentFlagValue): void {
  const newFlags = new Map(flags);
  newFlags.set(flag, value);
  flags = newFlags;
}

/** Resets the experiment flags for testing. */
export function resetFlagsForTest(): void {
  flags = new Map();
}
```

PR #424, opened 2026-02-13
feat: add ability to select runtime version to server picker

mounikasomisetty:
Add the ability to select a runtime version to the server picker. If there are no versions available, the step should be skipped and always default to "Latest" (value is ""). Also add testing utils for `getFlag` usage. [Screencast](https://screencast.googleplex.com/cast/NTYyNDcxNjcwOTU5MzA4OHw0YWNhNTIwYy03Ng)

For some reason the calls to provision the runtime are taking very long today.

### DIFF
--- a/src/colab/client.ts
+++ b/src/colab/client.ts
@@ -126,7 +126,7 @@ export class ColabClient {
    *
    * @param notebookHash - Represents a web-safe base-64 encoded SHA256 digest.
    * This value should always be a string of length 44.
-   * @param params - The assignment parameters {@link AssignParams} 
+   * @param params - The assignment parameters {@link AssignParams}
    * like variant, accelerator, shape and version.
    * @param signal - Optional {@link AbortSignal} to cancel the request.
    * @returns The assignment which is assigned to the user.


src/colab/experiment-state.ts
```diff
@@ -62,6 +62,21 @@ export class ExperimentStateProvider extends AsyncToggle {
...
+): void {
+  const newFlags = new Map(flags);
+  newFlags.set(flag, value);
+  flags = newFlags;
+}
+
+/** Resets the experiment flags for testing. */
+export function resetFlagsForTest(): void {
+  flags = new Map();
+}
```
hjjackyang: Let's not directly export the functions for test only. Let's follow the Google best practice and export them via a `TEST_ONLY` const.

http://go/typescript-testing#export-visibility

```suggestion
function setFlagForTest(
  flag: ExperimentFlag,
  value: ExperimentFlagValue,
): void {
  const newFlags = new Map(flags);
  newFlags.set(flag, value);
  flags = newFlags;
}

/** Resets the experiment flags for testing. */
function resetFlagsForTest(): void {
  flags = new Map();
}

export const TEST_ONLY = {
  setFlagForTest,
  resetFlagsForTest,
};
```

src/colab/client.ts
```diff
@@ -131,13 +133,15 @@ export class ColabClient {
     variant: Variant,
     accelerator?: string,
     shape?: Shape,
+    version?: string,
```
hjjackyang: Now that we support so many params in this function call, it gets very verbose to pass them around. Could we consider extracting these runtime related parameters into a common interface?

---

I see we have a `ColabServerDescriptor` interface that is very similar but not exactly the same:
https://github.com/googlecolab/colab-vscode/blob/main/src/jupyter/servers.ts#L18-L23

src/colab/server-picker.ts
```diff
@@ -120,18 +130,30 @@ export class ServerPicker {
       );
       if (shapePicks.length <= 1) {
         state.shape = defaultShape;
+        if (versions.length === 0) {
+          state.version = ''; // Latest version
+          return (input: MultiStepInput) =>
+            this.promptForAlias(input, state, /** totalSteps= */ 2);
```
hjjackyang: `/** */` is for TS/JSDoc. In this case, just use regular block comment `/* */`.

```suggestion
            this.promptForAlias(input, state, /* totalSteps= */ 2);
```

See http://go/tsjs-style#comments-when-calling-a-function for more info.

src/colab/server-picker.ts
```diff
@@ -120,18 +130,30 @@ export class ServerPicker {
...
+        }
         return (input: MultiStepInput) =>
-          this.promptForAlias(input, state, /** totalSteps= */ 2);
+          this.promptForVersion(input, state, versions, /** step= */ 2);
```
hjjackyang:
```suggestion
          this.promptForVersion(input, state, versions, /* step= */ 2);
```

src/colab/server-picker.ts
```diff
@@ -120,18 +130,30 @@ export class ServerPicker {
...
       }
       return (input: MultiStepInput) =>
-        this.promptForMachineShape(input, state, shapePicks);
+        this.promptForMachineShape(
+          input,
+          state,
+          shapePicks,
+          versions,
+          /** step= */ 2,
```
hjjackyang:
```suggestion
          /* step= */ 2,
```

src/colab/server-picker.ts
```diff
@@ -168,20 +193,35 @@ export class ServerPicker {
     );
     if (shapePicks.length <= 1) {
       state.shape = defaultShape;
+      if (versions.length === 0) {
+        state.version = ''; // Latest version
+        return (input: MultiStepInput) =>
+          this.promptForAlias(input, state, /** totalSteps= */ 3);
```
hjjackyang:
```suggestion
          this.promptForAlias(input, state, /* totalSteps= */ 3);
```

src/colab/server-picker.ts
```diff
@@ -168,20 +193,35 @@ export class ServerPicker {
...
+      }
       return (input: MultiStepInput) =>
-        this.promptForAlias(input, state, /** totalSteps= */ 3);
+        this.promptForVersion(input, state, versions, /** step= */ 3);
```
hjjackyang:
```suggestion
        this.promptForVersion(input, state, versions, /* step= */ 3);
```

src/colab/server-picker.ts
```diff
@@ -168,20 +193,35 @@ export class ServerPicker {
...
     }
     return (input: MultiStepInput) =>
-      this.promptForMachineShape(input, state, shapePicks);
+      this.promptForMachineShape(
+        input,
+        state,
+        shapePicks,
+        versions,
+        /** step= */ 3,
```
hjjackyang:
```suggestion
        /* step= */ 3,
```

src/colab/server-picker.ts
```diff
@@ -194,6 +234,46 @@ export class ServerPicker {
     if (!isShapeDefined(state)) {
       return;
     }
+    if (versions.length === 0) {
+      state.version = ''; // Latest version
+      return (input: MultiStepInput) =>
+        this.promptForAlias(input, state, totalSteps);
+    }
+    return (input: MultiStepInput) =>
+      this.promptForVersion(input, state, versions, /** step= */ step + 1);
```
hjjackyang: `step + 1` is clear enough that this param is for `step`. No need to add the comment here.

```suggestion
      this.promptForVersion(input, state, versions, step + 1);
```

src/colab/server-picker.ts
mounikasomisetty: Done

src/colab/server-picker.ts
mounikasomisetty: Done

src/colab/server-picker.ts
mounikasomisetty: Done

src/colab/server-picker.ts
mounikasomisetty: Done

src/colab/server-picker.ts
mounikasomisetty: Done

src/colab/server-picker.ts
mounikasomisetty: Done

src/colab/server-picker.ts
mounikasomisetty: Done

src/colab/experiment-state.ts
mounikasomisetty: Great suggestion!

src/colab/client.ts
mounikasomisetty: Done, added `AssignParams`

src/colab/server-picker.ts
hjjackyang: This one too.

src/colab/client.ts
```diff
@@ -120,6 +128,8 @@ export class ColabClient {
    * This value should always be a string of length 44.
    * @param variant - The machine variant to assign.
    * @param accelerator - The accelerator to assign.
+   * @param shape - The machine shape to assign.
+   * @param version - The runtime version to assign.
```
hjjackyang: Nit: Please update TSDoc now that these params are in the interface.

PR #425, opened 2026-02-14
chore: migrate `refreshConnection` to new API

hjjackyang:
**Why**? Colab is in the process of deprecating old `/tun/m/*` APIs. This PR migrates `refreshConnection` to new `/v1/runtime-proxy-token` API.

**Local test**:
GetRuntimeProxyToken_test

---

Internal tracking bug: b/483447038

### DIFF
--- a/src/colab/api.ts
+++ b/src/colab/api.ts
@@ -261,12 +261,18 @@ export const RuntimeProxyTokenSchema = z
     /** URL of the runtime proxy. */
     url: z.string(),
   })
-  .transform(({ tokenTtl, ...rest }) => ({
-    ...rest,
+  .transform(({ tokenTtl, ...rest }) => {
     // Convert from string with 's' suffix to number of seconds and rename to
     // match `RuntimeProxyInfoSchema`.
-    tokenExpiresInSeconds: Number(tokenTtl.slice(0, -1)),
-  }));
+    const tokenExpiresInSeconds = Number(tokenTtl.slice(0, -1));
+    return {
+      ...rest,
+      tokenExpiresInSeconds:
+        Number.isNaN(tokenExpiresInSeconds) || tokenExpiresInSeconds <= 0
+          ? DEFAULT_TOKEN_TTL_SECONDS
+          : tokenExpiresInSeconds,
+    };
+  });
 export type RuntimeProxyToken = z.infer<typeof RuntimeProxyTokenSchema>;
 
 /** The response when creating an assignment. */
@@ -515,3 +521,5 @@ export const ExperimentStateSchema = z.object({
 });
 /** The experiment state response. */
 export type ExperimentState = z.infer<typeof ExperimentStateSchema>;
+
+const DEFAULT_TOKEN_TTL_SECONDS = 3600;
--- a/src/colab/client.unit.test.ts
+++ b/src/colab/client.unit.test.ts
@@ -528,40 +528,48 @@ describe('ColabClient', () => {
       };
     });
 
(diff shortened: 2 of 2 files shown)


src/colab/api.ts
```diff
@@ -251,7 +251,23 @@ export const RuntimeProxyInfoSchema = z.object({
...
+export const RuntimeProxyTokenSchema = z
+  .object({
+    /** Token for the runtime proxy. */
+    token: z.string(),
+    /** Token TTL, serialized from `google.protobuf.Duration` as string. */
+    tokenTtl: z.string(),
+    /** URL of the runtime proxy. */
+    url: z.string(),
+  })
+  .transform(({ tokenTtl, ...rest }) => ({
```
prodonjs: Gotta say I only learned about zod from looking at Kevin's earlier code. I love it and its so convenient for things like this. Do we need to worry about the Number conversion resulting in `NaN` and throwing if that's the case?

src/colab/client.ts
```diff
@@ -224,21 +224,14 @@ export class ColabClient {
   async refreshConnection(
     endpoint: string,
     signal?: AbortSignal,
-  ): Promise<RuntimeProxyInfo> {
-    const url = new URL(
-      `${TUN_ENDPOINT}/runtime-proxy-token`,
-      this.colabDomain,
-    );
+  ): Promise<RuntimeProxyToken> {
+    const url = new URL('v1/runtime-proxy-token', this.colabGapiDomain);
```
prodonjs, line 228: Probably not for here, but wondering if we want to pull some of these URLs into a factory class that makes it a bit easier to see which hosts/methods we're dependent on at a glance.

src/colab/client.unit.test.ts
```diff
@@ -529,33 +529,39 @@ describe('ColabClient', () => {
...
             queryParams: {
               endpoint: assignedServer.endpoint,
               port: '8080',
             },
+            withAuthUser: false,
           }),
         )
         .resolves(
-          new Response(withXSSI(JSON.stringify(newConnectionInfo)), {
+          new Response(withXSSI(JSON.stringify(rawRuntimeProxyToken)), {
```
prodonjs: Does the XSSI stuff still apply to the OnePlatform API endpoints? I thought they just returned raw, parseable JSON?

prodonjs (approved):
This looks good to me. Just a few comments that are mostly questions from my vantage point.

src/colab/api.ts
hjjackyang: I'm pretty certain we can trust the JSON format of `Duration` per http://go/protobuf-json#wkt. That being said, it's not a bad idea to handle the edge cases just in case.

I added the logic to handle `NaN` as well as negative duration.

src/colab/client.ts
hjjackyang: I can look into this in a separate refactoring PR. Thanks for the suggestion!

src/colab/client.unit.test.ts
hjjackyang: I'm actually not sure why this was added to every test as the tests pass whether with or without it. Perhaps @amtoney524 can shed some light?

For now, I'll just leave it here for consistency (since the `v1/user-info` test also has it).

amtoney524: yeah, I'm not sure about why that's in every test - we'll have to ask @kevineger once he's back!

hjjackyang: Thanks @amtoney524 !

PR #428, opened 2026-02-17
fix: update clearcut flush interval response type

amtoney524:
The Clearcut `LogResponse` proto specifies the type `int64` but its jspb type is `string`. This PR updates our parsing to handle the string response, as the original handling resulted in the fallback `MIN_WAIT_BETWEEN_FLUSHES` being used. 
It also updates the relevant unit test to catch this failure in the future.

### DIFF
--- a/project-words.txt
+++ b/project-words.txt
@@ -14,7 +14,6 @@ grayscale
 highmem
 ipykernel
 ipynb
-JSPB
 jupyterclient
 kernelspecs
 KHTML


src/telemetry/api.ts
```diff
@@ -69,5 +69,5 @@ export interface LogRequest {
 /** The Clearcut log response structure. */
 export interface LogResponse {
   // Minimum wait time before the next request in milliseconds.
```
mounikasomisetty: *nit* Should we call out the type discrepancy here with a comment?

amtoney524: good idea, done! thanks for the review!!

src/telemetry/client.ts
```diff
@@ -130,8 +130,9 @@ export class ClearcutClient implements Disposable {
 
     let next_flush_millis = MIN_WAIT_BETWEEN_FLUSHES_MS;
     try {
-      const { next_request_wait_millis: wait } =
+      const { next_request_wait_millis } =
         (await response.json()) as LogResponse;
+      const wait = parseInt(next_request_wait_millis, 10);
```
hjjackyang: Nit: Should we just simply type-coerce with `Number()` instead of `parseInt`? Since I think `parseInt` may run some extra logic which we don't really need (but either way is fine, no strong opinion here).

```suggestion
      const wait = Number(next_request_wait_millis);
```

src/telemetry/api.ts
```diff
@@ -68,6 +68,8 @@ export interface LogRequest {
 
 /** The Clearcut log response structure. */
 export interface LogResponse {
-  // Minimum wait time before the next request in milliseconds.
-  next_request_wait_millis: number;
+  // Minimum wait time before the next request in milliseconds. Note that the
+  // Clearcut LogResponse proto specifies the type int64 but its JSPB type is
```
hjjackyang: FYI: I don't think this is specific to JSPB. I think this is a canonical encoding in all protobufs when formatted into JSON.

http://go/protobuf-json#field-representation listed out how all proto types are represented in JSON.

http://go/protobuf-json#int64-strings explains why `int64` is serialized into `string`.

hjjackyang (approved):
LGTM with 2 nit comments!

src/telemetry/client.ts
amtoney524: simplified to `Number()`, thanks!

src/telemetry/api.ts
amtoney524: this is super helpful, thank you! will update the comment and keep this in mind for the future!

amtoney524:
thanks for the reviews!!

PR #429, opened 2026-02-17
feat: periodically refresh experiment state

mounikasomisetty:
Add logic to the experiment state provider to refresh the experiment state every **five minutes** starting after the first call to fetch experiment state. This could help with the telemetry flag's kill switch.

### DIFF
--- a/src/colab/experiment-state.ts
+++ b/src/colab/experiment-state.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import vscode from 'vscode';
+import { Disposable } from 'vscode';
 import { log } from '../common/logging';
 import { AsyncToggle } from '../common/toggleable';
 import {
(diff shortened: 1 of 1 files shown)


src/colab/experiment-state.ts
```diff
@@ -22,30 +23,60 @@ export function getFlag(flag: ExperimentFlag): ExperimentFlagValue {
...
+
   constructor(private readonly client: ColabClient) {
     super();
     // Reset experiment flags.
     flags = new Map<ExperimentFlag, ExperimentFlagValue>();
     this.getExperimentState = this.getExperimentState.bind(this);
   }
 
+  dispose() {
+    this.isDisposed = true;
```
hjjackyang, line 48: Nit: Add `if (this.isDisposed) { return; }`?

src/colab/experiment-state.ts
```diff
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import vscode from 'vscode';
```
hjjackyang: Nit: I would just `import { Disposable } from 'vscode'` here to minimize the import scope.

src/colab/experiment-state.ts
```diff
@@ -22,30 +23,60 @@ export function getFlag(flag: ExperimentFlag): ExperimentFlagValue {
   return flags.get(flag) ?? EXPERIMENT_FLAG_DEFAULT_VALUES[flag];
 }
 
+const REFRESH_INTERVAL_MS = 5 * 60 * 1000; // 10 minutes.
```
hjjackyang: Typo in comment:
```suggestion
const REFRESH_INTERVAL_MS = 5 * 60 * 1000; // 5 minutes.
```

src/colab/experiment-state.ts
```diff
@@ -22,30 +23,60 @@ export function getFlag(flag: ExperimentFlag): ExperimentFlagValue {
...
+  private ensurePolling() {
+    if (this.isDisposed || this.refreshInterval) {
+      return;
+    }
+    this.refreshInterval = setInterval(() => {
+      void this.getExperimentState(this.isAuthorized);
+    }, REFRESH_INTERVAL_MS);
   }
 
   private async getExperimentState(
```
hjjackyang, line 78: Now that you've made `ExperimentStateProvider` a `Disposable`, we should stop calling the backend in this method if `ExperimentStateProvider` is disposed.

Let's guard this method just in case.

See our existing example:
https://github.com/googlecolab/colab-vscode/blob/main/src/jupyter/contents/file-system.ts#L163

PR #432, opened 2026-02-19
chore: migrate `/ccu-info` to new `v1/user-info` API

hjjackyang:
**Why**? Colab is in the process of deprecating old `/tun/m/*` APIs. This PR migrates `getCcuInfo` (also renamed to `getConsumptionUserInfo`) from `/ccu-info` to new `v1/user-info` API with `get_ccu_consumption_info=true`.

[Demo screencast](https://screencast.googleplex.com/cast/NjYzNTA0NzI2NzMzNjE5Mnw2ZGU1NWMxNi03Nw) showing GPU/TPU selector still working and new API calls logged with status 200.

---

There are 2 optimizations as part of this migration:
1. `assignServer` used to call both `getSubscriptionTier` and `getCcuInfo` to retrieve server descriptors. After this migration, we can get everything we need to assign a server from the basic `getUserInfo` call, reducing 2 HTTP calls to one.
2. `ConsumptionNotifier` used to call `getSubscriptionTier` while it already has access to a `ConsumptionUserInfo` instance via event emitter. After this migration, we can directly get subscription tier from `ConsumptionUserInfo` without needing to call `getUserInfo` again, removing the redundant HTTP call.

---

Internal tracking bug: b/485350944

### DIFF
--- a/src/colab/api.ts
+++ b/src/colab/api.ts
@@ -149,16 +149,28 @@ export const UserInfoSchema = z.object({
   eligibleAccelerators: z.array(Accelerator),
   /** The ineligible machine accelerators. */
   ineligibleAccelerators: z.array(Accelerator),
+});
+/** Colab user information. */
+export type UserInfo = z.infer<typeof UserInfoSchema>;
+
+/**
+ * The schema for top level information about a user's tier, usage and
+ * availability in Colab when CCU consumption info is requested (consumption
+ * fields are required).
+ */
+export const ConsumptionUserInfoSchema = UserInfoSchema.required({
+  paidComputeUnitsBalance: true,
+}).extend({
   /**
    * The current rate of consumption of the user's CCUs (paid or free) based on
    * all assigned VMs.
    */
-  consumptionRateHourly: z.number().optional(),
+  consumptionRateHourly: z.number(),
   /**
    * The number of runtimes currently assigned when the user's paid CCU balance
    * is positive.
    */
-  assignmentsCount: z.number().optional(),
+  assignmentsCount: z.number(),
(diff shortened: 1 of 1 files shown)


hosungs (approved):
Oops, reviewed/approved as colaboratory-team, may not matter, but just in case, approving as hosungs too. :) Thanks again!

hjjackyang:
@hosungs - Thanks for the quick review!

> I believe this will be also soon used for integration tests (Kokoro?) and hit the real sandbox endpoints soon?

Actually, our E2E tests run as part of this PR pre-merge checks, and they hit real production endpoints I believe:
https://github.com/googlecolab/colab-vscode/actions/runs/22204531648/job/64225543737?pr=432

So rest assured!

src/colab/client.ts
```diff
@@ -92,31 +92,32 @@ export class ColabClient {
...
-   * Gets the user's subscription tier.
+   * Gets the current user information.
    *
    * @param signal - Optional {@link AbortSignal} to cancel the request.
-   * @returns The user's subscription tier.
    */
-  async getSubscriptionTier(signal?: AbortSignal): Promise<SubscriptionTier> {
-    const userInfo = await this.issueRequest(
+  async getUserInfo(signal?: AbortSignal): Promise<UserInfo> {
+    return await this.issueRequest(
```
amtoney524, line 100: super nit: drop `await` keyword if not needed (here and below)?

hjjackyang: I see that we've always had `await` keyword in async functions in `client.ts`. I'd rather keep them to be consistent if that's okay.

PR #434, opened 2026-02-24
chore: migrate `listAssignments` to new OnePlatform API

hjjackyang:
**Why**? Colab is in the process of deprecating old `/tun/m/*` APIs. This PR migrates`listAssignments` from `/tun/m/assignments` to new `/v1/assignments` API.

[Demo screencast](https://screencast.googleplex.com/cast/NjI4Mjk1OTAwMTc0NzQ1Nnw5Y2JkMDA2NS0zMg) showing new API in action.

---

Internal tracking bug: b/485355746

### DIFF
--- a/src/colab/client.unit.test.ts
+++ b/src/colab/client.unit.test.ts
@@ -508,78 +508,101 @@ describe('ColabClient', () => {
     });
   });
 
-  const tests = [
-    {
-      name: 'Unspecified',
+  it('successfully lists multiple assignments', async () => {
+    const mockAssignment1 = {
+      endpoint: 'm-s-foo-1',
+      accelerator: 'A100',
       variant: 'VARIANT_UNSPECIFIED',
-      expectedVariant: Variant.DEFAULT,
-      shape: 'SHAPE_UNSPECIFIED',
-      expectedShape: Shape.STANDARD,
-    },
-    {
-      name: 'GPU Default',
+      machineShape: 'SHAPE_UNSPECIFIED',
+      runtimeProxyInfo: {
+        token: 'new_token',
+        tokenTtl: '3600s',
+        url: 'https://8080-m-s-foo-1.bar.prod.colab.dev',
+      },
+    };
+    const mockAssignment2 = {
+      endpoint: 'm-s-foo-2',
+      accelerator: 'T4',
       variant: 'VARIANT_GPU',
-      expectedVariant: Variant.GPU,
-      shape: 'SHAPE_DEFAULT',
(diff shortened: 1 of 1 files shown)


src/colab/api.ts
```diff
@@ -307,25 +328,27 @@ export type PostAssignmentResponse = z.infer<
...
 });
 /** An abbreviated, listed assignment in Colab. */
 export type ListedAssignment = z.infer<typeof ListedAssignmentSchema>;
 
 /** The schema of the Colab API's list assignments endpoint. */
 export const ListedAssignmentsSchema = z.object({
-  assignments: z.array(ListedAssignmentSchema),
+  assignments: z
+    .array(ListedAssignmentSchema)
+    .optional()
```
kevineger, line 350: The array is optional? Mind double checking? If it is, we should likely file a bug to always return `[]` (RESTful).

src/colab/client.unit.test.ts
```diff
@@ -518,28 +508,95 @@ describe('ColabClient', () => {
     });
   });
 
-  it('successfully lists assignments', async () => {
+  const tests = [
```
kevineger: Does this really test any behavioural difference in the SUT? It just changes a few properties on the returned object. May be worthwhile but I'd argue tests covering empty, single and multiple may be more valuable.

kevineger (approved):

src/colab/api.ts
hjjackyang: Yes, I started with non-optional and it failed immediately upon local test because assignments was `undefined`.

I also just confirmed the behavior in RpcStudio when I had no assignments:
image

kevineger: Fine to keep our code as is, but let's file a bug.

src/colab/client.unit.test.ts
hjjackyang: Yes, these are meant to cover logic in `normalizeShape` and `normalizeVariant` unless we want to add an `api.unit.test.ts` to cover those.

That being said, I will modify this test case to cover multiple assignments, and bake these permutations into several different assignments to achieve the same coverage.

src/colab/api.ts
hjjackyang: FYI, I think this may be a proto serialization thing.

I just looked at the backend implementation. It always initializes an empty list and sets the `assignments` before returning at the end:
http://google3/research/colab/boq/oneplatform/runtime/service/runtimeservice.go;l=125;rcl=875271835

So I'm not sure if this is something that can be fixed by us. cc @hosungs

PR #435, opened 2026-02-24
fix: ensure telemetry is not logged on client disposal when setting disabled

amtoney524:
why? There existed an edge case where, if the user disabled telemetry in the middle of their session, any previously pending logs would get flushed to Clearcut on client disposal or upon retrying a failed request.

This change updates the client to also check the VS Code telemetry setting to ensure we don't flush to clearcut when the setting is disabled.

Note that we also check the setting [here](https://github.com/googlecolab/colab-vscode/blob/main/src/telemetry/index.ts#L82) so that we don't add logs to the queue when the user setting's disabled.

### DIFF
--- a/src/telemetry/client.ts
+++ b/src/telemetry/client.ts
@@ -109,7 +109,7 @@ export class ClearcutClient implements Disposable {
    * Flushes queued events to Clearcut.
    *
    * @param force - Flushes to Clearcut regardless of whether a flush is in
-   *   progress or if the flush interval's been met. Note that the VS Code 
+   *   progress or if the flush interval's been met. Note that the VS Code
    *   telemetry setting must still be enabled along with Colab's telemetry
    *   experiment flag.
    */


src/telemetry/client.ts
```diff
@@ -71,15 +74,20 @@ export class ClearcutClient implements Disposable {
...
+   *   still be enabled.
    */
   private async flush(force = false) {
-    const canLogToClearcut = getFlag(ExperimentFlag.EnableTelemetry);
-    const canFlush =
-      canLogToClearcut &&
-      (force || (!this.isDoingFlush && new Date() >= this.nextFlush));
+    // Must be enabled by Colab and the user before flushing to Clearcut.
+    const isTelemetryEnabled =
+      getFlag(ExperimentFlag.EnableTelemetry) && this.vs.env.isTelemetryEnabled;
```
hjjackyang, line 119: I would argue that any logs queued prior to user disabling telemetry should be flushed because those were technically already logged before user disabled logging. Thoughts?

amtoney524: I think it could go either way - it was logged technically before, but should we still flush it if the user's opted out? my vote would be no since that seems like it could be the safest way to operate in this grey area 

@kevineger wdyt?

src/telemetry/client.ts
```diff
@@ -71,15 +74,20 @@ export class ClearcutClient implements Disposable {
...
+   *   progress or if the flush interval's been met. Note that telemetry must
+   *   still be enabled.
    */
   private async flush(force = false) {
-    const canLogToClearcut = getFlag(ExperimentFlag.EnableTelemetry);
-    const canFlush =
-      canLogToClearcut &&
-      (force || (!this.isDoingFlush && new Date() >= this.nextFlush));
+    // Must be enabled by Colab and the user before flushing to Clearcut.
+    const isTelemetryEnabled =
```
kevineger, line 118: Right now in the telemetry module, we're only checking for `env.isTelemetryEnabled`:

https://github.com/googlecolab/colab-vscode/blob/9eaf019d9cede96d775eabf1d280b61ba5bc3da6/src/telemetry/index.ts#L48 

Should we instead add the flag check there and have a single point/accessor for whether telemetry is enabled?

Currently, the `telemetry/index` module is what guards whether `log` is called or not so it feels like the single responsibility is leaking when we check here in `flush`. Would be ideal if it could all be in here, or there, but not both.

amtoney524: I think we may want to consider keeping both checks in the client if we want them to be together -
* imagine that Colab has its telemetry flag set to false for a bit and then sets it to true. if we continue to queue events, there's a chance those logs might actually make it to Clearcut, minimizing the loss
* however, if the user opts out of the telemetry setting, then we shouldn't queue events at all. in that case, maybe `client.log` could early return if the settings not enabled and we could drop the check in `index.ts`

wdyt?

src/telemetry/client.ts
kevineger: We should drop them. It'd be weird to see logging traffic after disabling the setting, regardless of whether the events were from earlier.

src/telemetry/client.ts
kevineger: On second thought, it's not important that their either both in `index` or here in the `client`. Forgot that the intentions are the `index` default export of `log` for simple calling and the client is a private impl detail.

But we should also evaluate the flag in `index` no?

amtoney524: I don't think so, since we only want the flag to guard whether logs can actually be sent to Clearcut. If events are queued while the flags false, they still have a chance of getting flushed if the flag's toggled to true during the session.

If we were to check the flag in `index`, then it would likely be to prevent queuing of events right?

src/telemetry/client.ts
hjjackyang: SG!

src/telemetry/client.ts
hjjackyang: I think this discussion is out of scope of this PR. Though chance is low, I second Ashley's opinion and also prefer to keep rather than lose logs when possible.

src/telemetry/client.ts
```diff
@@ -71,15 +74,20 @@ export class ClearcutClient implements Disposable {
    * Flushes queued events to Clearcut.
    *
    * @param force - Flushes to Clearcut regardless of whether a flush is in
-   *   progress or if the flush interval's been met.
+   *   progress or if the flush interval's been met. Note that telemetry must
```
kevineger: Nit: Mind changing to `Note that the VS Code telemetry setting must ...` just since we have a couple knobs for this?

src/telemetry/client.ts
kevineger: I had understood the flag as our remote knob to fully disable logging. After talking with Ashley, looks like the intention is for it to just disable the call to Clearcut. Either is fine. Going with the former would allow us to safely and gradually roll out this functional change, but I'm fine with either!

amtoney524: thanks both for the discussion! going to resolve for now, but we can revisit in a future PR if we decide to go with an alternate solution

src/telemetry/client.ts
amtoney524: good idea, expanded the comment!

PR #436, opened 2026-02-24
fix: cosmetic and test improvements to telemetry

kevineger:
This change covers a couple cosmetic improvments to the telemetry module and some improvements to the test suite.

- Removes leak of unhandled promise rejections in `issueRequest`.
- Docstrings for better intellisense.
- Accept configuration instead of hardcoded defaults. More on this below w.r.t. test latency improvements.
- Removes `as unknown` cast.
- Minor test naming improvements.
- Reduces race condition between `client.log` an calling `sinon.assert` since we `void` invoke the promise. We need to wait for promises to settle.
- Re-uses common header definition.
- A couple tests were running slow (~1/2 a second). By having the client accept configuration, the smaller max limits reduce latency to ~50ms.

### DIFF
--- a/src/telemetry/client.unit.test.ts
+++ b/src/telemetry/client.unit.test.ts
@@ -124,11 +124,10 @@ describe('ClearcutClient', () => {
 
       client.log(DEFAULT_LOG);
       await firstFetchCalled.promise;
-      await fakeClock.tickAsync(10);
       sinon.assert.calledOnce(fetchStub);
       fetchStub.reset(); // Reset to clear behavior and history
 
-      await fakeClock.tickAsync(MIN_FLUSH_WAIT_MS + 10);
+      await fakeClock.tickAsync(MIN_FLUSH_WAIT_MS);
       const OTHER_LOG = {
         ...DEFAULT_LOG,
         timestamp: new Date(fakeClock.now).toISOString(),
@@ -155,11 +154,10 @@ describe('ClearcutClient', () => {
 
         client.log(DEFAULT_LOG);
         await firstFetchCalled.promise;
-        await fakeClock.tickAsync(10);
         sinon.assert.calledOnce(fetchStub);
         fetchStub.reset();
 
-        await fakeClock.tickAsync(MIN_FLUSH_WAIT_MS + 10);
+        await fakeClock.tickAsync(MIN_FLUSH_WAIT_MS);
         const OTHER_LOG = {
           ...DEFAULT_LOG,
           timestamp: new Date(fakeClock.now).toISOString(),
@@ -200,7 +198,7 @@ describe('ClearcutClient', () => {
         const failedLogs = createLogEvents(3);
         for (const [i, log] of failedLogs.entries()) {
           if (i === failedLogs.length - 1) {
-            await fakeClock.tickAsync(MIN_FLUSH_WAIT_MS + 10);
+            await fakeClock.tickAsync(MIN_FLUSH_WAIT_MS);
           }
           client.log(log);
         }
@@ -215,7 +213,7 @@ describe('ClearcutClient', () => {
 
         // Resolve the pending flush, requeuing failed logs
         pendingFlush.resolve();
-        await fakeClock.tickAsync(10);
+        await fakeClock.tickAsync(1);
         fetchStub.reset();
 
         const secondFetchCalled = new Deferred<void>();
@@ -237,7 +235,7 @@ describe('ClearcutClient', () => {
         const failedLogs = createLogEvents(MAX_PENDING_EVENTS);
         for (const [i, log] of failedLogs.entries()) {
           if (i === failedLogs.length - 1) {
-            await fakeClock.tickAsync(MIN_FLUSH_WAIT_MS + 10);
+            await fakeClock.tickAsync(MIN_FLUSH_WAIT_MS);
           }
           client.log(log);
         }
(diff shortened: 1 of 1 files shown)


src/telemetry/client.unit.test.ts
```diff
@@ -53,56 +59,55 @@ describe('ClearcutClient', () => {
...
+    it('flushes an event to Clearcut', async () => {
+      const fetchCalled = new Deferred<void>();
+      fetchStub.callsFake(() => {
+        fetchCalled.resolve();
+        return Promise.resolve(FETCH_RESPONSE_OK);
+      });
+
       client.log(DEFAULT_LOG);
 
+      await fetchCalled.promise;
```
amtoney524, line 71: fwiw, I had left this out because I was not observing any flakiness (e.g., fetch not being called in time for the assertion). That doesn't mean this couldn't become flaky in the future though - thanks for taking the time to add this in!

src/telemetry/client.unit.test.ts
```diff
@@ -111,49 +116,62 @@ describe('ClearcutClient', () => {
...
+      fetchStub.onFirstCall().callsFake(() => {
+        firstFetchCalled.resolve();
+        return Promise.resolve(FETCH_RESPONSE_400);
+      });
 
       client.log(DEFAULT_LOG);
-      sinon.assert.calledOnceWithExactly(fetchStub, logRequest([DEFAULT_LOG]));
-      fetchStub.reset();
+      await firstFetchCalled.promise;
+      await fakeClock.tickAsync(10);
```
amtoney524: nit: is this needed since we're ticking the clock below? (ditto where applicable)

src/telemetry/client.unit.test.ts
```diff
@@ -111,49 +116,62 @@ describe('ClearcutClient', () => {
...
+      sinon.assert.calledOnce(fetchStub);
+      fetchStub.reset(); // Reset to clear behavior and history
 
-      await fakeClock.tickAsync(TEST_ONLY.MIN_WAIT_BETWEEN_FLUSHES_MS);
+      await fakeClock.tickAsync(MIN_FLUSH_WAIT_MS + 10);
```
amtoney524: nit: same here, is the +10 needed? is it for extra padding? (ditto where applicable)

src/telemetry/client.unit.test.ts
```diff
@@ -111,49 +116,62 @@ describe('ClearcutClient', () => {
...
 
-    const requeueStatusTest = [{ status: 500 }, { status: 501 }];
-    for (const { status } of requeueStatusTest) {
-      it(`requeues events when Clearcut responds with a 500+ status: ${status.toString()}`, async () => {
-        fetchStub
-          .onCall(0)
-          .resolves(new Response('', { status }))
-          .onCall(1)
-          .resolves(FETCH_RESPONSE_OK);
+    const requeueStatuses = [500, 501];
```
amtoney524, line 146: nice :)

src/telemetry/client.unit.test.ts
```diff
@@ -192,10 +215,17 @@ describe('ClearcutClient', () => {
 
         // Resolve the pending flush, requeuing failed logs
         pendingFlush.resolve();
-        await fakeClock.tickAsync(1);
-        fetchStub.resetHistory();
+        await fakeClock.tickAsync(10);
+        fetchStub.reset();
```
amtoney524, line 217: (just curious) why reset over resetHistory?

src/telemetry/client.unit.test.ts
```diff
@@ -318,75 +377,110 @@ describe('ClearcutClient', () => {
         // Resolve the pending flush and advance time to reach the flush
         // interval.
         flushPending.resolve();
-        await fakeClock.tickAsync(LOG_RESPONSE_FLUSH_INTERVAL);
+        await fakeClock.tickAsync(10);
+        await fakeClock.tickAsync(LOG_RESPONSE_FLUSH_INTERVAL + 10);
```
amtoney524: do we need both ticks here?

src/telemetry/client.ts
```diff
@@ -18,33 +18,62 @@ import {
...
+/**
+ * Maximum number of pending events before flushing. When exceeded, events will
+ * be dropped from the front of the queue.
+ */
+const DEFAULT_MAX_PENDING_EVENTS = 1000;
+/** Minimum wait time between flushes in milliseconds. */
+const DEFAULT_MIN_FLUSH_WAIT_MS = 10 * 1000;
+
+/** The configuration for the Clearcut client. */
+export interface ClearcutConfig {
```
amtoney524, line 32: good idea! really cool to see the effect on test latency, will keep this in mind for the future!

src/telemetry/client.unit.test.ts
kevineger: The micro-task queue _should_ always resolve as you were seeing, but if in the future we do more async things here or in the downstream methods you'd start to see some flake.

src/telemetry/client.unit.test.ts
kevineger: I had `+someNumber` to show we're _moving past_ the flush period but it works without it. I'll simplify to just `tickAsync(MIN_FLUSH_WAIT_MS)` to avoid confusion with the magic numbers (that in this case aren't magic haha).

src/telemetry/client.unit.test.ts
kevineger: MB, nope. Removed.

src/telemetry/client.unit.test.ts
kevineger: `reset` will reset both behaviour and history of the stub. Enables us to more precicely setup `onFirstCall` and verify `calledOnceWithExactly`.

src/telemetry/client.unit.test.ts
kevineger: Nope, fixed!

PR #438, opened 2026-02-24
chore: test helper type safety

kevineger:
This change removes the `as unknown` cast in `as unknown as Partial<typeof vscode> as typeof vscode`. Yes, to accomplish this we need to do a lot of other type gymnastics, but it is worth it to have type safety in our test helpers. Without it, test authors can run into issues where a weird test setup doesn't match the real vscode API.

At first glance all the `"Method not implemented"` errors can look unnecessarily busy, it makes it much easier to debug when something is wrong or missing with the test setup.

### DIFF
--- a/src/colab/commands/notebook.ts
+++ b/src/colab/commands/notebook.ts
@@ -49,13 +49,13 @@ export async function notebookToolbar(
 }
 
 /**
- * Inserts a new code cell in the active notebook below the last selected cell.
+ * Appends a new code cell to the bottom of the active notebook.
  *
  * @param cellContent - Code content to add in the new cell.
  * @param languageId - Language of the code.
  * @returns `true` if cell is inserted successfully; `false` otherwise.
  */
-export async function insertCodeCellBelow(
+export async function appendCodeCell(
   vs: typeof vscode,
   cellContent: string,
   languageId: string,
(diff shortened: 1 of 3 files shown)


src/test/helpers/workspace.ts
```diff
@@ -4,15 +4,121 @@
...
+  set(
+    uri: Uri,
+    edits: readonly [NotebookEdit, WorkspaceEditEntryMetadata | undefined][],
+  ): void;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  set(uri: Uri, edits: readonly any[]): void {
+    this.uri = uri as TestUri;
+    this.edits = edits as TestNotebookEdit[];
+  }
+  /* eslint-enable @typescript-eslint/unified-signatures */
```
hjjackyang: I feel the unified signature override might look cleaner than overloads + eslint-disable comments. WDYT?

```suggestion
 set(
    uri: Uri,
    edits: readonly (
      | TextEdit
      | SnippetTextEdit
      | NotebookEdit
      | [TextEdit | SnippetTextEdit, WorkspaceEditEntryMetadata | undefined]
      | [NotebookEdit, WorkspaceEditEntryMetadata | undefined]
    )[],
  ): void {
    this.uri = uri as TestUri;
    this.edits = edits as TestNotebookEdit[];
  }
```

kevineger: Way cleaner! I didn't even bother trying since I thought TS would complain it needed to be directly the same when implementing the interface. TIL, thanks!

PR #441, opened 2026-02-25
test: Drive mounting e2e

hjjackyang:
**Why**? Adding e2e test to cover Drive mounting which was launched a while ago.

Local e2e test output:
```
Launching browser...
Browser ready in 1204 ms
Launching tests...
  Colab Extension
    with a notebook authenticated and connected to a Colab server
      ✔ executes basic code cells (10642ms)
      ✔ mounts Google Drive (25391ms)

Shutting down the browser

  2 passing (1m)

✅ All tests passed!
```
---

At the same time, refactored the e2e tests by moving the initial auth and server connection to a `before` block, treating it as a one-time test setup, and clearing the test notebook after each test.

### DIFF
--- a/src/test/extension.e2e.test.ts
+++ b/src/test/extension.e2e.test.ts
@@ -138,22 +138,20 @@ df`);
       focusedCell = await driver.switchTo().activeElement();
       await focusedCell.sendKeys('df.plot()');
 
-      // Execute the notebook and poll for the success indicator (green check).
-      // Why not the cell output? Because the output is rendered in a webview.
       await workbench.executeCommand('Notebook: Run All');
       // Collapsing all cell outputs so execution status of all 3 cells are in
       // the viewport.
       await workbench.executeCommand('Notebook: Collapse All Cell Outputs');
 
-      await assertCellExecutionSuccess(driver, workbench, /* numSuccess= */ 3);
+      await assertAllCellsExecutedSuccessfully(driver, workbench);
     });
 
     it('mounts Google Drive', async () => {
-      // Input code into the first cell.
+      // Delete the initial empty cell first because Mount Drive command will
+      // insert code snippet in a new cell.
+      await workbench.executeCommand('Notebook: Delete Cell');
       await workbench.executeCommand('Colab: Mount Google Drive to Server...');
 
-      // Execute the notebook and poll for the success indicator (green check).
-      // Why not the cell output? Because the output is rendered in a webview.
       await workbench.executeCommand('Notebook: Run All');
 
       await pushDialogButton({
@@ -178,7 +176,7 @@ df`);
         dialog: 'Please complete the authorization in your browser.',
       });
 
-      await assertCellExecutionSuccess(driver, workbench);
+      await assertAllCellsExecutedSuccessfully(driver, workbench);
     });
   });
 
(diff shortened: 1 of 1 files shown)


src/test/extension.e2e.test.ts
```diff
@@ -340,3 +381,27 @@ async function safeClick(
     errorMsg,
   );
 }
+
+async function assertCellExecutions(
+  driver: WebDriver,
+  workbench: Workbench,
+  expected: { numSuccess: number; numErrors: number },
```
kevineger: `numErrors` is always 0. What about `assertSucces` which takes in a 0-based cell index and validates that cell ran successfully?

src/test/extension.e2e.test.ts
```diff
@@ -50,38 +51,15 @@ describe('Colab Extension', function () {
     testTitle = this.currentTest?.fullTitle() ?? '';
   });
 
-  describe('with a notebook', () => {
-    beforeEach(async () => {
+  describe('with a notebook authenticated and connected to a Colab server', () => {
```
kevineger, line 54: Totally fine for now, but to get the brainstorming started... I'd love to get to a place where we do this more as a test fixtures / setup. That'll enable us to define e2e tests in separate files/modules but lean on this common setup.

src/test/extension.e2e.test.ts
```diff
@@ -263,9 +303,10 @@ df`);
 
       // Check that the test account's authenticated. Close the browser window.
       await oauthDriver.wait(
-        until.urlContains('vscode/auth-success'),
+        until.urlContains(expectedRedirectUrl),
         ELEMENT_WAIT_MS,
       );
+      await oauthDriver.close();
```
kevineger: Why is `close` needed in addition to `quit` below?

src/test/extension.e2e.test.ts
```diff
@@ -340,3 +381,27 @@ async function safeClick(
...
+      const errorElements = await container.findElements(
+        By.className('codicon-notebook-state-error'),
+      );
+      return (
+        successElements.length === expected.numSuccess &&
+        errorElements.length === expected.numErrors
+      );
+    },
+    CELL_EXECUTION_WAIT_MS,
+    'Notebook: Run All failed',
```
kevineger: With this assert / `wait` lifted to a helper, there's no guarantee the cells were ran under a `Run All`. Please update message accordingly (ideally to reflect success of the cell).

src/test/extension.e2e.test.ts
hjjackyang: Changed to `assertCellExecutionSuccess` as discussed offline!

src/test/extension.e2e.test.ts
hjjackyang: Right, this is not needed. I added initially when I was debugging another issue and forgot to remove it.

src/test/extension.e2e.test.ts
hjjackyang: Updated the error message.

src/test/extension.e2e.test.ts
```diff
@@ -382,10 +375,10 @@ async function safeClick(
   );
 }
 
-async function assertCellExecutions(
+async function assertCellExecutionSuccess(
```
kevineger: The two call-sites only ever verify all cells are successful right? First test checks all 3 cells, new Drive test checks the single cell it adds. Could simplify as `assertAllCellsExecutedSuccessfully`.

hjjackyang: Done!

To make this work, I needed to delete the initial empty cell in the notebook for my Drive mounting test because the Mount Drive command will insert a new cell, leaving an empty cell in the notebook without execution status.

PR #444, opened 2026-02-25
fix: remove stat call when reading a file

kevineger:
Per VS Code [docs](https://code.visualstudio.com/api/references/vscode-api#FileSystem), `readFile` is invoked to _read the entire contents of a file_ and won't be called for folders. Restores behaviour before https://github.com/googlecolab/colab-vscode/pull/417 while preserving the addition of `type` to ensure `.ipynb` files are read properly.

### DIFF
--- a/src/jupyter/contents/file-system.ts
+++ b/src/jupyter/contents/file-system.ts
@@ -255,12 +255,12 @@ export class ContentsFileSystemProvider
         type: ContentsGetTypeEnum.File,
       });
 
-      if (
-        isDirectoryContents(content) ||
-        content.format !== ContentsGetFormatEnum.Base64 ||
-        typeof content.content !== 'string'
-      ) {
-        throw this.vs.FileSystemError.FileIsADirectory(uri);
+      if (typeof content.content !== 'string') {
+        const err = new Error(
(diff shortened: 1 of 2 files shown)


src/jupyter/contents/file-system.ts
```diff
@@ -261,6 +256,7 @@ export class ContentsFileSystemProvider
       });
 
       if (
+        isDirectoryContents(content) ||
```
hjjackyang: Is this check necessary if we force `type: 'file'` now and this won't ever be called for a folder?

kevineger: Needed to discriminate the `Contents` vs `DirectoryContents` type union.

`DirectoryContents.content` is a `Contents[]` whereas `Contents.content` is a `Content`. We then pass that string to `Buffer.from` below.

hjjackyang: Won't `typeof content.content !== 'string'` help discriminate this already?

kevineger: Took a closer look. If we ever get non-string content back, that's a programming error. Updated code to reflect that and throw a standalone new error.

PR #448, opened 2026-02-26
chore: support and log telemetry events for `AutoConnectEvent`, `AssignServerEvent`, `PruneServersEvent` and `RemoveServerEvent`

mounikasomisetty:
+ Add event interfaces and log functions for `AutoConnectEvent`, `AssignServerEvent`, `PruneServersEvent` and `RemoveServerEvent`
+ Call log functions for these events only when the action is successful. 
+ Also added `EventSource` that can be reused for events that require a source.


src/jupyter/assignments.ts
```diff
@@ -427,6 +438,7 @@ export class AssignmentManager implements vscode.Disposable {
    */
   async unassignServer(
     server: ColabAssignedServer | UnownedServer,
+    source: EventSource,
```
mounikasomisetty: I can move this to `server.ts` in the `removeServer` function if we don't like passing this prop into the method but I thought it would be nice to have all the telemetry calls happen from this file OR I can set a default value here if needed.

hjjackyang (commented):
I wonder if we should log the events at the beginning of the events rather than after they succeed? Reason I'm saying that is if the events fail and we log an error event, we will then have the option to associate the error events with the corresponding triggering events.

kevineger:
> I wonder if we should log the events at the beginning of the events rather than after they succeed? Reason I'm saying that is if the events fail and we log an error event, we will then have the option to associate the error events with the corresponding triggering events.

+1 - I view these logs as tracking usage of the individual features. E.g. help us answer the question of "Do folks click auto-connect or always _New Colab Server_"?

mounikasomisetty:
> > I wonder if we should log the events at the beginning of the events rather than after they succeed? Reason I'm saying that is if the events fail and we log an error event, we will then have the option to associate the error events with the corresponding triggering events.
> 
> +1 - I view these logs as tracking usage of the individual features. E.g. help us answer the question of "Do folks click auto-connect or always _New Colab Server_"?

I understand, and I went back and forth on that myself. How does that work for things like `AssignServerEvent` where we include the server endpoint? Should we drop server endpoint in the event?

@kevineger @hjjackyang

hjjackyang:
> > > I wonder if we should log the events at the beginning of the events rather than after they succeed? Reason I'm saying that is if the events fail and we log an error event, we will then have the option to associate the error events with the corresponding triggering events.
> > 
> > 
> > +1 - I view these logs as tracking usage of the individual features. E.g. help us answer the question of "Do folks click auto-connect or always _New Colab Server_"?
> 
> I understand, and I went back and forth on that myself. How does that work for things like `AssignServerEvent` where we include the server endpoint? Should we drop server endpoint in the event?

That's a good question. Perhaps in this case, we can drop the server endpoint from `AssignServerEvent`. If we really want to log the server endpoint somewhere, we can log a separate `ServerAssignedEvent` or something after a server is assigned? Open to discussion.

mounikasomisetty:
> > > > I wonder if we should log the events at the beginning of the events rather than after they succeed? Reason I'm saying that is if the events fail and we log an error event, we will then have the option to associate the error events with the corresponding triggering events.
> > > 
> > > 
> > > +1 - I view these logs as tracking usage of the individual features. E.g. help us answer the question of "Do folks click auto-connect or always _New Colab Server_"?
> > 
> > 
> > I understand, and I went back and forth on that myself. How does that work for things like `AssignServerEvent` where we include the server endpoint? Should we drop server endpoint in the event?
> 
> That's a good question. Perhaps in this case, we can drop the server endpoint from `AssignServerEvent`. If we really want to log the server endpoint somewhere, we can log a separate `ServerAssignedEvent` or something after a server is assigned? Open to discussion.

Discussed offline - Made the changes!

src/jupyter/assignments.ts
```diff
@@ -561,7 +564,10 @@ export class AssignmentManager implements vscode.Disposable {
     );
     switch (selectedAction) {
       case AssignmentsExceededActions.REMOVE_SERVER:
-        this.vs.commands.executeCommand(REMOVE_SERVER.id);
+        this.vs.commands.executeCommand(
+          REMOVE_SERVER.id,
+          CommandSource.COMMAND_SOURCE_NOTIFICATION,
```
amtoney524, line 569: nit: should we add this to the proto too?

src/telemetry/api.ts
```diff
@@ -47,11 +47,42 @@ export type ColabEvent =
...
+  COMMAND_SOURCE_COLAB_TOOLBAR = 2,
+  COMMAND_SOURCE_COMMAND_PALETTE = 3,
+  COMMAND_SOURCE_NOTIFICATION = 4,
+}
+
 /** An event representing extension activation. */
 type ColabActivationEvent = Record<string, never>;
 
+/** An event representing a server auto connection */
+type AutoConnectEvent = Record<string, never>;
```
amtoney524: nit: prefix with Colab like the other events? (ditto below)

src/telemetry/api.ts
```diff
@@ -47,11 +47,42 @@ export type ColabEvent =
...
+
+/** An event representing a server assignment */
+type AssignServerEvent = Record<string, never>;
```
amtoney524: drop a todo to populate the fields in the event (e.g., variant, accelerator) or update the proto if we're no longer logging those?

same with `RemoveServerEvent`

src/telemetry/api.ts
```diff
@@ -47,11 +47,42 @@ export type ColabEvent =
   | {
       /** An event representing an error. */
       error_event: ColabErrorEvent;
+    }
+  | {
+      /** An event that logs when the user selects the autoconnect option */
+      auto_connect_event: AutoConnectEvent;
```
amtoney524: nit: keep these alphabetized?

src/jupyter/assignments.ts
mounikasomisetty: Yep! The CL adding it is out for review :)

src/telemetry/api.ts
mounikasomisetty: Yep! CL updating the proto is out for review

src/telemetry/api.ts
mounikasomisetty: done!

src/telemetry/api.ts
mounikasomisetty: done

Issue #442, opened 2026-02-25
Authentical to google drive of a google workplace account fails


kislay-quant:
Type: <b>Bug</b>

Hello ! I have been a VSC user for a long time now and would like to complement the team for such a great product. 

I use colab runtime through the colab extension. For machine learning projects, i need to acess my google drive throug drive.mount() function. It asks for authentication and a separate browser opens. The authentication works perfectly fine for the drived linked to my personal gmail account (limited storage capacity) but fails for the drive linked to my google workspace account (high storage capacity). Please look into it 

Extension version: 0.3.0
VS Code version: Code 1.109.5 (Universal) (072586267e68ece9a47aa43f8c108e0dcbf44622, 2026-02-19T19:43:32.382Z)
OS version: Darwin x64 21.6.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-5350U CPU @ 1.80GHz (4 x 1800)|
|GPU Status|2d_canvas: enabled<br>GPU0: VENDOR= 0x8086 [Google Inc. (Intel Inc.)], DEVICE=0x1626 [ANGLE (Intel Inc., Intel(R) HD Graphics 6000, OpenGL 4.1 INTEL-18.8.16)], DRIVER_VENDOR=INTEL, DRIVER_VERSION=18.8.16 *ACTIVE*<br>Machine model name: MacBookAir<br>Machine model version: 7.2<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|2, 2, 2|
|Memory (System)|8.00GB (0.20GB free)|
|Process Argv|. --crash-reporter-id 9cff6346-25a7-4a2b-a09f-5cd14c891ede|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
binariesv615:30325510
nativeloc1:31344060
dwcopilot:31170013
dwoutputs:31242946
copilot_t_ci:31333650
e5gg6876:31282496
[...]

PR #449, opened 2026-02-27
refactor: use `URL.searchParams.set` instead of `append`

hjjackyang:
**Why**? Kevin noticed a double `authuser=0&authuser=0` params in https://github.com/googlecolab/colab-vscode/issues/442#issuecomment-3964349035. This is not an issue, but may be confusing and suboptimal.

Using `URL.searchParams.set` instead of `append` to set the params will overwrite the previous value if the param already exists, or create it if it doesn't. See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/set) for more info.

This change is a refactoring and is functionally a no-op for all existing use cases.

---

Screenshot showing the double `authuser=0&authuser=0` params gone:
searchParamFix

### DIFF
--- a/src/colab/client.ts
+++ b/src/colab/client.ts
@@ -113,7 +113,7 @@ export class ColabClient {
     signal?: AbortSignal,
   ): Promise<ConsumptionUserInfo> {
     const url = new URL('v1/user-info', this.colabGapiDomain);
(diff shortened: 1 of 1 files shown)


kevineger (approved):
👍

PR #457, opened 2026-02-27
test: deflake e2e Drive auth timeout

hjjackyang:
**Why**? New `mounts Google Drive` e2e test can be flaky due to Drive auth either times out or fails transiently.

Examples:
* https://github.com/googlecolab/colab-vscode/actions/runs/22496147376/job/65171184839
  Colab Extension with a notebook authenticated and connected to a Colab server mounts Google Drive (oauth window)
* https://github.com/googlecolab/colab-vscode/actions/runs/22503452552/job/65196230352?pr=453
  Colab Extension with a notebook authenticated and connected to a Colab server mounts Google Drive (oauth window)

So I made `doOauthSignIn` resolves to a `boolean`, so I can run it in a loop until it succeeds or times out (60s).

---

While I was here, I also extracted 2 existing hardcoded sleep times to constants at the top.

### DIFF
--- a/scripts/test_e2e.sh
+++ b/scripts/test_e2e.sh
@@ -116,7 +116,7 @@ build_test_cmd() {
     base_cmd=(extest setup-and-run "${storage_flag[@]+"${storage_flag[@]}"}")
   fi
 
-  base_cmd+=("-o" "./out/test/e2e/settings.json" "-i")
+  base_cmd+=("-o" "./out/test/e2e/settings.json")
 
   # Print each part of the command on a new line.
   printf "%s\n" "${base_cmd[@]}"


src/test/extension.e2e.test.ts
kevineger, line 264: This change will result in the same screenshot file being replaced each attempt. Is that desirable? Also is there any sense what the Drive auth transient failure is caused by? Ideally that's something that works all the time. Is there an issue with redirection or something?

hjjackyang: > This change will result in the same screenshot file being replaced each attempt. Is that desirable?

I think this should be fine for the 2 scenarios:
1. If Drive OAuth was flaky, we probably don't care about previous failed screenshot(s) as long as the latest succeeds.
2. If Drive OAuth consistently fails, the latest failed screenshot should suffice for our debugging.

> Also is there any sense what the Drive auth transient failure is caused by? Ideally that's something that works all the time. Is there an issue with redirection or something?

I think this is tricky because I've only seen 2 flaky cases (attached in PR description) and both failed slightly differently:
1. One just stuck at email page and timed out there.
2. The other failed with an error, but I don't know if the OAuth error is logged anywhere. If there are logs, they are probably logged in DriveFS team's project since the OAuth client ID is created there.

#2 is slightly hard to debug at this point, but I can simplify the solution to reduce chance of #1 by just bumping up the wait time for password page to come up:
https://github.com/googlecolab/colab-vscode/blob/main/src/test/e2e/extension.e2e.test.ts#L210

WDYT?

hjjackyang: Sent https://github.com/googlecolab/colab-vscode/pull/459 for now to simply bump the timeout.

PR #463, opened 2026-03-03
feat: add command to import notebook from URL

mounikasomisetty:
Adds command (`importNotebookFromURL`) that allows users to download a notebook into their local file system by providing a Colab/Drive URL pointing to a Colab notebook and a place to save it.  [screencast](https://screencast.googleplex.com/cast/NjY0NDA0MDI5MTMxOTgwOHw4NTEzZGJmOS0wZA)

Two URL formats are supported:
+ `https://colab.research.google.com/drive/<ID>`
+ `https://drive.google.com/file/d/<ID>`

There will be follow ups to this to add (1) an e2e test, (2) deep linking.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,14 @@
           "tags": [
             "experimental"
           ]
+        },
+        "colab.importNotebookFromUrl": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enables the ability to import a notebook file from a URL.",
+          "tags": [
+            "experimental"
+          ]
         }
       }
     },
@@ -289,6 +297,10 @@
         {
           "command": "colab.openTerminal",
           "when": "config.colab.terminal"
+        },
+        {
+          "command": "colab.importNotebookFromUrl",
+          "when": "config.colab.importNotebookFromUrl"
         }
       ],
       "explorer/context": [


src/drive/commands/import.ts
```diff
@@ -0,0 +1,115 @@
...
+ * @param driveClient - The provider for interacting with Google Drive.
+ * @param inputUrl - An optional URL to import the notebook from. If not
+ * provided, the user will be prompted to enter one.
+ */
+export async function importNotebookFromUrl(
+  vs: typeof vscode,
+  driveClient: DriveClient,
+  inputUrl?: string,
+): Promise<void> {
+  inputUrl ??= await vs.window.showInputBox({
```
kevineger, line 25: We should throw some validation on this - [`validateInput`](https://code.visualstudio.com/api/references/vscode-api#InputBoxOptions). Looks like we have most of the nuts and bolts we'd need in `resolveRemoteSource` already!

src/drive/commands/import.ts
```diff
@@ -0,0 +1,115 @@
...
+    const doc = await vs.workspace.openNotebookDocument(targetUri);
+    await vs.window.showNotebookDocument(doc);
+
+    vs.window.showInformationMessage(
+      `Successfully saved to ${targetUri.fsPath}`,
+    );
+  } catch (e: unknown) {
+    log.error('Failed to import notebook:', e);
+    if (e instanceof Error) {
+      vs.window.showErrorMessage(`Failed to import notebook: ${e.message}`);
```
kevineger: Nit: use `const msg` for the base here and in the `log.error`.

src/drive/commands/import.unit.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      'Failed to import notebook: Network error',
+    );
+  });
+
+  it('handles unknown errors', async () => {
+    const url = 'https://colab.research.google.com/drive/123';
+    vsCodeStub.window.showInputBox.resolves(url);
+    driveClientStub.getDriveFileName.callsFake(() =>
+      // To simulate a non-error thrown
+      /* eslint-disable @typescript-eslint/prefer-promise-reject-errors */
```
kevineger: Consider just changing to `new Error('blah')` instead of the lint ignore.

src/extension.ts
```diff
@@ -161,6 +164,14 @@ async function activateInternal(context: vscode.ExtensionContext) {
     { treeDataProvider: serverResourceTreeView },
   );
 
+  const driveClient = DriveClient.create(
```
kevineger: Nit: can you group this next to `ColabClient` (also in the `context.subscriptions.push`)

package.json
```diff
@@ -182,6 +182,12 @@
         "enablement": "colab.hasAssignedServer",
         "title": "Open Terminal",
         "icon": "$(terminal)"
+      },
+      {
+        "category": "Colab",
+        "command": "colab.importNotebookFromUrl",
+        "title": "Import notebook file from URL",
+        "icon": "${arrow-down}"
```
kevineger: Should the icon be `$(...)`?

src/drive/commands/import.ts
```diff
@@ -0,0 +1,115 @@
...
+      );
+    }
+  }
+}
+
+function resolveRemoteSource(url: string): string {
+  const supportedFormats = [
+    {
+      // Format 1: Colab Notebook URL
+      regex: /\/drive\/([a-zA-Z0-9_-]+)/,
```
kevineger: Should we have more robust URL checking (e.g. verifying it's exactly the _host_(s) we support)? In other words, only Colab and Drive links work (instead of just matching the slim path + ID).

I think right now someone couls pass in `foo-bar-baz/drive/123`.

src/drive/commands/import.ts
```diff
@@ -0,0 +1,115 @@
...
+): Promise<void> {
+  inputUrl ??= await vs.window.showInputBox({
+    prompt: 'Link to the Colab Notebook to import',
+    placeHolder: 'https://colab.research.google.com/drive/...',
+  });
+
+  if (!inputUrl) return;
+
+  try {
+    const id = resolveRemoteSource(inputUrl);
```
kevineger, line 34: Should we also support the `/open?id=` links (or whatever the "share" format is)? Fine if not, just curious.

src/drive/commands/import.unit.test.ts
```diff
@@ -0,0 +1,157 @@
...
+        vsCodeStub.window.showInputBox.resolves(url);
+      }
+      driveClientStub.getDriveFileName.withArgs(fileId).resolves(fileName);
+      vsCodeStub.window.showSaveDialog.resolves(saveUri);
+      driveClientStub.getDriveFileContent
+        .withArgs(fileId)
+        .resolves(fileContent);
+      const doc = { uri: saveUri };
+      vsCodeStub.workspace.openNotebookDocument.resolves(
+        doc as unknown as NotebookDocument,
```
kevineger: Could we `as Partial` for some type safety?

src/drive/commands/import.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    const content = await driveClient.getDriveFileContent(id);
+    await vs.workspace.fs.writeFile(targetUri, content);
+
+    const doc = await vs.workspace.openNotebookDocument(targetUri);
+    await vs.window.showNotebookDocument(doc);
+
+    vs.window.showInformationMessage(
+      `Successfully saved to ${targetUri.fsPath}`,
+    );
```
kevineger: The editor will display the saved file (indicates success). Would it make sense to omit?

src/drive/commands/import.ts
```diff
@@ -0,0 +1,115 @@
...
+    prompt: 'Link to the Colab Notebook to import',
+    placeHolder: 'https://colab.research.google.com/drive/...',
+  });
+
+  if (!inputUrl) return;
+
+  try {
+    const id = resolveRemoteSource(inputUrl);
+    const fileName = await driveClient.getDriveFileName(id);
+    const targetUri = await getSaveLocation(vs, fileName);
```
hjjackyang, line 36: CUJ question - I'm wondering if we should upload the imported notebook to a Colab server instead of saving it locally?

How do we plan to support bi-directional updates in the future if saved locally?

package.json
mounikasomisetty: oh shoot, good catch.

package.json
hjjackyang, line 1: Consider adding an experimental config to gate this new feature?

hjjackyang (changes requested):

src/drive/commands/import.ts
mounikasomisetty: Sure

src/drive/commands/import.ts
mounikasomisetty: It's an older format for non-workspace files but there's no harm in added it, so I did!

src/drive/commands/import.ts
mounikasomisetty: I guess I was thinking it didn't matter because the drive call would just fail if the ID was invalid, but also fair enough. Added!

src/drive/commands/import.ts
mounikasomisetty: So the bi-directional updates would only be for future files, I don't think we can retroactively apply it to files that are imported locally. 

Eventually, we'll have a way for users to pick if they want to import locally or still link with drive. 

As for the former suggestion, I'm not against it, but maybe as a follow-up?

package.json
mounikasomisetty: Added! But I let it default to true since our goal is to add the command to Colab Web to invoke it.

src/drive/commands/import.ts
hjjackyang: > I don't think we can retroactively apply it to files that are imported locally.

Understand that. My question was more towards how the CUJ would look like for bi-directional support with local import. But I'm good with improving it further in follow-ups. Thanks!

PR #464, opened 2026-03-04
test: simplify signatures of ui clicks

kevineger:
Brevity adds a bit of clarity here. Source mappings, logs, screenshots and the fact that the quickpick items are unique should be enough to determine failures. Makes the tests a little easier to read and write! Also, and perhaps most importantly, reduces the chance a rename or mismatch goes unnoticed in the test code.


amtoney524 (approved):
nice!

PR #466, opened 2026-03-04
test: global e2e setup and split existing tests

kevineger:
Main change here is the introduction of a global `before` in `test-setup.ts` which'll run once, before all tests in the `e2e` suite. With that, we're able to easily split the e2e tests across files without worrying about the expensive, one-time auth flow.

If an error occurs during the global setup, all tests will fail and the relevant screenshot from VS and Chrome will be captured (where previously it wouldn't).


src/test/e2e/mount-drive.e2e.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  // Connect to Colab.
+  await workbench.executeCommand('Notebook: Select Notebook Kernel');
+  // If the test is running on a machine with a configured Python environment,
+  // the "Select Another Kernel" option may appear instead of "Colab". If so, we
+  // need to click it first before selecting "Colab".
+  if (await hasQuickPickItem(driver, 'Select Another Kernel')) {
+    await selectQuickPickItem(driver, 'Select Another Kernel');
+  }
+  await selectQuickPicksInOrder(driver, ['Colab', 'Auto Connect']);
```
hjjackyang, line 37: Nit: Shall we consider wrapping these lines in a helper function like `autoConnectToColab`? I foresee this function being useful in future e2e tests that don't care about which specific Colab server to connect to.

src/test/e2e/mount-drive.e2e.test.ts
hjjackyang, line 1: Name this file `mount-drive.e2e.test.ts` instead? Just in case we write e2e tests for different Drive related capabilities, e.g. Mounika's new Drive notebook integration.

src/test/e2e/auth.ts
hjjackyang, line 1: Nit: I think it's fine right now since we don't have too many e2e tests. But if our e2e test cases grow in the future, it'll be easier to differentiate tests vs test helpers if we group them in separate sub-folders.

hjjackyang (approved):
Thanks for the cleanup! Looks great!

Left a few minor comments for posterity.

src/test/e2e/mount-drive.e2e.test.ts
kevineger: Can do

src/test/e2e/mount-drive.e2e.test.ts
kevineger: We could, but maybe let's wait until there's a second test that does this? At the same time, seeing the individual steps does make the e2e flow more descriptive of what's going on.

[go/tott/643](https://goto.google.com/tott/643)

PR #467, opened 2026-03-04
feat: log sign-in attempts and sign-outs

kevineger:


src/telemetry/api.ts
```diff
@@ -103,6 +111,26 @@ interface ColabRemoveServerEvent {
...
+export enum AuthFlow {
+  AUTH_FLOW_UNSPECIFIED = 0,
+  /** The loopback authentication flow. */
+  AUTH_FLOW_LOOPBACK = 1,
+  /** The proxied redirect authentication flow. */
+  AUTH_FLOW_PROXIED_REDIRECT = 2,
+}
+
+/** An event representing a sign-in. */
+interface SignInEvent {
```
amtoney524: nit: should we prefix with `Colab` like the other events and keep them alphabetized? (ditto in `ColabEvent` above)

kevineger: Woops, missed we had that convention. I'll snap to it. But curious, why bother prefixing the type? Is it just to make the prospective proto migration easier?

PR #471, opened 2026-03-06
feat: log several command click events

hjjackyang:
Log the following 5 command events to Clearcut as part of Telemetry MVP:
* `ColabToolbarEvent`
* `MountDriveSnippetEvent`
* `MountServerEvent`
* `OpenColabWebEvent`
* `UpgradeToProEvent`

*Note: These `telemetry.log*` calls are still no-ops until telemetry module is officially launched.*

---

Internal doc reference: go/colab-vscode-telemetry
Internal tracking bug: b/487743354

### DIFF
--- a/src/telemetry/api.ts
+++ b/src/telemetry/api.ts
@@ -121,6 +121,17 @@ interface ColabErrorEvent {
   stack: string;
 }
 
+/** An event representing a click to insert Drive mounting snippet. */
+interface ColabMountDriveSnippetEvent {
+  source: CommandSource;
+}
+
+/** An event representing a Colab server mounting. */
+interface ColabMountServerEvent {
+  source: CommandSource;
+  server?: string;
+}
+
 /** An event representing server pruning */
 interface ColabPruneServersEvent {
   servers: string[];
@@ -154,24 +165,13 @@ type ColabSignOutEvent = Record<string, never>;
 /** An event representing a Colab toolbar click. */
 type ColabToolbarEvent = Record<string, never>;
 
-/** An event representing a click to insert Drive mounting snippet. */
-interface ColabMountDriveSnippetEvent {
+/** An event representing a click to upgrade to Colab Pro. */
+interface ColabUpgradeToProEvent {
   source: CommandSource;
-}
-
-/** An event representing a Colab server mounting. */
-interface ColabMountServerEvent {
-  source: CommandSource;
-  server: string;
 }
 
 /** An event representing a click to open Colab web. */
 interface OpenColabWebEvent {
-  source: CommandSource;
-}
-
-/** An event representing a click to upgrade to Colab Pro. */
-interface ColabUpgradeToProEvent {
   source: CommandSource;
 }
 
--- a/src/telemetry/index.ts
+++ b/src/telemetry/index.ts
@@ -63,6 +63,15 @@ export const telemetry = {
   logActivation: () => {
     log({ activation_event: {} });
   },
+  logAutoConnect: () => {
+    log({ auto_connect_event: {} });
+  },
+  logAssignServer: () => {
+    log({ assign_server_event: {} });
+  },
+  logColabToolbar: () => {
+    log({ colab_toolbar_event: {} });
+  },
   logError: (e: unknown) => {
     if (e instanceof Error) {
       log({
@@ -75,11 +84,14 @@ export const telemetry = {
       log({ error_event: { name: 'Error', msg, stack: '' } });
     }
   },
-  logAutoConnect: () => {
-    log({ auto_connect_event: {} });
+  logMountDriveSnippet: (source: CommandSource) => {
+    log({ mount_drive_snippet_event: { source } });
   },
-  logAssignServer: () => {
-    log({ assign_server_event: {} });
+  logMountServer: (source: CommandSource, server?: string) => {
+    log({ mount_server_event: { source, server } });
+  },
+  logOpenColabWeb: (source: CommandSource) => {
+    log({ open_colab_web_event: { source } });
   },
   logPruneServers: (servers: string[]) => {
     log({ prune_servers_event: { servers } });
@@ -96,18 +108,6 @@ export const telemetry = {
   },
   logSignOut: () => {
     log({ sign_out_event: {} });
-  },
-  logColabToolbar: () => {
-    log({ colab_toolbar_event: {} });
-  },
-  logMountDriveSnippet: (source: CommandSource) => {
-    log({ mount_drive_snippet_event: { source } });
-  },
-  logMountServer: (source: CommandSource, server?: string) => {
-    log({ mount_server_event: { source, server: server ?? '' } });
-  },
-  logOpenColabWeb: (source: CommandSource) => {
-    log({ open_colab_web_event: { source } });
   },
   logUpgradeToPro: (source: CommandSource) => {
     log({ upgrade_to_pro_event: { source } });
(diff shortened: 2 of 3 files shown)


src/colab/commands/server.ts
```diff
@@ -73,37 +73,44 @@ export async function mountServer(
...
+      ).value;
+      if (!selectedServer) {
+        return;
+      }
+      fs.mount(selectedServer);
+
+      return undefined;
+    });
+  } finally {
+    telemetry.logMountServer(source, selectedServer?.endpoint);
```
kevineger, line 112: Sorry if I'm reading this wrong, but won't this log when the user dismisses the select (`return undefined;`)? Shouldn't we check `if(selectedServer)`?

src/colab/commands/server.unit.test.ts
```diff
@@ -163,12 +165,15 @@ describe('Server Commands', () => {
   });
 
   describe('mountServer', () => {
+    const source = CommandSource.COMMAND_SOURCE_COMMAND_PALETTE;
     let assignmentManagerStub: SinonStubbedInstance<AssignmentManager>;
     let fsStub: SinonStubbedInstance<ContentsFileSystemProvider>;
+    let logStub: sinon.SinonStub;
```
kevineger: Nit: consider typing the stub.

src/colab/commands/server.unit.test.ts
```diff
@@ -177,10 +182,16 @@ describe('Server Commands', () => {
...
+      await mountServer(
+        vsCodeStub.asVsCode(),
+        assignmentManagerStub,
+        fsStub,
+        source,
+      );
 
       sinon.assert.notCalled(vsCodeStub.window.createQuickPick);
       sinon.assert.notCalled(fsStub.mount);
+      sinon.assert.calledOnceWithExactly(logStub, source, undefined);
```
kevineger, line 194: Curious on your thoughts here, how important do you think having logging assertions is?

I didn't bother adding them for sign-in/out (https://github.com/googlecolab/colab-vscode/pull/467), but can be convinced and circle back to adding them if you or @amtoney524 feel we should have them.

src/colab/commands/server.unit.test.ts
```diff
@@ -237,27 +264,35 @@ describe('Server Commands', () => {
...
           assignmentManagerStub,
           fsStub,
+          source,
         );
         await quickPickStub.nextShow();
         quickPickStub.onDidHide.yield();
 
         await expect(mount).to.eventually.be.fulfilled;
         sinon.assert.notCalled(fsStub.mount);
+        sinon.assert.calledOnceWithExactly(logStub, source, undefined);
```
kevineger, line 286: Related to comment in the `finally` of the SUT, here I'd expect no telemetry event.

src/telemetry/api.ts
```diff
@@ -131,6 +151,30 @@ interface ColabSignInEvent {
...
+
+/** An event representing a click to insert Drive mounting snippet. */
+interface ColabMountDriveSnippetEvent {
+  source: CommandSource;
+}
+
+/** An event representing a Colab server mounting. */
+interface ColabMountServerEvent {
+  source: CommandSource;
+  server: string;
```
kevineger: Pending answer to other comment(s) w.r.t. logging when the user cancels the flow, if we're keeping that logic this should probably be `?: string`.

src/colab/commands/server.ts
hjjackyang: Isn't our intention to always log when user triggers the "mount server" command whether they go through with the entire flow or not?

src/colab/commands/server.unit.test.ts
hjjackyang: Done!

src/colab/commands/server.unit.test.ts
hjjackyang: I think it can be a judgement call. For me, I only added the logging assertions for `mountServer` because it has some variations I'd like to test and make sure work.

But for some other simple scenarios, I didn't bother adding logging assertions.

src/telemetry/api.ts
hjjackyang: I can definitely make this `?: string`, but it doesn't change the [corresponding proto definition](http://google3/logs/proto/colab/vscode_event.proto;l=154;rcl=875783137).

One thing I wasn't sure about is how this will get serialized from JSON to proto on Clearcut side if I make this `?: string` and end up passing `{ server: undefined }` to Clearcut. I was afraid that it might get serialized into `server: "undefined"`, so I explicitly made it `: string` and make sure I set it as empty string when upstream passes `undefined`.

src/telemetry/index.ts
```diff
@@ -97,6 +97,21 @@ export const telemetry = {
   logSignOut: () => {
     log({ sign_out_event: {} });
   },
+  logColabToolbar: () => {
```
amtoney524: nit: worth keeping alphabetized?

src/telemetry/api.ts
```diff
@@ -131,6 +151,30 @@ interface ColabSignInEvent {
 /** An event representing a sign-out. */
 type ColabSignOutEvent = Record<string, never>;
 
+/** An event representing a Colab toolbar click. */
+type ColabToolbarEvent = Record<string, never>;
```
amtoney524, line 166: nit: keep alphabetized? (ditto above)

src/colab/commands/server.unit.test.ts
amtoney524: having it be a judgement call also sgtm!

src/telemetry/api.ts
kevineger: All fields are implicitly optional for proto so `undefined` should work fine. No biggy either way.

amtoney524: I'm good with passing an empty string or making the server field optional proto-side if we're keeping the logging as is.

also, might be worth adding a call to `initializeTelemetry` and then sending over an event with an `undefined` server to see what happens (once an event is logged, it takes about 10 min to show up in the query)

src/colab/commands/server.ts
kevineger: For this, and other multi-step flows, it'd be great to be able to disambiguate between "flows started" and "flows finished". Would help us know if there's a drop-off rate, or some problem with the flow. But I think for this one in particular that delta isn't interesting, your call!

kevineger (approved):

src/colab/commands/server.ts
hjjackyang: Thanks! In this case, I'll keep as is for now. We can always revisit and refactor after telemetry is launched.

src/telemetry/api.ts
hjjackyang: > All fields are implicitly optional for proto so `undefined` should work fine. No biggy either way.

Thanks! I changed the interface to `?: string` and will do a local test later to confirm if this works as expected.

src/telemetry/index.ts
hjjackyang: Done!

src/telemetry/api.ts
hjjackyang: Done. But one thing to note is that our types/interfaces here are all prefixed with `Colab-`, which makes it a bit awkward for event protos already contain word `Colab` in them, e.g. `ColabToolbarEvent` and `OpenColabWebEvent`.

For now, I prefixed all other events and alphabetized them, but I'm considering to remove these unnecessary prefixes in a separate PR later so they are aligned with their corresponding proto messages.

Let me know if you think otherwise.

PR #472, opened 2026-03-06
feat: log uncaught activation errors

amtoney524:
Adds a call to telemetry to log uncaught errors during extension activation.

To support this, a wrapper function is introduced to decorate a given function with error logging behaviors.

Note that this is currently a noop as a `intializeTelemetry` is not being called.

https://github.com/user-attachments/assets/4b651c0e-f87c-40bb-8c0d-8c137a16b7e1

### DIFF
--- a/src/telemetry/wrappers.ts
+++ b/src/telemetry/wrappers.ts
@@ -29,7 +29,7 @@ export function withErrorTracking<
     if (isPromiseLike(result)) {
       return Promise.resolve(result).catch((error: unknown) => {
         telemetry.logError(error);
-        return Promise.reject(error as Error);
+        throw error;
       }) as ReturnType<T>;
     }
 
--- a/src/telemetry/wrappers.unit.test.ts
+++ b/src/telemetry/wrappers.unit.test.ts
@@ -27,7 +27,7 @@ describe('withErrorTracking', () => {
       const error = new Error('error');
       const fn = async () => {
         await fnCalled.promise;
-        throw error;
+        return Promise.reject(error);
       };
 
       const result = withErrorTracking(fn)();


src/extension.ts
```diff
@@ -56,6 +56,14 @@ import { CommandSource } from './telemetry/api';
 
 // Called when the extension is activated.
 export async function activate(context: vscode.ExtensionContext) {
+  try {
+    await activateInternal(context);
+  } catch (err) {
+    telemetry.logError(err);
```
kevineger: Should we not still throw the `err` so VS Code can correctly report that the activation failed to activate due to it? Otherwise I suspect it'll either wait for a timeout or not give any reason/indication?

amtoney524: yes! nice catch, thank you!

src/extension.ts
```diff
@@ -56,6 +56,15 @@ import { CommandSource } from './telemetry/api';
 
 // Called when the extension is activated.
 export async function activate(context: vscode.ExtensionContext) {
+  try {
```
kevineger: This is totally fine, but wanted to share one option we might consider (now, or in the future).

A Typescript-idiomatic way of doing this may be with a decorator:

```ts
@TrackError
static async activate(context: vscode.ExtensionContext) {
  // ... logic here
}
```

Pros:

- Clean and intent is clear.
- Highly reusable across multiple methods
- Keeps logic contained and consistent to a single function

Cons:

- Not as explicit.
- Decorators aren't super broadly understood and can look _weird_ at first if you're not used to them (personally, I'm a big fan and have used them for our existing logging module).

amtoney524: thanks a lot for the suggestion! I agree that a decorator would be useful here, but I don't think we can use them for functions not included in a class. To offer similar functionality, I created a wrapper function that handles the error logging - lmk what you think!

kevineger: Oh woops, you're spot on about classes/methods and not functions. My bad.

mounikasomisetty:
So clever!

src/telemetry/wrappers.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @returns A new function that wraps the original function with error tracking.
+ */
+export function withErrorTracking<
+  T extends (...args: Parameters<T>) => ReturnType<T>,
+>(fn: T): T {
+  return function (this: unknown, ...args: Parameters<T>): ReturnType<T> {
+    let result: ReturnType<T>;
+    try {
+      result = fn.apply(this, args);
+    } catch (error) {
```
hjjackyang: Nit: Explicitly catch this error as `: unknown` as well?

```suggestion
    } catch (error: unknown) {
```

src/telemetry/wrappers.unit.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    sinon.restore();
+  });
+
+  describe('on an async function', () => {
+    it('logs and rethrows errors', async () => {
+      const fnCalled = new Deferred<void>();
+      const error = new Error('error');
+      const fn = async () => {
+        await fnCalled.promise;
+        throw error;
```
hjjackyang: Nit: To better simulate an async call, should we do `Promise.reject` instead of `throw` here?

src/telemetry/wrappers.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
hjjackyang:
```suggestion
 * Copyright 2026 Google LLC
```

src/telemetry/wrappers.unit.test.ts
```diff
@@ -0,0 +1,77 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
hjjackyang:
```suggestion
 * Copyright 2026 Google LLC
```

src/telemetry/wrappers.ts
amtoney524: done!

src/telemetry/wrappers.unit.test.ts
amtoney524: done!

src/telemetry/wrappers.ts
amtoney524: thanks for the catch :)

src/telemetry/wrappers.unit.test.ts
amtoney524: ditto, thanks!

amtoney524:
thanks all for the detailed reviews!

PR #473, opened 2026-03-06
refactor: drop redundant `Colab-` prefixes

hjjackyang:
**Why**?
1. The `Colab-` prefix is quite redundant and does not add additional context to these internal event types / interfaces.
2. Dropping the prefix can align the naming of these TS types / interfaces with their corresponding proto messages in [`VscodeEvent`](https://source.corp.google.com/piper///depot/google3/logs/proto/colab/vscode_event.proto) proto.


amtoney524 (approved):
lgtm, thank you!!

PR #477, opened 2026-03-09
feat: add ability to store multiple auth sessions in AuthStorage

mounikasomisetty:
Refactor `AuthStorage` to support multiple auth sessions as a precursor to drive support

### DIFF
--- a/src/auth/storage.unit.test.ts
+++ b/src/auth/storage.unit.test.ts
@@ -85,9 +85,7 @@ describe('AuthStorage', () => {
     it('finds a session that satisfies requested scopes', async () => {
       secretsStub.get.resolves(JSON.stringify([SESSION_1, SESSION_2]));
 
-      const found = await authStorage.getSession([
-        'https://www.googleapis.com/auth/drive.readonly',
-      ]);
+      const found = await authStorage.getSession(['corgi']);
 
       expect(found?.id).to.equal(SESSION_2.id);
     });


src/auth/storage.ts
```diff
@@ -6,52 +6,97 @@
...
     }
-    const sessions = parseAuthenticationSessions(sessionJson);
-    if (sessions.length != 1) {
-      throw new Error(
-        `Unexpected number of sessions: ${sessions.length.toString()}`,
-      );
+
+    try {
+      this.cachedSessions = parseAuthenticationSessions(sessionJson);
+    } catch (err) {
```
kevineger: The only time this should error is if we have a programming error (there's something stored which doesn't conform to our schema). In that case, are we sure we want to silently handle the error by returning `[]` and leaving the corrupt data persisted?

I haven't thought about this too deply, but it feels like we should either `this.secrets.delete(SESSIONS_KEY)` or throw the error, as is done today. I think I lean towards the latter.

src/auth/storage.ts
```diff
@@ -6,52 +6,97 @@
...
+   * @returns The refreshable authentication session, if it exists. Otherwise,
+   * `undefined`.
+   */
+  async getSession(
+    scopes: readonly string[] | string[],
+  ): Promise<RefreshableAuthenticationSession | undefined> {
+    const sessions = await this.getSessions();
+    return sessions.find((session) =>
+      scopes.every((scope) => session.scopes.includes(scope)),
+    );
```
kevineger, line 62: The persisted sessions could have scopes in addition to the `scopes` provided. That coupled with `find` returning the _first match_ may be problematic. Consider the case where we have sessions with scopes `[a, b]` and `[a]` stored. If the caller invokes `getSession([a])` we'd return the session with `[a, b]`, effectively elevating their permission with the returned session. That and it's not clear what the API _should_ return.

I think we should do the following, either:

1. Change `getSession` to `getSessionsWithScopes` which returns an array of them.
1. Change `getSession` to match _exactly_ the provided scopes.

Deciding which largely depends on how you intend to consume this downstream in the `GoogleAuthProvider`. Since I suspect this is being added to back the `getSessions` call, option (1) probably makes the most sense?

src/auth/storage.ts
```diff
@@ -6,52 +6,97 @@
...
+  }
+
+  /**
+   * Retrieves a specific session by its unique ID.
+   *
+   * @param sessionId - The session ID.
+   * @returns The refreshable authentication session, if it exists. Otherwise,
+   * `undefined`.
+   */
+  async getSessionById(
```
kevineger: Do you have a draft PR or branch I could peek at where this is called? Curious what the use case is.

src/auth/storage.ts
```diff
@@ -6,52 +6,97 @@
...
+    sessionId: string,
+  ): Promise<RefreshableAuthenticationSession | undefined> {
+    const sessions = await this.getSessions();
+    return sessions.find((session) => session.id === sessionId);
+  }
+
+  /**
+   * Stores a session, replacing any existing session with the same ID.
```
kevineger: Nit: Since this is the only mutable entry-point to the sessions, we can change the verbiage here to indicate it will replace the single session with the same ID if it exists.

```suggestion
   * Stores a session, replacing the session with a matching ID if it has already been stored.
```

src/auth/storage.ts
```diff
@@ -60,15 +105,31 @@ export class AuthStorage {
...
+      await this.secrets.delete(SESSIONS_KEY);
+    }
+    return sessionToRemove;
+  }
+
+  /**
+   * Internal helper to commit the current cache to the secret store.
+   */
+  private async save(): Promise<void> {
+    if (this.cachedSessions) {
```
kevineger: Sorry I'm not totally following why we'd only `store` if `chachedSessions` is truthy. Looking at the logic in combination with the two call sites, it looks right, it's just a bit hard to trace since we've split up the places we touch `cachedSessions`.

---

Would you consider completely removing the caching? We don't have it today and it's unlikely to be needed. When we consider the total time spent making authenticated calls (that need these sessions), reading the secrets file from disk is small potatoes compared to the round trip network time. Are we reading from disk unecessary and could we front it with an in-mem cache? Absolutely. But I think for this PR just the migration from a single session to multiple would be best.

src/auth/storage.ts
```diff
@@ -92,3 +153,7 @@ function parseAuthenticationSessions(
 
   return z.array(RefreshableAuthenticationSessionSchema).parse(sessions);
 }
+
+export const TEST_ONLY = {
```
kevineger: I'm averse to having _test_ code in prod code. I'd prefer if we keep `SESSIONS_KEY` as an implementation detail to the module and instead focus on testing public behaviour. In other words, the test (bodies, anyways) shouldn't even _know_ that we're using VS Code's secret store. All assertions we need to make should be supported by the public API of the SUT.

src/auth/storage.ts
mounikasomisetty: Changed to the latter, I have no preference here.

src/auth/storage.ts
mounikasomisetty: Ah I do not have a branch as the changes are stashed but I have them copied to a [paste](https://paste.googleplex.com/4742205773316096) 

This will be called from `removeSession` to get the stored session, to then get access to the refresh token. Since there will be multiple sessions, the credentials will need to be set on the `oAuth2` client before revoking them BUT since there should be only one session per scope set (explained in the other comment response), I could change this to get by scope, but I thought this was an easy add.

src/auth/storage.ts
mounikasomisetty: Done

src/auth/storage.ts
mounikasomisetty: Caching has been removed!

src/auth/storage.ts
mounikasomisetty: Ah makes sense, I was thinking something looked off. 

Changed to option #2 because there should only be one session per scope set.

src/auth/storage.ts
```diff
@@ -6,52 +6,90 @@
...
-      throw new Error(
-        `Unexpected number of sessions: ${sessions.length.toString()}`,
-      );
+
+    try {
+      return parseAuthenticationSessions(sessionJson);
+    } catch (err: unknown) {
+      // e.g. if storage is corrupted
+      log.error(`Failed to parse stored authentication sessions:`, err);
+      throw new Error(`Failed to parse stored authentication sessions`);
```
kevineger: Could we `throw err` to preserve the underlying error? Or at least include it's `message` if `err` is an `Error`?

src/auth/storage.ts
```diff
@@ -6,52 +6,90 @@
...
+   * `undefined`.
+   */
+  async getSession(
+    scopes: readonly string[],
+  ): Promise<RefreshableAuthenticationSession | undefined> {
+    const sessions = await this.getSessions();
+    return sessions.find(
+      (session) =>
+        scopes.every((scope) => session.scopes.includes(scope)) &&
+        scopes.length === session.scopes.length,
```
kevineger: Uber optional nit: check length first to save iterating the `scopes`. In practice this really doesn't matter, totally optional.

src/auth/storage.ts
```diff
@@ -60,15 +98,28 @@ export class AuthStorage {
...
+    const sessionToRemove = await this.getSessionById(sessionId);
+    if (!sessionToRemove) {
       return undefined;
     }
-    if (session.id !== sessionId) {
-      return undefined;
+
+    const updatedSessions = sessions.filter((s) => s.id !== sessionId);
+    await this.save(updatedSessions);
+    return sessionToRemove;
```
kevineger, line 101: - Calling `this.getSessions`, then `this.getSessionsById` ends up being two calls to read from storage.
- Calling `this.getSessionById` and then the `filter` to remove iterates twice.

Could we change this body to just something like:

```suggestion
    const sessions = await this.getSessions();
    let toRemove: RefreshableAuthenticationSession | undefined;
    const toKeep: RefreshableAuthenticationSession[] = [];
    for (const s of sessions) {
        if (s.id === sessionId) {
            toRemove = s;
        } else {
            toKeep.push(s);
        }
    }
    if (toRemove) {
        await this.save(toKeep);
    }

    return toRemove;
```

src/auth/storage.ts
kevineger: May not need it if we inline the body of `removeSession` to avoid the double class/iterations.

kevineger (changes requested):
Thanks for working through feedback, sorry just a couple last things/questions.

src/auth/storage.ts
mounikasomisetty: Done

src/auth/storage.ts
mounikasomisetty: Done

src/auth/storage.ts
mounikasomisetty: Ah I meant `GoogleAuthProvider.removeSession` but I can remove this for now and add it when it's used

src/auth/storage.ts
```diff
@@ -6,52 +6,76 @@
...
     }
-    return sessions[0];
   }
 
   /**
-   * Stores the refreshable authentication session.
+   * Retrieves a session that matches the requested scopes exactly.
+   *
+   * @param scopes - An array of scopes. The returned session will have all of
+   * these scopes included in its permissions.
```
kevineger: Nit

```suggestion
   * @param scopes - An array of scopes. The returned session will have these
   * exact scopes.
```

src/auth/storage.ts
```diff
@@ -6,52 +6,76 @@
...
+    return sessions.find(
+      (session) =>
+        scopes.length === session.scopes.length &&
+        scopes.every((scope) => session.scopes.includes(scope)),
+    );
+  }
+
+  /**
+   * Stores a session, replacing the session with a matching ID
+   * if it has already been stored
```
kevineger: Nit: full sentence.

```suggestion
   * if it has already been stored.
```

src/auth/storage.ts
```diff
@@ -60,15 +84,33 @@ export class AuthStorage {
...
+    if (sessionToRemove) {
+      await this.save(sessionsToKeep);
     }
-    if (session.id !== sessionId) {
-      return undefined;
+    return sessionToRemove;
+  }
+
+  /**
+   * Internal helper to commit the current cache to the secret store.
```
kevineger: Update comment now that we aren't caching.

src/auth/storage.unit.test.ts
```diff
@@ -12,17 +12,21 @@ import { SecretStorageFake } from '../test/helpers/secret-storage';
...
   refreshToken: '//42',
   account: { id: 'foo', label: 'bar' },
   scopes: ['baz'],
 };
+const SESSION_2: RefreshableAuthenticationSession = {
+  id: '2',
+  refreshToken: 'token-2',
+  account: { id: 'user1@gmail.com', label: 'User One' },
+  scopes: ['https://www.googleapis.com/auth/drive.readonly'],
+};
```
kevineger, line 26: Totally optional nit: consider metasyntactic variables to be concise and consistent with existing test data.

```suggestion
const SESSION_2: RefreshableAuthenticationSession = {
  id: '2',
  refreshToken: '\\43',
  account: { id: 'qux', label: 'quux },
  scopes: ['corge'],
};
```

src/auth/storage.unit.test.ts
```diff
@@ -37,100 +41,156 @@ describe('ServerStorage', () => {
...
+
+      const sessions = await authStorage.getSessions();
 
+      expect(sessions).to.deep.equal([]);
       sinon.assert.calledOnceWithExactly(secretsStub.get, SESSIONS_KEY);
     });
 
-    it('returns a single session when one is stored', async () => {
-      await authStorage.storeSession(DEFAULT_SESSION);
+    it('returns stored sessions', async () => {
```
kevineger: Optional: We've got a test case for corupt data (two), empty and multiple. Consider adding one for a single session. Also consider grouping the "empty" cases together at the top, then the single, then multiple.

src/auth/storage.unit.test.ts
```diff
@@ -37,100 +41,156 @@ describe('ServerStorage', () => {
...
+    });
+
+    it('returns an empty array if schema validation fails', async () => {
+      secretsStub.get.resolves(JSON.stringify([{ id: 'missing-fields' }]));
+
+      await expect(authStorage.getSessions()).to.be.rejected;
+    });
+  });
+
+  describe('getSession (by scopes)', () => {
```
kevineger: There's no other `getSession`.

```suggestion
  describe('getSession', () => {
```

src/auth/storage.unit.test.ts
```diff
@@ -37,100 +41,156 @@ describe('ServerStorage', () => {
...
+      const found = await authStorage.getSession(SESSION_1.scopes);
+
+      expect(found).to.deep.equal(SESSION_1);
     });
   });
 
   describe('storeSession', () => {
-    it('stores the provided session', async () => {
-      await authStorage.storeSession(DEFAULT_SESSION);
+    it('stores a single session', async () => {
```
kevineger: Nit: the API only lets you pass a single `session`. Recommend reverting to avoid confusion that there's a way to store multiple sessions.

Or maybe `stores a session when none are stored` to disambiguate with the following test case.

src/auth/storage.unit.test.ts
```diff
@@ -37,100 +41,156 @@ describe('ServerStorage', () => {
...
         SESSIONS_KEY,
-        JSON.stringify([DEFAULT_SESSION]),
+        JSON.stringify([SESSION_1]),
       );
     });
 
-    it('overrides the existing stored session', async () => {
-      const newSession = { ...DEFAULT_SESSION, id: '2' };
-      await authStorage.storeSession(DEFAULT_SESSION);
+    it('adds a new session when it has a non-existent ID', async () => {
```
kevineger: Nit: clarity with what exactly is being tested. At first class "when it has a non-existent ID" might mean we're trying to store a session which doesn't have ID set (which isn't possible given schema validation).

```suggestion
    it('stores a session when there is a different one already persisted', async () => {
```

src/auth/storage.unit.test.ts
```diff
@@ -37,100 +41,156 @@ describe('ServerStorage', () => {
...
+      const updatedSession = { ...SESSION_1, refreshToken: 'new-token' };
+
+      await authStorage.storeSession(updatedSession);
+
+      const sessions = await authStorage.getSessions();
+      expect(sessions.length).to.equal(1);
+      expect(sessions[0].refreshToken).to.equal('new-token');
+    });
+
+    it('updates secrets after storing', async () => {
```
kevineger: I think we can remove this test case as it may have been left behind from when we were using the secret key directly? The preceding test cases do a great job covering the behaviour of the SUT. A caller should not have to know/care that the secret stub is used to back the impl.

PR #482, opened 2026-03-10
feat: wrap registered commands with error tracking

amtoney524:
Wraps registered commands with error tracking.

note: this is currently a no-op as telemetry is not yet being initialized

https://github.com/user-attachments/assets/4036850e-98b2-4c38-957d-46c285bfad60

### DIFF
--- a/project-words.txt
+++ b/project-words.txt
@@ -34,7 +34,6 @@ tpus
 unassigning
 unassignment
 unassigns
-unregisters
 veryhighmem
 vscode
 vsdiag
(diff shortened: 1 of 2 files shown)


project-words.txt
```diff
@@ -34,6 +34,7 @@ tpus
 unassigning
 unassignment
 unassigns
+unregisters
```
hjjackyang: Nit: How is this not a legit word 🤣 !

I wonder if the grammatically correct word is "deregisters"? So we can avoid adding a project word.

hjjackyang (approved):
Neat changes!

kevineger (approved):

project-words.txt
amtoney524: hey looks like "deregisters" does the trick, thanks Jack!

amtoney524:
thanks for the reviews!!

PR #483, opened 2026-03-11
feat: log ephemeral auth event

hjjackyang:
**Why**? Upon supporting `drive.mount` and `auth.authenticate_user` in VS Code, it's beneficial to find out how often they are used in VS Code.

*Note: This logging call is still a no-op until telemetry module is officially launched.*

---

Internal tracking bug: b/491493637

### DIFF
--- a/src/auth/ephemeral.ts
+++ b/src/auth/ephemeral.ts
@@ -10,7 +10,6 @@ import { ColabClient } from '../colab/client';
 import { log } from '../common/logging';
 import { ColabAssignedServer } from '../jupyter/servers';
 import { telemetry } from '../telemetry';
-import { withErrorTracking } from '../telemetry/wrappers';
 
 /**
  * Handles ephemeral authorization by triggering an OAuth consent flow and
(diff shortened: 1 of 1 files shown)


src/auth/ephemeral.ts
```diff
@@ -26,6 +28,21 @@ export async function handleEphemeralAuth(
   apiClient: ColabClient,
   server: ColabAssignedServer,
   authType: AuthType,
+): Promise<void> {
+  telemetry.logHandleEphemeralAuth(authType);
+  await withErrorTracking(handleEphemeralAuthInternal)(
```
kevineger: Why wrap `withErrorTracking` here? Would tracking at the `ColabWebSocket` make more sense? That's the full seam between the Jupyter extension (process) and us (our process).

hjjackyang: Are you proposing to wrap around this event listener function?
https://github.com/googlecolab/colab-vscode/blob/main/src/jupyter/colab-proxy-websocket.ts#L70

Also, do we think it's also valuable to track errors from our `send` method override?
https://github.com/googlecolab/colab-vscode/blob/main/src/jupyter/colab-proxy-websocket.ts#L109

If so, I can split the websocket error tracking into a separate PR for clarity.

hjjackyang: I reverted the `withErrorTracking` change in this PR for now. Will create a separate PR for it.

amtoney524 (approved):

PR #484, opened 2026-03-11
feat: log process uncaught exceptions and unhandled rejections

amtoney524:
This change logs uncaught exceptions and unhandled rejections for the extension process

note:  this is currently a noop as the extension's not being initialized

https://github.com/user-attachments/assets/88547309-5755-4e6a-8267-92ee86bb9ad9

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -61,12 +61,8 @@ export async function activate(context: vscode.ExtensionContext) {
 }
 
 async function activateInternal(context: vscode.ExtensionContext) {
-  process.on('uncaughtException', (err: unknown) => {
-    telemetry.logError(err);
-  });
-  process.on('unhandledRejection', (reason: unknown) => {
-    telemetry.logError(reason);
-  });
+  process.on('uncaughtException', telemetry.logError);
+  process.on('unhandledRejection', telemetry.logError);
   const logging = initializeLogger(vscode, context.extensionMode);
   const jupyter = await getJupyterApi(vscode);
   logEnvInfo(jupyter);


src/extension.ts
```diff
@@ -61,6 +61,13 @@ export async function activate(context: vscode.ExtensionContext) {
 }
 
 async function activateInternal(context: vscode.ExtensionContext) {
+  process.on('uncaughtException', (err: unknown) => {
+    telemetry.logError(err);
+    throw err;
```
kevineger: No need to `throw` here right, since we're in the listener of the error having been thrown and unhanded by the process?

In fact, wouldn't it cause the listener to fire again? I suspect no as there's likely infinite recursion protection here.

amtoney524: I think you're right. Originally, I thought this could be similar to activation where rethrowing would result in a notification being rendered. Without the listener, it doesn't look like a notification gets triggered at all; so it doesn't seem like there's a point to rethrowing, and I removed that bit.

(Interestingly, rethrowing did not cause the listener to fire again)

src/extension.ts
```diff
@@ -61,6 +61,8 @@ export async function activate(context: vscode.ExtensionContext) {
 }
 
 async function activateInternal(context: vscode.ExtensionContext) {
+  process.on('uncaughtException', telemetry.logError);
+  process.on('unhandledRejection', telemetry.logError);
```
hjjackyang, line 65: Nit: Just wondering, do we need to add these to all functions where we'd like to log uncaught errors? If so, will it work to move these 2 lines into the `withErrorTracking` wrapper?

amtoney524: good question, I think the extension events are associated with the same process. I added `pid` logging in activation and server assignment and observed that they're consistent. 

Screenshot 2026-03-11 at 5 39 24 PM

hjjackyang: Thanks for confirming!

PR #486, opened 2026-03-11
feat: add support for additional scopes to auth provider

mounikasomisetty:
Mainly, in `createSession` if the `scopes` provided are not part of `REQUIRED_SCOPES` but part of `ALLOWED_SCOPES`, append `REQUIRED_SCOPES` to `scopes` to use incremental authorization to upgrade the session. All functionality should work the same if `scopes` is only `REQUIRED_SCOPES`.

Also for incremental authorization, setting prompt to undefined seemed to work best. Here are the two experiences:

[with the change](https://screencast.googleplex.com/cast/NTAxMjE4OTA3NjI1ODgxNnw0ODQ3YWRjYS1kMQ)

[with prompt always set to consent](https://screencast.googleplex.com/cast/NTM5Mzc3OTk3NTUyMDI1NnwxZWIxNjU2OC05OA)

Also:
+ Move scope declarations and utils to their own file (`scopes.ts`)
+ Added `LoginOptions`

### DIFF
--- a/src/auth/auth-provider.ts
+++ b/src/auth/auth-provider.ts
@@ -251,40 +251,27 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
   async createSession(scopes: string[]): Promise<AuthenticationSession> {
     this.assertReady();
     try {
-      if (!areScopesAllowed(scopes)) {
+      const sortedScopes = Array.from(new Set(scopes).values());
+      if (!areScopesAllowed(sortedScopes)) {
         throw new Error(
           `Only supports the following scopes: ${Array.from(ALLOWED_SCOPES.values()).join(', ')}`,
         );
       }
 
-      let shouldRequestIncrementalAuth = false;
-      let loginHint: string | undefined;
-      let finalScopes = scopes;
-      const existingSession = await this.getSession();
-      // If provided scopes, don't match REQUIRED_SCOPES
       if (
-        scopes.length !== REQUIRED_SCOPES.length ||
-        !REQUIRED_SCOPES.every((r) => scopes.includes(r))
+        sortedScopes.length < REQUIRED_SCOPES.length ||
+        !REQUIRED_SCOPES.every((r) => sortedScopes.includes(r))
       ) {
-        if (!existingSession) {
-          // Scope should have provided scopes and required scopes so the user
-          // does not have to login again to perform colab functions
-          finalScopes = Array.from(
-            new Set([...scopes, ...REQUIRED_SCOPES]).values(),
-          );
-        } else {
-          // Incremental authorization, so we are just adding the provided
-          // scopes to the existing scopes
-          shouldRequestIncrementalAuth = true;
-          loginHint = existingSession.account.id;
-          finalScopes = Array.from(
-            new Set([...existingSession.scopes, ...scopes]).values(),
-          );
-        }
+        throw new Error(
+          `Sessions must request at least the required scopes: ${Array.from(REQUIRED_SCOPES.values()).join(', ')}`,
+        );
       }
-      const sortedScopes = finalScopes.sort();
+      const existingSession = await this.getSession();
+      const loginHint = existingSession
+        ? existingSession.account.id
+        : undefined;
       const tokenInfo = await this.login(sortedScopes, {
-        includeGrantedScopes: shouldRequestIncrementalAuth,
+        includeGrantedScopes: !!existingSession,
         loginHint,
       });
       const user = await this.getUserInfo(tokenInfo.access_token);
@@ -295,7 +282,7 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
           id: user.email,
           label: user.name,
         },
-        scopes: finalScopes.sort(),
+        scopes: sortedScopes,
       };
       await this.storage.storeSession(newSession);
       this.oAuth2Client.setCredentials(tokenInfo);
--- a/src/auth/auth-provider.unit.test.ts
+++ b/src/auth/auth-provider.unit.test.ts
@@ -686,6 +686,12 @@ describe('GoogleAuthProvider', () => {
             credentials: DEFAULT_CREDENTIALS,
           },
           {
+            scopes: SCOPES,
+            includeGrantedScopes: true,
+            loginHint: 'foo@example.com',
+            credentials: DEFAULT_CREDENTIALS,
+          },
+          {
             scopes: UPGRADED_SCOPES,
             includeGrantedScopes: false,
             loginHint: undefined,
@@ -762,14 +768,14 @@ describe('GoogleAuthProvider', () => {
           hasValidSession: true,
         });
         sinon.assert.calledOnceWithExactly(loginStub, SCOPES, {
-          includeGrantedScopes: false,
-          loginHint: undefined,
+          includeGrantedScopes: true,
+          loginHint: DEFAULT_REFRESH_SESSION.account.id,
         });
       });
 
       it('creates a upgraded session', async () => {
         const signedInContext = signedInContextCalledWith();
-        const session = await authProvider.createSession(ADDITIONAL_SCOPES);
+        const session = await authProvider.createSession(UPGRADED_SCOPES);
 
         const newSession = {
           ...UPGRADED_AUTH_SESSION,
@@ -796,7 +802,7 @@ describe('GoogleAuthProvider', () => {
       it('upgrades an existing session', async () => {
         storageStub.getSessions.resolves([DEFAULT_REFRESH_SESSION]);
 
-        const session = await authProvider.createSession(ADDITIONAL_SCOPES);
+        const session = await authProvider.createSession(UPGRADED_SCOPES);
 
         expect(session).to.deep.equal({
           ...UPGRADED_AUTH_SESSION,
--- a/src/auth/flows/flows.ts
+++ b/src/auth/flows/flows.ts
@@ -33,8 +33,8 @@ export interface OAuth2TriggerOptions {
    *
    * If the value is "consent" - the consent dialog is forced to be displayed
    * every time, even if the user has previously granted access.
-   * If the value is not specified - the consent dialog defaults to it's default
-   * i.e. it only shows for the scopes the user did not already consent to.
+   * If the value is not specified - the consent dialog defaults to its default
+   * e.g., it only shows for the scopes the user did not already consent to.
    *
    * This should always be "consent" unless `includeGrantedScopes` is `true`
    * to allow for incremental auth.
--- a/src/colab/client.unit.test.ts
+++ b/src/colab/client.unit.test.ts
@@ -9,6 +9,7 @@ import { expect } from 'chai';
 import fetch, { Response } from 'node-fetch';
 import { SinonStub, SinonMatcher } from 'sinon';
 import * as sinon from 'sinon';
+import { REQUIRED_SCOPES } from '../auth/scopes';
 import { Session } from '../jupyter/client/generated';
 import { ColabAssignedServer } from '../jupyter/servers';
 import { TestUri } from '../test/helpers/uri';
@@ -73,7 +74,7 @@ const DEFAULT_ASSIGNMENT: Assignment = {
 
 describe('ColabClient', () => {
   let fetchStub: sinon.SinonStubbedMember<typeof fetch>;
-  let sessionStub: SinonStub<[], Promise<string>>;
+  let sessionStub: SinonStub<[readonly string[]], Promise<string>>;
   let client: ColabClient;
   let onAuthErrorStub: SinonStub<[], Promise<void>>;
 
@@ -81,7 +82,12 @@ describe('ColabClient', () => {
     fetchStub = sinon.stub(fetch, 'default').callsFake(() => {
       throw new Error('fetch was called with non-matching call');
     });
-    sessionStub = sinon.stub<[], Promise<string>>().resolves(BEARER_TOKEN);
+    sessionStub = sinon
+      .stub<[readonly string[]], Promise<string>>()
+      .callsFake(() => {
+        throw new Error('sessionStub was called with non-matching call');
+      });
+    sessionStub.withArgs(REQUIRED_SCOPES).resolves(BEARER_TOKEN);
     onAuthErrorStub = sinon.stub();
     client = new ColabClient(
       new URL(`https://${COLAB_HOST}`),


src/auth/scopes.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/** Minimum required scope set for executing a runtime */
```
kevineger: Nit: When possible trying to stick to _server_ instead of _runtime_ for consistency.

```suggestion
/** Minimum required scope set for assigning and using a Colab server. */
```

src/auth/scopes.ts
```diff
@@ -0,0 +1,49 @@
...
+export const REQUIRED_SCOPES = [
+  'profile',
+  'email',
+  'https://www.googleapis.com/auth/colaboratory',
+] as const;
+
+/** Scopes required to use the drive integration */
```
kevineger: Nit: capitalize proper noun.

```suggestion
/** Scopes required to use the Drive integration */
```

src/auth/scopes.ts
```diff
@@ -0,0 +1,49 @@
...
+] as const;
+
+/** Set of all scopes that are permitted to be used by this extension */
+export const ALLOWED_SCOPES = new Set<string>([
+  ...REQUIRED_SCOPES,
+  ...DRIVE_SCOPES,
+]);
+
+/** Returns true if the provided scopes are all supported */
+export function areScopesAllowed(scopes?: readonly string[]): boolean {
```
kevineger, line 26: Curious, why allow passing `undefined`? Doesn't look like any of the current callers leverage that.

src/auth/scopes.ts
```diff
@@ -0,0 +1,49 @@
...
+  return scopes.every((scope) => ALLOWED_SCOPES.has(scope));
+}
+
+/** Function to check if scopesToCheck has AT LEAST every scope in scopes */
+export function hasScopes(
+  scopesToCheck: readonly string[],
+  scopes: readonly string[],
+): boolean {
+  return scopes.every((r) => scopesToCheck.includes(r));
+}
```
kevineger: This is only used once and personally I find reading the `scopes.every` more obvious at a glance than just `hasScopes` where from the callsite it's unclear which is the scopes to check and which is the criteria. Recommend simply inlining the code, which is a 1-liner.

kevineger: I feel the same about `matchesRequiredScopes` and `upgradeScopes`. Here I think inlining the single usages improves readability and makes following the logic a bit easier.

src/auth/auth-provider.ts
```diff
@@ -69,10 +71,8 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
    * Initializes the GoogleAuthProvider.
    *
    * @param vs - The VS Code API.
-   * @param context - The extension context used for managing lifecycle.
+   * @param storage - The storage client for persisting sessions
```
kevineger: Nit: end with `.` for consistency with other `@param`-s here and throughout. Apply below in `getOrCreateSession` and `createSession` as well.

Thanks for capturing the mismatch! Seems like there's a lint rule that needs to be setup to catch this...

src/auth/auth-provider.ts
```diff
@@ -226,39 +229,44 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
...
         throw new Error(
-          `Only supports the following scopes: ${sortedScopes.join(', ')}`,
+          `Only supports the following scopes: ${Array.from(ALLOWED_SCOPES.values()).join(', ')}`,
         );
       }
+
+      let targetScopes = scopes;
+      if (!matchesRequiredScopes(scopes)) {
+        targetScopes = upgradeScopes(scopes);
+      }
```
kevineger: I'm not totally following the logic here, mind walking me through it? Full disclosure, I'm hitting late afternoon / EOD and my brain's certainly starting to slow down.

If `createSession` is called with anything but the following, we should error:

- `[...REQUIRED_SCOPES]`
- `[...REQUIRED_SCOPES, ...DRIVE_SCOPES]`

Otherwise, we can progress with calling `this.login` to either create the first session, or upgrade an existing one.

So why do we `upgradeScopes` the provided `scopes`?

src/auth/auth-provider.unit.test.ts
```diff
@@ -53,6 +50,31 @@ const DEFAULT_AUTH_SESSION: vscode.AuthenticationSession = {
   account: DEFAULT_REFRESH_SESSION.account,
   scopes: DEFAULT_REFRESH_SESSION.scopes.sort(),
 };
+const ADDITIONAL_SCOPES = Array.from(DRIVE_SCOPES);
```
kevineger: Nit: spread for consistency and readability?

```suggestion
const ADDITIONAL_SCOPES = [...DRIVE_SCOPES];
```

src/auth/auth-provider.unit.test.ts
```diff
@@ -125,6 +147,9 @@ describe('GoogleAuthProvider', () => {
       loginStub,
     );
     authProvider.onDidChangeSessions(onDidChangeSessionsStub);
+
+    // By default there are no stored sessions
+    storageStub.getSessions.resolves([]);
```
kevineger: I could be mistaken, but won't this be the stub default behaviour already?

src/auth/auth-provider.unit.test.ts
```diff
@@ -307,6 +333,21 @@ describe('GoogleAuthProvider', () => {
...
+            DEFAULT_REFRESH_SESSION,
+          ]);
+        });
+
+        it('rejects', async () => {
+          await expect(authProvider.initialize()).to.eventually.be.rejectedWith(
+            /at most 1 session/,
+          );
+        });
+      });
```
kevineger: Single case, flatten `it`.

```suggestion
      it('rejects when there are multiple stored sessions', async () => {
        storageStub.getSessions.resolves([
          DEFAULT_REFRESH_SESSION,
          DEFAULT_REFRESH_SESSION,
        ]);

        await expect(authProvider.initialize()).to.eventually.be.rejectedWith(
          /at most 1 session/,
        );
      });
```

src/auth/auth-provider.unit.test.ts
```diff
@@ -452,25 +493,16 @@ describe('GoogleAuthProvider', () => {
         await authProvider.initialize();
       });
 
-      it('returns an empty array when scopes deviate from the supported set', async () => {
```
kevineger, line 455: Should we not be preserving this behaviour?

src/auth/auth-provider.unit.test.ts
mounikasomisetty: Sure, I was just following how `SCOPES` was declared but I'll change them both

src/auth/auth-provider.unit.test.ts
mounikasomisetty: I couldn't find a reference to this in the docs but the default behaviour is to return `undefined`, and in this case since the method doesn't actually return `undefined` we don't check for it before we call `.length` - that throws an erorr nesitating this call

src/auth/auth-provider.unit.test.ts
mounikasomisetty: the behavior is persisted, but since we first check for `!this.session` it's not a valid test case in this `describe` block. There is a similar test case in the next `describe` block though.

src/auth/scopes.ts
mounikasomisetty: `AuthProvider.getSessions` uses it when it's `scopes` is `undefined`

src/auth/scopes.ts
mounikasomisetty: Done

src/auth/auth-provider.ts
mounikasomisetty: This was due to my not understanding what "upgrading" a session meant. I was trying to replace the existing session in place.

mounikasomisetty: Made the changes to actually upgrade the session, will add tests asap

src/auth/scopes.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
hjjackyang:
```suggestion
 * Copyright 2026 Google LLC
```

src/auth/scopes.ts
```diff
@@ -0,0 +1,30 @@
...
+  'https://www.googleapis.com/auth/colaboratory',
+] as const;
+
+/** Scopes required to use the Drive integration */
+export const DRIVE_SCOPES = [
+  'https://www.googleapis.com/auth/drive.readonly',
+] as const;
+
+/** Set of all scopes that are permitted to be used by this extension */
+export const ALLOWED_SCOPES = new Set<string>([
```
hjjackyang: Nit: Explicitly type `ALLOWED_SCOPES` as `ReadonlySet` to avoid unintentional mutation.

```suggestion
export const ALLOWED_SCOPES: ReadonlySet<string> = new Set<string>([
```

src/auth/flows/flows.ts
```diff
@@ -25,6 +25,21 @@ export interface OAuth2TriggerOptions {
...
   /** The PKCE challenge string which if specific should be included with the auth request. */
   readonly pkceChallenge?: string;
+  /** Whether to include granted scopes in the new request. */
+  readonly includeGrantedScopes?: boolean;
+  /** The login hint to pre-fill the email (for incremental authorization). */
+  readonly loginHint?: string;
+  /** Influences how the user is presented with the consent dialog.
+   *
+   * If the value is "consent" - the consent dialog is forced to be displayed
+   * every time,even if the user has previously granted access.
```
hjjackyang:
```suggestion
   * every time, even if the user has previously granted access.
```

src/auth/flows/flows.ts
```diff
@@ -25,6 +25,21 @@ export interface OAuth2TriggerOptions {
...
+  /** Whether to include granted scopes in the new request. */
+  readonly includeGrantedScopes?: boolean;
+  /** The login hint to pre-fill the email (for incremental authorization). */
+  readonly loginHint?: string;
+  /** Influences how the user is presented with the consent dialog.
+   *
+   * If the value is "consent" - the consent dialog is forced to be displayed
+   * every time, even if the user has previously granted access.
+   * If the value is not specified - the consent dialog defaults to it's default
+   * i.e. it only shows for the scopes the user did not already consent to.
```
kevineger: Nit: small grammatical thing

```suggestion
   * If the value is not specified - the consent dialog defaults to its default
   * e.g., it only shows for the scopes the user did not already consent to.
```

src/auth/auth-provider.ts
```diff
@@ -226,47 +227,75 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
...
+          finalScopes = Array.from(
+            new Set([...existingSession.scopes, ...scopes]).values(),
+          );
+        }
+      }
+      const sortedScopes = finalScopes.sort();
+      const tokenInfo = await this.login(sortedScopes, {
+        includeGrantedScopes: shouldRequestIncrementalAuth,
+        loginHint,
+      });
```
kevineger: Can this be simplified to: check if the scopes are allowed, validate they're "at least" the required ones, and then get the relevant config from the existing session if it exists?

I don't think we want to augment the provided `scopes`. In other words, a user requesting scopes `[foo, bar]` shouldn't create a session with `[foo, bar, ...REQUIRED_SCOPES]`.

```suggestion
      const sortedScopes = Array.from(new Set(scopes).values());
      if (!areScopesAllowed(Array.from(sortedScopes.values()))) {
        throw new Error(
          `Only supports the following scopes: ${Array.from(ALLOWED_SCOPES.values()).join(', ')}`,
        );
      }

      if (
        scopes.length <= REQUIRED_SCOPES.length ||
        !REQUIRED_SCOPES.every((r) => scopes.includes(r))
      ) {
        throw new Error(
          `Sessions must request at least the required scopes: ${Array.from(ALLOWED_SCOPES.values()).join(', ')}`,
        );
      }

      const existingSession = await this.getSession();
      const loginHint = existingSession
        ? existingSession.account.id
        : undefined;
      const tokenInfo = await this.login(sortedScopes, {
        includeGrantedScopes: !!existingSession,
        loginHint,
      });
```

src/colab/client.ts
```diff
@@ -465,7 +467,10 @@ export class ColabClient {
     endpoint: URL,
     init: RequestInit,
     schema?: z.ZodType,
-    { requireAccessToken = true }: IssueRequestOptions = {},
+    {
+      requireAccessToken = true,
+      scopes = REQUIRED_SCOPES,
```
kevineger, line 472: Consider updating `client.unit.test.ts`'s `sessionStub` to ensure we are always passing the `REQUIRED_SCOPES`.

```ts
sessionStub = sinon
  .stub<[string[]], Promise<string>>()
  .callsFake((scopes) => {
    expect(scopes).to.deep.equal(REQUIRED_SCOPES);
    return Promise.resolve(BEARER_TOKEN);
  });
```

kevineger (approved):
Great work! Thanks again for working through all the feedback.

1. Sorry it took me so long to get back to this. Been crazy busy and randomized the past week (oncall, etc.).
2. Looking good! I think we can simplify `createSession` a bit. Hopefully my suggestion made sense.
3. LG on incremental (`consent`/`undefined`) choice, thanks for ironing that out!

src/auth/auth-provider.ts
mounikasomisetty: I think I understand what you are suggesting but it is not clear to me how you want to proceed with the case where a drive session is created first. Would you prefer that drive calls pass in `[...DRIVE_SCOPES, ...REQUIRED_SCOPES]`?

kevineger: If a Drive session is created first, is `createSession` ever called?

1. Drive session created
2. Drive session stored
3. API call made to create an assignment (just needs `REQUIRED_SCOPES`)
4. `getSessions` returns the super-set scope that includes Drive

> Would you prefer that drive calls pass in `[...DRIVE_SCOPES, ...REQUIRED_SCOPES]`?

That's what I had assumed we were going to do, yes. That ok?

mounikasomisetty: > If a Drive session is created first, is `createSession` ever called?

Maybe I'm misunderstanding but yes, in step 1. 

That's fine, I can make the change, it just made sense to me that we would only pass in additional scopes, but it's semantics

src/colab/client.ts
mounikasomisetty: functionally I made this change but it looks a little different to help me not have to change it again downstream

PR #488, opened 2026-03-13
feat: log errors from jupyter server provider

amtoney524:
This change logs errors originating from the public jupyter server provider methods.

Additionally, `telemetry/wrappers.ts` was renamed to `telemetry/decorators.ts` to reflect the addition of a decorator and be consistent with the `common/logging` module.

Note: this change is a no-op as telemetry is not yet initialized 

https://github.com/user-attachments/assets/7de57268-2d6d-4ea1-820d-d7f325eba322

### DIFF
--- a/src/telemetry/decorators.ts
+++ b/src/telemetry/decorators.ts
@@ -24,13 +24,7 @@ export function trackErrors<
     return descriptor;
   }
 
-  descriptor.value = function (
-    this: unknown,
-    ...args: Parameters<T>
-  ): ReturnType<T> {
-    return withErrorTracking(originalMethod).apply(this, args);
-  } as T;
-
+  descriptor.value = withErrorTracking(originalMethod);
   return descriptor;
 }
 


src/telemetry/decorators.unit.test.ts
```diff
@@ -0,0 +1,157 @@
+/**
```
amtoney524, line 1: I'm having a hard time marking this as `renamed` while amending the commit - any ideas on what to try next?

kevineger: Git tracks content, not explicit renames. Even `git mv` is a convenience shortcut that performs a `git rm` on the source and a `git add` on the destination file in one go. There's a few techniques to fix _marking it has renamed/moved_ after you've committed, all largely follow the same approach: rebase and reset changes, commit the move/rename (without content changes) and then make your changes.

src/telemetry/decorators.unit.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    throw new Error('Synchronous error');
+  }
+
+  @trackErrors
+  async asyncErrorMethod(): Promise<void> {
+    return Promise.reject(new Error('Asynchronous error'));
+  }
+}
+
+describe('trackError', () => {
```
kevineger:
```suggestion
describe('trackErrors', () => {
```

src/telemetry/decorators.ts
```diff
@@ -7,6 +7,33 @@
...
+  descriptor: TypedPropertyDescriptor<T>,
+): TypedPropertyDescriptor<T> {
+  const originalMethod = descriptor.value;
+
+  // Ensure the property is a function.
+  if (typeof originalMethod !== 'function') {
+    return descriptor;
+  }
+
+  descriptor.value = function (
```
kevineger: Replacing the `descriptor.value` with this anonymous wrapping function will lose the original function's name (and other) property which can complicate debugging.

I think there's some approaches with `Object.defineProperty` and setting/copying properties attached to the original function we could explore.

But before doing that, I think we may need to take a look at how we're wrapping...

`trackErrors` calls `withErrorTracking` inside the wrapper function on every **method invocation**. This will unnecessarily create a new closure + function binding every time the method is called. I think we can instead wrap the method once when the decorator is applied (`descriptor.value = withErrorTracking(originalMethod)`).

src/telemetry/decorators.unit.test.ts
amtoney524: I had tried to move it in [one commit](https://github.com/googlecolab/colab-vscode/pull/488/changes/36200a367d1c7d3fa142cee5c3a0f82781bf0264) and add the new test cases in the following commit, and still no luck :/ the end result looks like an added file based on the content change

src/telemetry/decorators.ts
amtoney524: good point, changing the descriptor value as described above seemed to simplify things! I also reviewed the trace logs with this change (threw in `assignments.assignServer`) and the resulting stack seemed ok, wdyt? 
Screenshot 2026-03-13 at 2 03 30 PM

src/telemetry/decorators.unit.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      const fnCalled = new Deferred<void>();
+      const fn = async (input: string) => {
+        await fnCalled.promise;
+        return input + 'bar';
+      };
+
+      const result = withErrorTracking(fn)('foo');
+      fnCalled.resolve();
+
+      await expect(result).to.eventually.equal('foobar');
```
kevineger, line 130: Is this a potential race? L123 will finish but it's only `foobar` because we've `await`-ed the `expect` here (queued behind in the microtask queue)?

amtoney524: This is intentional, I think we need the `await` here -
1. If we comment out `fnCalled.resolve()` and continue to call  `await expect(result).to.eventually.equal('foobar');`, the test will fail with a timeout error as expected.
2. If we then remove the `await`, the test erroneously passes.

For this case, I wanted to also assert the return value is as expected when using the wrapper.

I'm going to submit as is, but if I'm missing something, please un-resolve and I'll address it!

PR #494, opened 2026-03-17
feat: api and client definitions for resource monitoring

kevineger:
This API definition was largely _reverse engineered_ from our web front-end, our backend impl. and the observed Chrome response(s) for `api/colab/resources`.

_The following all contain internal links, but leaving as breadcrumbs for @hjjackyang who's expressed interest in building this out._

- [`Resources`](http://google3/logs/proto/colab/resourcestats.proto;l=20;rcl=877490900)
- [Backend definitions (partial)](http://google3/research/colab/datalab/tunnel/tunnelutil.go;l=406;rcl=868754694)
- Some of the fields (for accelerators) get merged in from the runtime ([example](http://google3/third_party/py/google/colab/_serverextension/_resource_monitor.py;l=227;rcl=783013361))

Support for this endpoint is a bit convoluted and varies slightly per machine/accelerator. I've defined what's in this PR as best I could, but I wouldn't be surprised if there are little mistakes here and there. Hopefully is at least a good starting point 🤞.

### DIFF
--- a/src/colab/api.ts
+++ b/src/colab/api.ts
@@ -422,6 +422,91 @@ export const SessionSchema: z.ZodType<GeneratedSession> = z.object({
 });
 export type Session = z.infer<typeof SessionSchema>;
 
+/** Information about memory usage on a Colab runtime. */
+export const MemorySchema = z.object({
+  /** Total memory available in bytes. */
+  totalBytes: z.number(),
+  /** Free memory available in bytes. */
+  freeBytes: z.number(),
+});
+/** Memory usage on a Colab runtime. */
+export type Memory = z.infer<typeof MemorySchema>;
+
+/** Information about a GPU on a Colab runtime. */
+export const GpuInfoSchema = z.object({
+  /** The name of the GPU. */
+  name: z.string(),
+  /** Memory used in bytes. */
+  memoryUsedBytes: z.number(),
+  /** Total memory in bytes. */
+  memoryTotalBytes: z.number(),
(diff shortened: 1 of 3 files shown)


src/colab/api.ts
```diff
@@ -422,6 +422,91 @@ export const SessionSchema: z.ZodType<GeneratedSession> = z.object({
...
+    totalBytes: z.number(),
+    /** Used space on the filesystem in bytes (legacy). */
+    usedBytes: z.number().optional(),
+    /** Free space on the filesystem in bytes. */
+    freeBytes: z.number().optional(),
+  })
+  .transform((val) => ({
+    name: val.name ?? val.label ?? '',
+    totalBytes: val.totalBytes,
+    freeBytes: val.freeBytes ?? val.totalBytes - (val.usedBytes ?? 0),
```
hjjackyang, line 470: Nit: Is this transformation necessary? I see we currently only display Disk in terms of `usedBytes` / `totalBytes`.

Image

Image

hjjackyang (approved):
Left a nit, but in any case, feel free to merge as is, and I can adjust it accordingly when I work on the remaining.

src/colab/api.ts
kevineger: Exists for backwards compat due to `usedBytes` being legacy. Wasn't totally clear to me from the impl. if we needed it or not so left it in.

I'll merge with this in but if your downstream work finds it's unecessary please feel free to remove!

PR #496, opened 2026-03-18
chore: lint and fix doc-blocks

kevineger:
Bit of a nasty task... Tried to leverage AI automation to fix all the issues which did a surprisingly lousy job. After getting most of the way with that, I did one manual pass fixing up the issues.

I know there's likely to be a few spots where the comment(s) could be better, but at least now with linting we're guaranteed to have consistency and comments that list all params, refer to the correct params by name, are formatted consistently, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -369,6 +369,7 @@
     "eslint-plugin-check-file": "^3.1.0",
     "eslint-plugin-headers": "^1.3.3",
     "eslint-plugin-import": "^2.31.0",
+    "eslint-plugin-jsdoc": "^62.8.0",
     "eslint-plugin-tsdoc": "^0.4.0",
     "mocha": "^10.8.2",
     "prettier": "3.4.2",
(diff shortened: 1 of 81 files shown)


package.json
```diff
@@ -369,6 +369,7 @@
     "eslint-plugin-check-file": "^3.1.0",
     "eslint-plugin-headers": "^1.3.3",
     "eslint-plugin-import": "^2.31.0",
+    "eslint-plugin-jsdoc": "^62.8.0",
     "eslint-plugin-tsdoc": "^0.4.0",
```
hjjackyang, line 373: Nit: There is currently a moderate vulnerability associated with `eslint-plugin-tsdoc` between version `0.3.0` and `0.5.0` (can view from running `npm audit`).  Since you are touching JSDoc lint in this PR, I wonder if it makes sense to also bump `eslint-plugin-tsdoc` to the latest version `0.5.2`?

(This is a minor version bump, so should be backward compatible.)

kevineger: I'll take a look and handle in isolation.

PR #498, opened 2026-03-19
fix: update drive scopes to include `REQUIRED_SCOPES` and use `drive.…

mounikasomisetty:
Update drive scopes to include `REQUIRED_SCOPES` and use `drive.file` as it's already supported. 

Also fix tests since `ADDITIONAL_SCOPES` and `UPGRADED_SCOPES` are the same now.

### DIFF
--- a/src/auth/scopes.ts
+++ b/src/auth/scopes.ts
@@ -13,7 +13,8 @@ export const REQUIRED_SCOPES = [
 
 /** Scopes required to use the Drive integration */
 export const DRIVE_SCOPES = [
-  'https://www.googleapis.com/auth/drive.readonly',
+  ...REQUIRED_SCOPES,
+  'https://www.googleapis.com/auth/drive.file',
 ] as const;
 
 /** Set of all scopes that are permitted to be used by this extension */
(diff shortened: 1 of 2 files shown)


src/auth/scopes.ts
```diff
@@ -13,7 +13,8 @@ export const REQUIRED_SCOPES = [
 
 /** Scopes required to use the Drive integration */
 export const DRIVE_SCOPES = [
-  'https://www.googleapis.com/auth/drive.readonly',
+  ...REQUIRED_SCOPES,
+  'https://www.googleapis.com/auth/drive.file',
```
hjjackyang, line 17: Nit: What's the reason of changing `drive.readonly` to `drive.file` scope here? Will `drive.file` scope imply write permission?

mounikasomisetty: `drive.readonly` isn't in our current oAuth2 config and adding it would be a process to say the least. Since `drive.file` is already there and works, we had to pivot to that. Also `drive.file` is a less sensitive scope than `drive.readonly`

hjjackyang: SG!

kevineger: `drive.readonly` is a restricted scope and they recommend `drive.file` wherever possible. Restricted scopes require app review and I'm unsure we'd qualify. Since `drive.file` works for access to all files created by our _app_ (which includes Colab web), then it's sufficient for our use-case here.

https://developers.google.com/workspace/drive/api/guides/api-specific-auth#restricted-scopes

PR #501, opened 2026-03-19
refactor: move `issueRequest` logic to it's own class

mounikasomisetty:
Move `ColabClient.issueRequest` logic to it's own class so it can be used for other clients (will be adding `DriveClient` in a follow up CL).

I left the `ColabClient` test's as-is for the most part as it demonstrates that the refactor is WAI.

### DIFF
--- a/src/colab/client.ts
+++ b/src/colab/client.ts
@@ -73,7 +73,6 @@ interface AssignParams {
  * A client for interacting with the Colab APIs.
  */
 export class ColabClient {
-  private readonly requester: Transport;
   private readonly httpsAgent?: https.Agent;
 
   /**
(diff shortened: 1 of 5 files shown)


src/colab/client.ts
```diff
@@ -71,16 +67,11 @@ interface AssignParams {
...
-interface IssueRequestOptions {
-  requireAccessToken?: boolean;
-  scopes?: readonly string[];
-}
-
 /**
  * A client for interacting with the Colab APIs.
  */
 export class ColabClient {
+  private readonly requester: Transport;
```
hjjackyang: Nit: Just providing some early feedback - Have you considered making `ColabClient` extends `Transport` (perhaps name it `BaseHttpClient`?) instead of initializing it as a member?

This way all these low-level `issueRequest` methods can stay `protected`.

mounikasomisetty: I did consider it, but I think generally composition is preferred to inheritance.

I'm OK with using inheritance though, I just went back and forth on it.

src/colab/transport.ts
```diff
@@ -0,0 +1,177 @@
...
+   * handling errors and parses the response to the provided schema.
+   *
+   * @param endpoint - The endpoint to issue the request to.
+   * @param init - The request init to use for the fetch.
+   * @param schema - The schema to validate the response against.
+   * @param options - The additional options for the request.
+   * @returns A promise that resolves the parsed response when the request is
+   * complete.
+   */
+  async issueRequestAndParse<T extends z.ZodType>(
```
hjjackyang, line 72: Nit: Any reason you abandoned the original method overload approach?

src/colab/transport.ts
```diff
@@ -0,0 +1,177 @@
...
+  if (!s.startsWith(XSSI_PREFIX)) {
+    return s;
+  }
+  return s.slice(XSSI_PREFIX.length);
+}
+
+/**
+ * Wrapper for errors thrown from issuing requests.
+ */
+export class ColabRequestError extends Error {
```
hjjackyang, line 159: Nit: This class probably deserves its own file?

src/colab/transport.ts
mounikasomisetty: Ah so apparently overloaded signatures are not visible outside of the class

PR #503, opened 2026-03-20
feat: add `DriveClient` to interact with the Google Drive API

mounikasomisetty:
This PR adds `DriveClient` which interacts with the Google Drive API to support the upcoming "Import notebook" command.  

Currently this is a no-op as nothing interacts with this client.

### DIFF
--- a/src/drive/api.ts
+++ b/src/drive/api.ts
@@ -11,7 +11,7 @@
 import { z } from 'zod';
 
 /**
- * The schema for Drive file metadata.
+ * Drive file metadata.
  *
  * This is a small subset of the full object that is returned.
  */
--- a/src/drive/client.ts
+++ b/src/drive/client.ts
@@ -25,26 +25,28 @@ export class DriveClient {
    *
    * @param getAccessToken - Function to retrieve the access token.
    * @param onAuthError - Callback when an auth error occurs.
-   * @returns A new ColabClient instance.
+   * @returns A new DriveClient instance.
    */
   static create(
     getAccessToken: () => Promise<string>,
     onAuthError: (() => Promise<void>) | undefined,
   ): DriveClient {
-    return new DriveClient(
-      buildFetchChain(
-        [
-          createAcceptJsonMiddleware(),
-          createErrorMiddleware(),
-          createAuthMiddleware(getAccessToken, onAuthError),
-        ],
-        fetch,
-      ),
-    );
+    const baseMiddleware = [
+      createErrorMiddleware(),
+      createAuthMiddleware(getAccessToken, onAuthError),
+    ];
+    const jsonMiddleware = [...baseMiddleware, createAcceptJsonMiddleware()];
+    const blobFetch = buildFetchChain(baseMiddleware, fetch);
+    const jsonFetch = buildFetchChain(jsonMiddleware, fetch);
+    return new DriveClient(blobFetch, jsonFetch);
   }
 
   private constructor(
-    private readonly authenticatedFetch: (
+    private readonly blobFetch: (
+      url: string | Request,
+      init?: RequestInit,
+    ) => Promise<Response>,
+    private readonly jsonFetch: (
       url: string | Request,
       init?: RequestInit,
     ) => Promise<Response>,
@@ -52,6 +54,10 @@ export class DriveClient {
 
   /**
    * Retrieves the content of a file from Google Drive.
+   *
+   * TODO: Convert to a stream to better support larger files as
+   * response.arrayBuffer() loads the entire file into memory at once.
+   * The tricky part here is that VsCode Web does not accept streams.
    *
    * @param fileId - The ID of the Drive file.
    * @param signal - Optional {@link AbortSignal} to cancel the request.
@@ -62,7 +68,7 @@ export class DriveClient {
     signal?: AbortSignal,
   ): Promise<Uint8Array> {
     const url = new URL(`${FILES_ENDPOINT}/${fileId}?alt=media`);
-    const response = await this.authenticatedFetch(url.toString(), {
+    const response = await this.blobFetch(url.toString(), {
       method: 'GET',
       signal,
     });
@@ -99,7 +105,7 @@ export class DriveClient {
     url.searchParams.append('fields', 'name');
 
     return fetchAndParse(
-      this.authenticatedFetch,
+      this.jsonFetch,
       url.toString(),
       DriveFileMetadataSchema,
       {
--- a/src/drive/client.unit.test.ts
+++ b/src/drive/client.unit.test.ts
@@ -37,13 +37,10 @@ describe('DriveClient', () => {
   describe('getDriveFileContent', () => {
     it('should return the file content', async () => {
       const fileId = 'test-file-id';
-      const fileContent = { cells: [], metadata: {} };
-      const expectedContent = new TextEncoder().encode(
-        JSON.stringify(fileContent),
-      );
-      fetchStub.resolves(
-        new Response(JSON.stringify(fileContent), { status: 200 }),
-      );
+      // .ipynb files are just JSON files
+      const fileContent = JSON.stringify({ cells: [], metadata: {} });
+      const expectedContent = new TextEncoder().encode(fileContent);
+      fetchStub.resolves(new Response(fileContent, { status: 200 }));
 
       const result = await client.getDriveFileContent(fileId);
 


src/drive/api.ts
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * API types for interacting with Google Drive's APIs.
+ */
+
+import { z } from 'zod';
+
+export const DriveFileMetadataSchema = z.object({
```
kevineger, line 18: Some homework for me to lint `export const` variables to have docblocks. Optionally consider adding a description here. Fine to ignore since linting _should_ be the one to enforce this.

In the comment, consider dropping a note that naturally this is a small subset of the full return which has a bunch more.

src/drive/client.ts
```diff
@@ -0,0 +1,109 @@
...
+        fetch,
+      ),
+    );
+  }
+
+  private constructor(
+    private readonly fetch: (
+      url: string | Request,
+      init?: RequestInit,
+    ) => Promise<Response>,
```
kevineger: Curious why inject this? It can be stubbed without (see other tests which stub `fetch`).

src/drive/client.ts
```diff
@@ -0,0 +1,109 @@
...
+  async getDriveFileContent(
+    fileId: string,
+    signal?: AbortSignal,
+  ): Promise<Uint8Array> {
+    const encoder = new TextEncoder();
+    const url = new URL(`${FILES_ENDPOINT}/${fileId}?alt=media`);
+    const response = await fetchAndParse(
+      this.fetch,
+      url.toString(),
+      z.unknown(),
```
kevineger: Since this isn't JSON we need to validate, should this be something like `z.instanceof(Buffer)` ? Or `Blob` or `Uint8Array`?

Or perhaps the better question, should this be calling the Zod-based `fetchAndParse`? It doesn't need the _XSSI_ stripping and we aren't parsing the schema.

src/drive/client.ts
```diff
@@ -0,0 +1,109 @@
...
+   * Retrieves the metadata for a file from Google Drive.
+   *
+   * @param id - The ID of the Drive file.
+   * @param signal - Optional {@link AbortSignal} to cancel the request.
+   * @returns A promise that resolves to the file metadata.
+   */
+  async getDriveFileMetadata(
+    id: string,
+    signal?: AbortSignal,
+  ): Promise<DriveFileMetadata> {
```
kevineger: Since all we need right now is the `name`, should we maybe make this `getDriveFileName` and return a `string`?

kevineger: Ah, just saw you have exactly that below! Would it make sense to make this method private and/or inline it in `getDriveFileName`?

src/drive/client.ts
```diff
@@ -0,0 +1,109 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
hjjackyang, line 5: Nit: Add a blank line below the license docblock.

```suggestion
 */
 
```

src/drive/client.ts
hjjackyang: +1 on making this `private` for the time being. We can lift the visibility when future use case arises.

src/drive/client.ts
mounikasomisetty: Good point! Switched to just using `this.fetch`

src/drive/client.ts
mounikasomisetty: I'm sorry - I don't understand the suggestion/question as in my mind I was just following the pattern in ColabClient. Could you explain more?

I renamed this to `authenticatedFetch` to distinguish it from the standard `fetch`.

src/drive/api.ts
```diff
@@ -0,0 +1,25 @@
...
+ * The schema for Drive file metadata.
+ *
+ * This is a small subset of the full object that is returned.
+ */
+export const DriveFileMetadataSchema = z.object({
+  name: z.string(),
+});
+
+/**
+ * The type for Drive file metadata.
```
kevineger, line 23:
```suggestion
 * Drive file metadata.
```

src/drive/client.ts
kevineger: My gosh, please ignore me - sorry! That's what I get for reviewing while tired and sick 🤧

src/drive/client.ts
```diff
@@ -0,0 +1,111 @@
...
+      this.authenticatedFetch,
+      url.toString(),
+      DriveFileMetadataSchema,
+      {
+        method: 'GET',
+        signal,
+      },
+    );
+  }
+}
```
kevineger, line 117: `DriveClient.create` applies the JSON middleware to the entire fetch chain. But we don't want that for `getDriveFileContent`, which uses `alt=media` to fetch binary data. Forcing Accept: `application/json` on binary download requests is misleading to the server and could lead to unexpected behavior if the Drive API strictly adheres to the `Accept` header over the `alt` parameter.

```suggestion
    static create(
    getAccessToken: () => Promise<string>,
    onAuthError: (() => Promise<void>) | undefined,
  ): DriveClient {
    const baseMiddleware = [
      createErrorMiddleware(),
      createAuthMiddleware(getAccessToken, onAuthError),
    ];
    const jsonMiddleware = [...baseMiddleware, createAcceptJsonMiddleware()];
    const blobFetch = buildFetchChain(baseMiddleware, fetch);
    const jsonFetch = buildFetchChain(jsonMiddleware, fetch);
    return new DriveClient(blobFetch, jsonFetch);
  }

  private constructor(
    private readonly blobFetch: (
      url: string | Request,
      init?: RequestInit,
    ) => Promise<Response>,
    private readonly jsonFetch: (
      url: string | Request,
      init?: RequestInit,
    ) => Promise<Response>,
  ) {}

  /**
   * Retrieves the content of a file from Google Drive.
   *
   * @param fileId - The ID of the Drive file.
   * @param signal - Optional {@link AbortSignal} to cancel the request.
   * @returns A promise that resolves to the file content as a Uint8Array.
   */
  async getDriveFileContent(
    fileId: string,
    signal?: AbortSignal,
  ): Promise<Uint8Array> {
    const url = new URL(`${FILES_ENDPOINT}/${fileId}?alt=media`);
    const response = await this.blobFetch(url.toString(), {
      method: 'GET',
      signal,
    });

    return new Uint8Array(await response.arrayBuffer());
  }

  /**
   * Retrieves the name of a file from Google Drive.
   *
   * @param fileId - The ID of the Drive file.
   * @param signal - Optional {@link AbortSignal} to cancel the request.
   * @returns A promise that resolves to the file name.
   */
  async getDriveFileName(
    fileId: string,
    signal?: AbortSignal,
  ): Promise<string> {
    return (await this.getDriveFileMetadata(fileId, signal)).name;
  }

  /**
   * Retrieves the metadata for a file from Google Drive.
   *
   * @param id - The ID of the Drive file.
   * @param signal - Optional {@link AbortSignal} to cancel the request.
   * @returns A promise that resolves to the file metadata.
   */
  private async getDriveFileMetadata(
    id: string,
    signal?: AbortSignal,
  ): Promise<DriveFileMetadata> {
    const url = new URL(`${FILES_ENDPOINT}/${id}`);
    url.searchParams.append('fields', 'name');

    return fetchAndParse(
      this.jsonFetch,
      url.toString(),
      DriveFileMetadataSchema,
      {
        method: 'GET',
        signal,
      },
    );
  }
}
```

src/drive/client.ts
```diff
@@ -0,0 +1,111 @@
...
+    fileId: string,
+    signal?: AbortSignal,
+  ): Promise<Uint8Array> {
+    const url = new URL(`${FILES_ENDPOINT}/${fileId}?alt=media`);
+    const response = await this.authenticatedFetch(url.toString(), {
+      method: 'GET',
+      signal,
+    });
+
+    return new Uint8Array(await response.arrayBuffer());
```
kevineger, line 76: `response.arrayBuffer()` loads the entire file into memory at once, which'll be a memory squeeze for large files. Looking at the [draft PR where this is consumed](https://github.com/googlecolab/colab-vscode/pull/463/changes#diff-5a66dd2281cf7913fb61068fc8a2f27b56b6ab11997214502bed1927e73f9611R40), looks like the current approach is to read and save the full content. I think it may be better to stream the data directly to disk. Something like:

```ts
import { pipeline } from 'stream/promises';
import { Readable } from 'stream';

const driveStream: Readable = await driveClient.getDriveFileStream(id);
await pipeline(driveStream, fs.createWriteStream(targetUri.fsPath));
```

If something like that works, then we'd maybe change `getDriveFileContent` to `getDriveFileStream` and return a `Readable` stream.

---

I haven't looked in detail, so there may be some complications or considerations of Desktop/Remote VS Code vs. Web. Also, Node `fetch` returns a _web_ `ReadableStream` but `fs.createWriteStream` works with `Readable`. So you may need to [`Readable.fromWeb`](https://nodejs.org/api/stream.html#streamreadablefromwebreadablestream-options).

---

I'm fine if we want to get this in and working with the full content read and adapt later. Please just drop a TODO with a summary of the above.

src/drive/client.ts
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * A client for interacting with the Google Drive API
+ */
+export class DriveClient {
+  /**
+   * Creates a new instance of DriveClient.
+   *
+   * @param getAccessToken - Function to retrieve the access token.
+   * @param onAuthError - Callback when an auth error occurs.
+   * @returns A new ColabClient instance.
```
kevineger:
```suggestion
   * @returns A new DriveClient instance.
```

src/drive/client.unit.test.ts
```diff
@@ -0,0 +1,110 @@
...
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  describe('getDriveFileContent', () => {
+    it('should return the file content', async () => {
+      const fileId = 'test-file-id';
+      const fileContent = { cells: [], metadata: {} };
+      const expectedContent = new TextEncoder().encode(
+        JSON.stringify(fileContent),
```
kevineger: Should this be updated since it isn't JSON content?

src/drive/client.ts
mounikasomisetty: Added the TODO

src/drive/client.unit.test.ts
mounikasomisetty: The JSON.stringify there is for the .ipynb file content since it's just a JSON string. I moved things around to make that more clear

src/drive/client.ts
mounikasomisetty: Ah! I guess since it worked when I tested it,  I didn't think it would be an issue. Change made

kevineger (approved):

PR #505, opened 2026-03-20
refactor: client middleware

kevineger:
This pattern is pretty popular for HTTP clients and servers alike. It enables us to have a single place to add cross-cutting concerns like logging, error handling, and retry logic without having to modify the core client logic. It separates concerns nicely, making the code easier to maintain and extend in the future. This refactor enables us to more easily author the `DriveClient` which will leverage a subset of the middleware.


src/extension.ts
```diff
@@ -86,14 +87,14 @@ async function activateInternal(context: vscode.ExtensionContext) {
...
   );
-  const colabClient = new ColabClient(
+  const colabClient = ColabClient.create(
     new URL(CONFIG.ColabApiDomain),
     new URL(CONFIG.ColabGapiDomain),
-    (scopes: readonly string[]) =>
-      GoogleAuthProvider.getOrCreateSession(vscode, scopes).then(
+    { appName: vscode.env.appName, extensionVersion: packageInfo.version },
+    () =>
+      GoogleAuthProvider.getOrCreateSession(vscode, [...REQUIRED_SCOPES]).then(
```
mounikasomisetty: *nit* Do we really need to use the spread operator here?

src/common/fetch-utils.ts
```diff
@@ -0,0 +1,34 @@
+/**
```
mounikasomisetty, line 1: Should any of this be in `src/common`? If so would you prefer the drive stuff to be in its own top level directory?

src/common/middleware.ts
```diff
@@ -0,0 +1,114 @@
+/**
```
mounikasomisetty, line 1: same question about the location of the file

mounikasomisetty (approved):
LGTM for the most part! Just a one question and a nit.

src/colab/client.unit.test.ts
mounikasomisetty: We should revert this back to it's previous state before the auth changes

kevineger: Yep, fair enough!

src/common/fetch-utils.ts
kevineger: Good point, done!

src/common/middleware.ts
```diff
@@ -0,0 +1,114 @@
...
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
hjjackyang, line 5: Nit: Add a blank line below this line?

```suggestion
 */
 
```

src/common/middleware.unit.test.ts
```diff
@@ -0,0 +1,164 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
hjjackyang, line 5: Nit: Same here.

```suggestion
 */
 
```

hjjackyang (approved):
Sorry, it took me a while to wrap my head around this middleware implementation. LGTM!

Issue #326, opened 2025-12-20
Feature Request: Persistent Resource Usage Top Bar


Alvin0412:
## Is your feature request related to a problem?

Yes. Currently, there is no persistent way to monitor the remote server's resource usage (CPU, RAM, GPU, Disk) directly within the VS Code interface. To check if I am about to run out of memory or to see if the GPU is being utilized, I have to interrupt my workflow to run code cells like `!nvidia-smi` and `!free -h`

## Describe the solution you'd like to see

I would like to request a persistent Resource Usage Top Bar similar to the UI found in Kaggle Notebooks.

### Specific requirements:
- Visual Indicators: Small progress bars showing usage for HDD, CPU, RAM, and GPU.
- Placement: Ideally located at the top of the notebook editor or integrated into the VS Code Status Bar so it is always visible.
- Hover Details: Hovering over the indicators should show the specific metrics (e.g., "GPU Memory: 3GB / 15GB" or "Session duration: 53m").
- Status Colors: The bars should change color (e.g., green to red) as resources near their limits.

## Describe alternatives you've considered

Running !nvidia-smi manually in a new cell repeatedly.

## Additional context

The "Draft Session" UI in Kaggle is a perfect reference for this. It is compact, informative, and provides all necessary hardware health checks at a glance.

<img width="667" height="310" alt="Image" src="https://github.com/user-attachments/assets/c4f7b131-56e3-49e2-92bf-01b5592b9982" />

jchwenger:
This would be great, plussing that! I must say, even if it's probably obvious, also having the current accelerator displayed could be good, since sometimes Colab will not assign the desired one, and automatically use another type instead (a warning about that would be good of course, maybe it exists but I haven't seen it yet).

hjjackyang:
📣 We have just released `v0.5.0` today with a resource monitoring experimental feature!

See https://github.com/googlecolab/colab-vscode/wiki/User-Guide#resource-monitor for more info.

There are still a few touch-ups I'd like to do before turning this feature on by default. Meanwhile, I will also keep this issue open to collect feedback. Thanks!

hjjackyang:
📣 We just released `v0.7.3` today with a bunch of stability improvements to resource monitoring view and a new shiny Colab status bar to view your compute usage and balance!

See https://github.com/googlecolab/colab-vscode/wiki/User-Guide#monitoring-usage for more info.

With that, I'm closing this issue. Please file new issues for feedback and bug reports :)

PR #506, opened 2026-03-20
feat: implement `ResourceTreeProvider`

hjjackyang:
This is the *second* of a series of PRs to implement the resources tree view.

E2E screencast preview can be found [here](https://screencast.googleplex.com/cast/NjQyMjQ5MDQ0MzE1MzQwOHwxNGMyN2FkNC00NA).

*Notes*:
* This PR is currently a no-op as the new `ResourceTreeProvider` is not yet registered in the extension. The next PR will register the provider as a new view in the activity bar.
* Similar to `ServerTreeProvider`, I can't seem to write tests in `*.unit.test.ts` for these classes because they extend classes from 'vscode' package, hence tests in `*.vscode.test.ts` (executed as part of `npm run test:integration` command).

---

### DIFF
--- a/src/colab/resource-monitor/resource-tree.ts
+++ b/src/colab/resource-monitor/resource-tree.ts
@@ -55,6 +55,7 @@ export class ResourceTreeProvider
     authChange: Event<AuthChangeEvent>,
     private readonly client: ColabClient,
   ) {
+    // TODO: Handle rapid assignment changes and race conditions
     this.assignmentListener = assignmentChange(this.refresh.bind(this));
     this.authListener = authChange(this.handleAuthChange.bind(this));
   }
@@ -118,6 +119,7 @@ export class ResourceTreeProvider
 
     // If no element is passed (requested at root level), fetch and cache all
     // servers and their resources.
+    // TODO: Handle resources fetch error for single server gracefully
     const servers = await this.assignments.getServers('extension');
     const serverItems: ResourceItem[] = [];
     for (const s of servers) {


src/colab/resource-monitor/resource-item.ts
```diff
@@ -0,0 +1,142 @@
...
+   */
+  static fromDisk(endpoint: string, disk: Disk): ResourceItem {
+    const filesystem = disk.filesystem;
+    let diskSubLabel = '';
+    if (filesystem.label?.length && filesystem.label !== 'kernel') {
+      const diskName = filesystem.label.split('/').pop();
+      if (diskName) {
+        diskSubLabel = ` [ ${diskName} ]`;
+      }
+    }
```
hjjackyang, line 72: Note: This disk label logic is mimicked from http://google3/research/colab/frontend/common/components/resource_usage_display/tab.ts;l=161-163;rcl=846002899.

src/colab/resource-monitor/resource-item.ts
```diff
@@ -0,0 +1,142 @@
...
+   * @returns A {@link ResourceItem} instance representing GPU usage.
+   */
+  static fromGpus(endpoint: string, gpus: GpuInfo[]): ResourceItem {
+    const gpuUsage = gpus.reduce(
+      (acc, gpu) => ({
+        memoryUsedBytes: acc.memoryUsedBytes + gpu.memoryUsedBytes,
+        memoryTotalBytes: acc.memoryTotalBytes + gpu.memoryTotalBytes,
+      }),
+      { memoryUsedBytes: 0, memoryTotalBytes: 0 },
+    );
```
hjjackyang, line 97: Note: This GPU aggregation logic is mimicked from http://google3/research/colab/frontend/common/kernel/kernel_impl.ts;l=1511-1515;rcl=843778100.

src/colab/resource-monitor/resource-tree.ts
```diff
@@ -0,0 +1,154 @@
...
+    if (!this.isAuthorized) {
+      return [];
+    }
+
+    // If the item is a 'server', return corresponding cached resources.
+    if (element?.type === ResourceType.SERVER) {
+      return this.resourceItemsByEndpoint.get(element.endpoint) ?? [];
+    }
+
+    // All resource items other than 'server' have no children.
```
mounikasomisetty: *nit* this may just be personal preference but I like to see everything that return `[]` up top

mounikasomisetty:
Do the `vscode` tests still run when `npm run test:unit` is invoked?

src/colab/resource-monitor/resource-tree.ts
hjjackyang: Good call! I refactored the conditions a bit. Though the `return []` is still not fully at the top, I hope it's a bit more clear and easier to read/understand the conditions now.

hjjackyang:
> Do the `vscode` tests still run when `npm run test:unit` is invoked?

@mounikasomisetty - I don't think so. I use `npm run test:integration` to run these `vscode` tests.

src/colab/resource-monitor/resource-item.ts
```diff
@@ -0,0 +1,142 @@
...
+ */
+export enum ResourceType {
+  // The server itself, shown as the root of the tree.
+  SERVER = 'server',
+  // System RAM resource.
+  RAM = 'ram',
+  // Disk resource.
+  DISK = 'disk',
+  // GPU RAM resource, applicable for GPU accelerators only.
+  GPU = 'gpu',
```
kevineger: Nit: docblocks so these are shown properly with VS Code intellisense.

src/colab/resource-monitor/resource-item.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  /**
+   * Creates a new instance of {@link ResourceItem} representing memory usage.
+   *
+   * @param endpoint - The server endpoint URL.
+   * @param memory - Colab server memory usage information.
+   * @returns A {@link ResourceItem} instance representing memory usage.
+   */
+  static fromMemory(endpoint: string, memory: Memory): ResourceItem {
```
kevineger, line 48: Nit: could we consistently use _ram_ or _memory_ verbiage, but not both?

src/colab/resource-monitor/resource-item.ts
```diff
@@ -0,0 +1,142 @@
...
+    const usedBytes = memory.totalBytes - memory.freeBytes;
+    const tooltip = percentUsedString(usedBytes, memory.totalBytes);
+    const label = `System RAM: ${bytesToGbString(usedBytes)} / ${bytesToGbString(memory.totalBytes)} GB`;
```
kevineger: Nit: local for readability.

```suggestion
    const used = bytesToGbString(usedBytes);
    const total = bytesToGbString(memory.totalBytes);
    const label = `System RAM: ${used} / ${total} GB`;
```

Please consider applying to other long template strings throughout (e.g. in `fromDisk`).

src/colab/resource-monitor/resource-item.ts
```diff
@@ -0,0 +1,142 @@
...
+      this.collapsibleState = TreeItemCollapsibleState.Expanded;
+    }
+  }
+}
+
+function bytesToGbString(bytes: number, precision = 2): string {
+  return (bytes / (1024 * 1024 * 1024)).toFixed(precision);
+}
+
+function percentUsedString(
```
kevineger: Nit: avoid unnecessary type suffix.

```suggestion
function asPercentUsed(
```

src/colab/resource-monitor/resource-tree.vscode.test.ts
```diff
@@ -0,0 +1,214 @@
...
+import { ResourceTreeProvider } from './resource-tree';
+
+const DEFAULT_SERVER = {
+  label: 'Colab GPU A100',
+  endpoint: 'm-s-foo',
+  connectionInformation: {
+    baseUrl: Uri.parse('https://example.com'),
+    token: '123',
+  },
+} as ColabAssignedServer;
```
kevineger, line 29: Can the cast be avoided if you type this `const`?

src/colab/resource-monitor/resource-tree.ts
```diff
@@ -0,0 +1,154 @@
...
+      // Otherwise, return corresponding cached resources.
+      return this.resourceItemsByEndpoint.get(element.endpoint) ?? [];
+    }
+
+    // If no element is passed (requested at root level), fetch and cache all
+    // servers and their resources.
+    const servers = await this.assignments.getServers('extension');
+    const serverItems: ResourceItem[] = [];
+    for (const s of servers) {
+      const resources = await this.client.getResources(s);
```
kevineger, line 126: If a single server fails to report its resources, then the entire `getChildren` call rejects which would prevent the entire tree view from rendering. We should probably catch errors and render an error `ResourceItem` for the failed server(s) (preferred), or omit them?

If you'd rather follow up, feel free to add a TODO.

src/colab/resource-monitor/resource-tree.ts
```diff
@@ -0,0 +1,154 @@
...
+      }
+      // Otherwise, return corresponding cached resources.
+      return this.resourceItemsByEndpoint.get(element.endpoint) ?? [];
+    }
+
+    // If no element is passed (requested at root level), fetch and cache all
+    // servers and their resources.
+    const servers = await this.assignments.getServers('extension');
+    const serverItems: ResourceItem[] = [];
+    for (const s of servers) {
```
kevineger, line 125: A user can only have up to 3 servers, so this isn't a big deal, but we could fire these all off and `Promise.all` them. May also simplify error handling as recommended in other comment.

src/colab/resource-monitor/resource-tree.ts
```diff
@@ -0,0 +1,154 @@
...
+   * @param authChange - The Auth change event.
+   * @param client - The API client instance.
+   */
+  constructor(
+    private readonly assignments: AssignmentManager,
+    assignmentChange: Event<AssignmentChangeEvent>,
+    authChange: Event<AuthChangeEvent>,
+    private readonly client: ColabClient,
+  ) {
+    this.assignmentListener = assignmentChange(this.refresh.bind(this));
```
kevineger, line 59: It's unlikely, but highly possible this introduces race conditions. Rapid consecutive `assignmentChange` events will trigger `refresh` multiple times. This causes VS Code to dispatch multiple concurrent `getChildren` calls. Because there is no concurrency control or cancellation mechanism (e.g. the `LatestCancelable` utility), this will leak overlapping network requests and cause race conditions when blindly overwriting the `this.resourceItemsByEndpoint` map.

Please address or add a TODO to follow-up.

We should also double check if this needs to be addressed in `src/colab/server-browser/server-tree.ts`.

src/colab/resource-monitor/resource-item.ts
```diff
@@ -0,0 +1,142 @@
...
+   * @returns A {@link ResourceItem} instance representing disk usage.
+   */
+  static fromDisk(endpoint: string, disk: Disk): ResourceItem {
+    const filesystem = disk.filesystem;
+    let diskSubLabel = '';
+    if (filesystem.label?.length && filesystem.label !== 'kernel') {
+      const diskName = filesystem.label.split('/').pop();
+      if (diskName) {
+        diskSubLabel = ` [ ${diskName} ]`;
+      }
```
kevineger, line 71: Sorry if misreading, but if `filesystem.label` is the root directory "/", the expression `filesystem.label.split('/').pop()` will be an empty string. This evaluates to falsy, so the root partition woulid just be "Disk:", no?

src/colab/resource-monitor/resource-item.ts
hjjackyang: Done!

src/colab/resource-monitor/resource-item.ts
hjjackyang: Since the backend proto named this `Memory`, I kept the *memory* verbiage instead of *ram*.

src/colab/resource-monitor/resource-item.ts
hjjackyang: Done.

src/colab/resource-monitor/resource-item.ts
hjjackyang: Done.

src/colab/resource-monitor/resource-tree.vscode.test.ts
hjjackyang: It can be, but then I'd have to initialize all non-optional fields of `ColabAssignedServer` which I don't care about in this test. I'd prefer to keep what I have to avoid initializing unrelated fields.

src/colab/resource-monitor/resource-tree.ts
hjjackyang: Good call! Added a TODO to follow up on this!

src/colab/resource-monitor/resource-tree.ts
hjjackyang: Will address this in a follow-up with the error handling.

src/colab/resource-monitor/resource-tree.ts
hjjackyang: Added a TODO for this.

src/colab/resource-monitor/resource-item.ts
hjjackyang: Great catch! I fixed the logic to now render `'Disk [  ]:'` for root directory. Also added a test case for this.

PR #507, opened 2026-03-20
feat: register `ResourceTreeProvider` in extension

hjjackyang:
This is the *third* of a series of PRs to implement the resources tree view behind a new experimental config.

E2E screencast preview can be found [here](https://screencast.googleplex.com/cast/NjQyMjQ5MDQ0MzE1MzQwOHwxNGMyN2FkNC00NA).

For non-Google users, here is a static preview of the new resources tree view upon releasing this PR:
image

*Notes*:
* This PR will deliver the first version of resources tree view which only reacts to auth and assignment changes, and can only be manually refreshed at the moment.
* Because we now have 2 views in the same Activity Bar view container, I also renamed the existing config and view names to differentiate from the new ones, and to make things more clear.
* The next PR will add periodic polling to refresh the resources tree view at a fixed interval.

---

### DIFF
--- a/src/colab/resource-monitor/resource-item.ts
+++ b/src/colab/resource-monitor/resource-item.ts
@@ -12,13 +12,13 @@ import { Disk, GpuInfo, Memory } from '../api';
  * Types of resources that can be displayed in resource monitor tree view.
  */
 export enum ResourceType {
-  // The server itself, shown as the root of the tree.
+  /** The server itself, shown as the root of the tree. */
   SERVER = 'server',
-  // System RAM resource.
-  RAM = 'ram',
-  // Disk resource.
+  /** System RAM resource. */
+  MEMORY = 'memory',
+  /** Disk resource. */
   DISK = 'disk',
-  // GPU RAM resource, applicable for GPU accelerators only.
+  /** GPU RAM resource, applicable for GPU accelerators only. */
   GPU = 'gpu',
 }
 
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -203,7 +222,8 @@ function watchConsumption(colab: ColabClient): {
 function registerCommands(
   authProvider: GoogleAuthProvider,
   assignmentManager: AssignmentManager,
-  serverTreeProvider: ServerTreeProvider,
+  contentTreeProvider: ServerTreeProvider,
```
amtoney524, line 225: nit/opt: worth renaming `ServerTreeProvider` to sth like `ContentTreeProvider` to help distinguish as you're doing with the view ids / variables?

hjjackyang: Good call! I will defer this to a separate refactoring PR since there are a few relevant files to be renamed and references to update.

package.json
```diff
@@ -72,11 +72,19 @@
...
-        "colab.activityBar": {
+        "colab.activityBarContentView": {
           "type": "boolean",
           "default": true,
           "markdownDescription": "Enables the Colab activity bar where you can view servers and interact with their `/content` directories."
         },
+        "colab.activityBarResourceView": {
+          "type": "boolean",
+          "default": false,
+          "description": "Enables the Colab activity bar where you can monitor server resources.",
```
kevineger: Nit: in the context of VS Code settings, _resources_ is ambiguous.

```suggestion
          "description": "Enables the Colab activity bar where you can monitor server resources (RAM and disk usage).",
```

PR #508, opened 2026-03-23
feat: implement refresh polling in `ResourceTreeProvider`

hjjackyang:
This is the *fourth* of a series of PRs to implement the resources tree view.

In my previous [PR](https://github.com/googlecolab/colab-vscode/pull/507), I registered the resource monitor with only manual refresh capability. This PR adds the capability to poll resource updates every 10s.

E2E screencast preview for refresh polling can be found [here](https://screencast.googleplex.com/cast/NTI0MjQ2ODgxNDU1MzA4OHw5NzY4ZGQ1Mi00MA).

*Note*: The 10s poll interval config is read from [Colab web development flags](http://google3/googledata/experiments/colab/features/development_flags.gcl;l=461;rcl=886858846).

---

### DIFF
--- a/src/colab/resource-monitor/resource-tree.ts
+++ b/src/colab/resource-monitor/resource-tree.ts
@@ -60,9 +60,7 @@ export class ResourceTreeProvider
     private readonly client: ColabClient,
   ) {
     // TODO: Handle rapid assignment changes and race conditions
-    this.assignmentListener = assignmentChange(
-      this.refresh.bind(this, undefined),
-    );
+    this.assignmentListener = assignmentChange(this.refresh.bind(this));
     this.authListener = authChange(this.handleAuthChange.bind(this));
 
     // Read poll interval from experiment config once at runner initialization.
@@ -70,14 +68,17 @@ export class ResourceTreeProvider
     if (typeof refreshIntervalMs === 'number') {
       this.refreshRunner = new SequentialTaskRunner(
         {
-          intervalTimeoutMs: refreshIntervalMs,
+          intervalTimeoutMs: Math.max(
+            refreshIntervalMs,
+            MIN_REFRESH_INTERVAL_MS,
+          ),
           taskTimeoutMs: REFRESH_TIMEOUT_MS,
           abandonGraceMs: 0, // Nothing to cleanup, abandon immediately.
         },
         {
           name: ResourceTreeProvider.name,
-          run: (signal: AbortSignal) => {
-            this.refresh.call(this, signal);
+          run: () => {
+            this.refresh.call(this);
             return Promise.resolve();
           },
         },
@@ -102,13 +103,8 @@ export class ResourceTreeProvider
 
   /**
    * Refreshes the tree view, optionally for specific items.
-   *
-   * @param signal - Optional signal to abort the refresh if set.
    */
-  refresh(signal?: AbortSignal): void {
-    if (signal?.aborted) {
-      return;
-    }
+  refresh(): void {
     this.guardDisposed();
     this.resourceItemsByEndpoint.clear();
     this.changeEmitter.fire(undefined);
@@ -193,4 +189,5 @@ export class ResourceTreeProvider
 }
 
 // The refresh() call should be instant, but giving it a 2s timeout to be safe.
-const REFRESH_TIMEOUT_MS = 2 * 1000; // 2 seconds.
+const REFRESH_TIMEOUT_MS = 2000; // 2 seconds.
+const MIN_REFRESH_INTERVAL_MS = 5000; // 5 seconds.
--- a/src/colab/resource-monitor/resource-tree.vscode.test.ts
+++ b/src/colab/resource-monitor/resource-tree.vscode.test.ts
@@ -221,6 +221,17 @@ describe('ResourceTreeProvider', () => {
     });
   });
 
+  describe('refresh', () => {
+    it('fires an undefined change event', () => {
+      const changeSpy = sinon.spy();
+      tree.onDidChangeTreeData(changeSpy);
+
+      tree.refresh();
+
+      sinon.assert.calledOnceWithExactly(changeSpy, undefined);
+    });
+  });
+
   describe('refresh polling', () => {
     let clock: SinonFakeTimers;
     let refreshSpy: sinon.SinonSpy;


src/colab/resource-monitor/resource-tree.ts
```diff
@@ -57,6 +60,7 @@ export class ResourceTreeProvider
   ) {
     this.assignmentListener = assignmentChange(this.refresh.bind(this));
     this.authListener = authChange(this.handleAuthChange.bind(this));
+    this.ensureRefreshPolling();
```
kevineger: Is this a no-op? `this.isAuthorized` is initialized to false so wouldn't `ensureRefreshPolling` immediately return due to `!this.isAuthorized`?

src/colab/resource-monitor/resource-tree.ts
```diff
@@ -142,6 +147,32 @@ export class ResourceTreeProvider
...
+  private ensureRefreshPolling() {
+    if (this.isDisposed || !this.isAuthorized || this.refreshInterval) {
+      return;
+    }
+
+    const refreshIntervalMs = getFlag(ExperimentFlag.ResourcePollIntervalMs);
+    if (typeof refreshIntervalMs !== 'number') {
+      return;
+    }
+    this.refreshInterval = setInterval(() => {
```
kevineger: Should we consider using `SequentialTaskRunner` with `StartMode` and `OverrunPolicy`) which accounts for overlapping execution and handle overruns gracefully? Using a bare `setInterval` will repeatedly invalidate the cache and fire a UI refresh (which triggers `getChildren` and network calls) even if previous requests are still pending or hung.

src/colab/resource-monitor/resource-tree.ts
```diff
@@ -142,6 +147,32 @@ export class ResourceTreeProvider
...
     this.refresh();
+    if (this.isAuthorized) {
+      this.ensureRefreshPolling();
+    } else {
+      this.clearRefreshPolling();
+    }
+  }
+
+  private ensureRefreshPolling() {
+    if (this.isDisposed || !this.isAuthorized || this.refreshInterval) {
```
kevineger: I think it's fine, but maybe worth adding a comment that we don't respect changes to `ExperimentFlag.ResourcePollIntervalMs`. In other words, it's static.

hjjackyang:
> Should we remove the manual mechanism?

I'm going to leave it in for now just in case the polling isn't working or users are inpatient to wait for poll 😛

src/colab/resource-monitor/resource-tree.ts
hjjackyang: Good call!

src/colab/resource-monitor/resource-tree.ts
hjjackyang: Implemented the `SequentialTaskRunner` approach. Thanks for the tips!

src/colab/resource-monitor/resource-tree.ts
hjjackyang: Added a comment to call this out.

src/colab/resource-monitor/resource-tree.ts
```diff
@@ -58,9 +59,31 @@ export class ResourceTreeProvider
...
-    this.assignmentListener = assignmentChange(this.refresh.bind(this));
+    // TODO: Handle rapid assignment changes and race conditions
+    this.assignmentListener = assignmentChange(
+      this.refresh.bind(this, undefined),
+    );
     this.authListener = authChange(this.handleAuthChange.bind(this));
-    this.ensureRefreshPolling();
+
+    // Read poll interval from experiment config once at runner initialization.
+    const refreshIntervalMs = getFlag(ExperimentFlag.ResourcePollIntervalMs);
```
kevineger, line 67: Should we clamp this at a _minimum_? Avoid the (hopefully unlikely) case where the flag gets configured with `0`, or some low number and we effectively spam ourselves? Something like `Math.max(refreshIntervalMs, 5000)`.

src/colab/resource-monitor/resource-tree.ts
```diff
@@ -79,8 +102,13 @@ export class ResourceTreeProvider
...
   /**
    * Refreshes the tree view, optionally for specific items.
+   *
+   * @param signal - Optional signal to abort the refresh if set.
    */
-  refresh(): void {
+  refresh(signal?: AbortSignal): void {
+    if (signal?.aborted) {
+      return;
+    }
```
kevineger: Why accept an `AbortSignal`? With no async/await here, there's no opportunity for the `signal` to be aborted between when it's called and when the body excecutes.

It also doesn't look like there's a call-path which will set this?

src/colab/resource-monitor/resource-tree.vscode.test.ts
```diff
@@ -214,22 +222,15 @@ describe('ResourceTreeProvider', () => {
...
     let clock: SinonFakeTimers;
     let refreshSpy: sinon.SinonSpy;
 
     beforeEach(() => {
-      FLAGS_TEST_ONLY.setFlagForTest(
-        ExperimentFlag.ResourcePollIntervalMs,
-        TEST_RESOURCE_POLL_INTERVAL_MS,
-      );
       clock = sinon.useFakeTimers();
       refreshSpy = sinon.spy(tree, 'refresh');
```
kevineger, line 241: Why not test at the observable public API behaviour? Subscribe to the event which should fire?

src/colab/resource-monitor/resource-tree.ts
hjjackyang: > It also doesn't look like there's a call-path which will set this?

Oh, I added this because the `SequentialTaskRunner`'s `task.run` signature was `(signal: AbortSignal) => Promise<void>`.

> With no async/await here, there's no opportunity for the signal to be aborted between when it's called and when the body executes.

You are right! Removed the unnecessary `AbortSignal` from `refresh()` method.

src/colab/resource-monitor/resource-tree.ts
hjjackyang: Good call! Added a `MIN_REFRESH_INTERVAL_MS` of 5 sec as a safeguard.

src/colab/resource-monitor/resource-tree.vscode.test.ts
hjjackyang: In these polling tests, I think it's more readable to spy on the `refresh()` method than spying on the underlying change event because what's important in these tests are that polling happens as expected.

But I do agree that we should test the underlying refresh behavior, so added a test for `refresh()` method itself.

Issue #502, opened 2026-03-02
Does not fallback to a similar server when exact type is unavailable.


MarsDenizen:
**Describe the current behavior**
Given a Colab Pro User logged in VSCode, the Google Colab VS Code extension always fails to connect to an H100 GPU runtime, returning a 503 Service Unavailable error. While the Google Colab web browser successfully connects to an H100 instantly or after few  automated retry attempts, the VS Code extension fails "silently" or throws a terminal error without attempting to retry the connection.

**Describe the expected behavior**
The extension should implement the same exponential backoff/retry logic as the web browser to successfully provision high-demand resources like the H100.

**What web browser you are using**
Chrome Version 145.0.7632.117 (Official Build) (64-bit)

**Additional context**
The connection to CPU and L4 GPU runtimes works without issue. The H100 connects in the browser on the same network/account, confirming this is an extension-specific provisioning failure.
As a paid subscriber, this lack of parity between the web and VS Code environments is preventing use of the H100 for complex development tasks. The relevant error log from the exthost log has been attached here.
[exthostlog.txt](https://github.com/user-attachments/files/25684624/exthostlog.txt)

kevineger:
Thanks for the report @MarsDenizen - that sounds like a bug. Looking into it.

kevineger:
Took a look, bit of a _good news, bad news_ situation.

First, the good news - I'll get a fix out to match the behaviour in the web UI.

The bad news, if you're clicking H100 in the web UI, waiting for a minute and then it _connects_, you likely aren't connecting to an H100. Double check the _Connect_ button/dialog in the top right to check the machine type. Those machines are super limited in our fleet and getting one is pretty rare. The web UI keeps the request open for a minute in case it can be satisfied, then it has fallback logic to connect to a _similar_ machine.

MarsDenizen:
@kevineger Thanks for taking a look—really appreciate it. The fix to bring VS Code in line with the web UI behavior is great news and should unblock my workflow.

The fallback behavior in the web UI actually works well for me—getting a similar machine is predictable and usable, even if it’s not always an H100. The main issue was the lack of retry/fallback in VS Code causing hard failures, so this approach sounds good.

Happy to test once the fix is available or share any additional logs if needed.

kevineger:
Just pushed `	0.5.0` which includes https://github.com/googlecolab/colab-vscode/pull/511.

From PR description:

[...]

PR #511, opened 2026-03-25
feat: fallback to available accelerator(s)

kevineger:
https://github.com/googlecolab/colab-vscode/issues/502

This mirrors Colab (web) behaviour; when the requested accelerator is unavailable, fallback to one that is.

Admittedly, this leaves the UI in a bit of a weird, seemingly "hanging" state, where the kernel selector is just spinning as we wait a while for the server and fallback (up to a minute per accelerator). We should consider popping a progress notification on assignment, or at least calls that take over `n` seconds? I'd prefer to get this functional fix in before enriching the UX with something like that.

image

### DIFF
--- a/src/jupyter/assignments.ts
+++ b/src/jupyter/assignments.ts
@@ -622,26 +622,26 @@ export class AssignmentManager implements Disposable {
       if (!(error instanceof AcceleratorUnavailableError)) {
         throw error;
       }
+      let newFallback: typeof fallback;
       // The initial attempt failed, start falling back.
       if (!fallback) {
-        const all = await this.getAvailableServerDescriptors();
-        const toAttemptSet = new Set<string>();
+        const all = await this.getAvailableServerDescriptors(signal);
+        const toAttempt = new Set<string>();
         for (const d of all) {
           if (
             d.variant === variant &&
             d.accelerator &&
             d.accelerator !== accelerator
           ) {
-            toAttemptSet.add(d.accelerator);
+            toAttempt.add(d.accelerator);
           }
         }
-        const toAttempt = Array.from(toAttemptSet);
-        fallback = {
-          toAttempt,
+        newFallback = {
+          toAttempt: Array.from(toAttempt),
           attempted: [accelerator],
         };
       } else {
-        fallback = {
+        newFallback = {
           toAttempt: fallback.toAttempt.slice(1),
           attempted: [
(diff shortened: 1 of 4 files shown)


src/jupyter/assignments.ts
```diff
@@ -560,6 +569,77 @@ export class AssignmentManager implements vscode.Disposable {
...
+        );
+      }
+      return assignment;
+    } catch (error) {
+      if (!(error instanceof AcceleratorUnavailableError)) {
+        throw error;
+      }
+      // The initial attempt failed, start falling back.
+      if (!fallback) {
+        const all = await this.getAvailableServerDescriptors();
```
hjjackyang: Nit: Pass the `AbortSignal` here?

```suggestion
        const all = await this.getAvailableServerDescriptors(signal);
```

src/jupyter/assignments.ts
```diff
@@ -560,6 +569,77 @@ export class AssignmentManager implements vscode.Disposable {
...
+          if (
+            d.variant === variant &&
+            d.accelerator &&
+            d.accelerator !== accelerator
+          ) {
+            toAttemptSet.add(d.accelerator);
+          }
+        }
+        const toAttempt = Array.from(toAttemptSet);
+        fallback = {
```
hjjackyang: Nit: I personally prefer not mutating existing variables, making the code easier to read and reason with. Let's make a new `updatedFallback` variable instead.

```suggestion
        const updatedFallback = {
```

src/jupyter/assignments.ts
```diff
@@ -560,6 +569,77 @@ export class AssignmentManager implements vscode.Disposable {
...
+          toAttempt,
```
hjjackyang: Nit: This `toAttempt` variable is only used once. Perhaps cleaner to just inline it?

```suggestion
        fallback = {
          toAttempt: Array.from(toAttemptSet),
```

hjjackyang (approved):

src/jupyter/assignments.ts
kevineger: Done

src/jupyter/assignments.ts
kevineger: Done

src/jupyter/assignments.ts
kevineger: Done

PR #531, opened 2026-04-03
fix: join colab: URI paths with POSIX separators on Windows

dhunganapramod9:
# fix: join `colab:` URI paths with POSIX separators on Windows

## Summary

Adds `joinUriPath()` in `src/colab/files.ts` and uses **`path.posix.join`** with `uri.with({ path })` instead of `vscode.Uri.joinPath()` for Colab-related URIs. `Uri.joinPath()` follows OS path rules and can insert **backslashes** on Windows; `colab:` paths must use **forward slashes** or uploads, the content browser, tree items, and the mounted Jupyter contents provider break (including unit tests on Windows).

## Problem

- `Uri.joinPath()` is **platform-dependent**.
- On Windows, `uri.path` can contain backslashes (e.g. `colab://endpoint\content\file.txt` instead of `colab://endpoint/content/file.txt`).
- That breaks comparisons, Jupyter path usage, and tests that expect `/`-separated paths.

## Solution

- **`joinUriPath(uri, ...segments)`** — normalizes `\` → `/` per segment, then `path.posix.join(uri.path, ...segments)`, returns `uri.with({ path })`.
- **`buildColabFileUri()`** — uses `joinUriPath` for relative paths.
- **Updated call sites:** `src/colab/commands/files.ts`, `src/colab/content-browser/commands.ts`, `src/colab/content-browser/content-tree.ts`, `src/jupyter/contents/file-system.ts`.
- **Unchanged:** `Uri.joinPath` for normal workspace / `file:` URIs (e.g. `src/drive/commands/import.ts`).

## Tests

- `src/colab/files.unit.test.ts` — `joinUriPath` (join, backslash normalization, `..`).
- Full unit suite passes locally.

## Verification (local)

`npm run build`, `npm run lint`, `npm run typecheck`, `npm run test:unit`.

CI (`format:check`, spell check, etc.) is authoritative on Linux.

## Note for reviewers

Please sanity check that new or changed `colab:` paths in these flows use **only `/`** in `.path`. This PR only changes URI path construction, not the Colab or Jupyter APIs.

### DIFF
--- a/src/colab/commands/files.ts
+++ b/src/colab/commands/files.ts
@@ -8,7 +8,7 @@ import vscode, { QuickPickItem } from 'vscode';
 import { log } from '../../common/logging';
 import { AssignmentManager } from '../../jupyter/assignments';
 import { ColabAssignedServer } from '../../jupyter/servers';
-import { buildColabFileUri } from '../files';
+import { buildColabFileUri, joinUriPath } from '../files';
 import { UPLOAD } from './constants';
 
 /**
@@ -149,8 +149,8 @@ async function buildUploadPlan(
       for (const [name] of entries) {
         try {
           await planUploadsRecursively(
-            vs.Uri.joinPath(source, name),
-            vs.Uri.joinPath(dest, name),
+            joinUriPath(source, name),
+            joinUriPath(dest, name),
             operations,
           );
         } catch (err) {
@@ -169,7 +169,7 @@ async function buildUploadPlan(
       failCount++;
       continue;
     }
-    const destUri = vs.Uri.joinPath(serverRootUri, itemName);
+    const destUri = joinUriPath(serverRootUri, itemName);
     try {
       await planUploadsRecursively(inputUri, destUri, operations);
     } catch (err) {
--- a/src/colab/content-browser/commands.ts
+++ b/src/colab/content-browser/commands.ts
@@ -5,6 +5,7 @@
  */
 
 import vscode, { Uri } from 'vscode';
+import { joinUriPath } from '../files';
 import type { ContentItem } from './content-item';
 
 /**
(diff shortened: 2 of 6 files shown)


hjjackyang:
@dhunganapramod9 - Do you see any particular issue interacting with files on Windows today? I just sanity tested file interactions via the latest extension on my Windows laptop and everything seemed to be working fine. So I'm hesitant to approve this PR.

cc @kevineger

dhunganapramod9:
@hjjackyang I dug into this more carefully and I think my original framing overstated the runtime impact.
The Windows failures I was seeing appear to come from the test helper implementation rather than from real VS Code URI behavior. In particular, the test `TestUri.joinPath()` uses Node’s path joining, which introduces backslashes on Windows, while real URI joining appears to stay POSIX-style.
So I don’t think this is a strong runtime bug report as I originally presented it. Thanks for pushing on that. I’m happy to close this out, or follow up separately with a narrower test-helper fix if that would be useful.

kevineger:
> I’m happy to close this out, or follow up separately with a narrower test-helper fix

If there's a bug with [src/test/helpers/uri.ts](https://github.com/googlecolab/colab-vscode/blob/main/src/test/helpers/uri.ts) preventing tests from running on Windows, happy to re-purpose this PR to review a fix.

kevineger:
> > I’m happy to close this out, or follow up separately with a narrower test-helper fix
> 
> If there's a bug with [src/test/helpers/uri.ts](https://github.com/googlecolab/colab-vscode/blob/main/src/test/helpers/uri.ts) preventing tests from running on Windows, happy to re-purpose this PR to review a fix.

Going to close this PR but please feel free to open another if there's a test fix needed for Windows.

PR #540, opened 2026-04-07
refactor: `ExperimentStateProvider` to use `SequentialTaskRunner`

hjjackyang:
**Why**? We already have a very robust `setInterval` abstraction implemented in `SequentialTaskRunner` which properly handles task timeout and overrun. This PR updates `ExperimentStateProvider` to leverage `SequentialTaskRunner` for flags polling and greatly simplifies the code.

This PR should be a no-op other than a more robust overrun handling.

[Preview screencast](https://screencast.googleplex.com/cast/NDgzMjYyMzE3MDk0NTAyNHw1MjE5MGM4OC0wZA)

### DIFF
--- a/src/colab/experiment-state.ts
+++ b/src/colab/experiment-state.ts
@@ -59,7 +59,6 @@ export class ExperimentStateProvider implements Toggleable, Disposable {
       },
       OverrunPolicy.AbandonAndRun,
     );
-    this.refreshPoller.start();
   }
 
   /**
@@ -79,7 +78,7 @@ export class ExperimentStateProvider implements Toggleable, Disposable {
   on(): void {
     this.guardDisposed();
     this.isAuthorized = true;
-    this.refreshPoller.runNow();
+    this.ensurePollingAndRunOnce();
   }
 
   /**
@@ -88,6 +87,11 @@ export class ExperimentStateProvider implements Toggleable, Disposable {
   off(): void {
     this.guardDisposed();
     this.isAuthorized = false;
+    this.ensurePollingAndRunOnce();
+  }
+
+  private ensurePollingAndRunOnce(): void {
+    this.refreshPoller.start();
     this.refreshPoller.runNow();
   }
 
--- a/src/colab/experiment-state.ts
+++ b/src/colab/experiment-state.ts
@@ -110,7 +110,10 @@ export class ExperimentStateProvider implements Toggleable, Disposable {
       );
       if (result.experiments) {
         flags = result.experiments;
-        log.trace('Experiment state updated:', Object.fromEntries(flags));
+        log.trace(
+          `Experiment state updated while ${requireAccessToken ? 'authorized' : 'not authorized'}:`,
+          Object.fromEntries(flags),
+        );
       }
     } catch (e: unknown) {
       log.error('Failed to update experiment state:', e);
--- a/src/colab/experiment-state.ts
+++ b/src/colab/experiment-state.ts
@@ -6,7 +6,8 @@
 
 import { Disposable } from 'vscode';
 import { log } from '../common/logging';
-import { AsyncToggle } from '../common/toggleable';
+import { OverrunPolicy, SequentialTaskRunner } from '../common/task-runner';
+import { Toggleable } from '../common/toggleable';
 import {
   ExperimentFlag,
   ExperimentFlagValue,
@@ -25,13 +26,14 @@ export function getFlag(flag: ExperimentFlag): ExperimentFlagValue {
 }
 
 const REFRESH_INTERVAL_MS = 5 * 60 * 1000; // 5 minutes.
+const REFRESH_TIMEOUT_MS = 30 * 1000; // 30 seconds.
 
 /**
  * Provides experiment state information from the Colab backend.
  */
-export class ExperimentStateProvider extends AsyncToggle implements Disposable {
+export class ExperimentStateProvider implements Toggleable, Disposable {
+  private readonly refreshPoller: SequentialTaskRunner;
   private isAuthorized = false;
-  private refreshInterval?: NodeJS.Timeout;
   private isDisposed = false;
 
   /**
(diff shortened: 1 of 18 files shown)


src/colab/experiment-state.ts
```diff
@@ -40,10 +42,24 @@ export class ExperimentStateProvider extends AsyncToggle implements Disposable {
...
+      },
+      {
+        name: ExperimentStateProvider.name,
+        run: async (signal: AbortSignal) => {
+          await this.getExperimentState(this.isAuthorized, signal);
+        },
+      },
+      OverrunPolicy.AbandonAndRun,
+    );
+    this.refreshPoller.start();
```
kevineger: Previously, `ExperimentStateProvider` only started polling after `on()` was called. Now, `this.refreshPoller.start()` is called in the constructor, meaning polling begins before the user is authorized. Additionally, `off()` no longer stops polling, it just triggers an immediate fetch and flips the `isAuthorized` flag, but the interval keeps running. This could result in unnecessary / not user initiated auth flows triggering.

kevineger (changes requested):

src/colab/experiment-state.ts
hjjackyang: Actually, if you look more closely at the previous implementation, `off()` actually does *not* stop polling. Instead, it just *ensures* polling and also immediately polls once with `isAuthorized` as `false`:

https://github.com/googlecolab/colab-vscode/blob/29c25421138a73237a7fa01cff967863963879fa/src/colab/experiment-state.ts#L98-L99

The only diff with my new implementation is that the polling is now scheduled at construction time instead of on `on()` or `off()`. However, since the poll interval is 5 min, it's almost guaranteed that `on()` or `off()` will be called before the first scheduled poll kicks in.

mounikasomisetty: I agree with @hjjackyang - seems like functionally this is the same.

src/colab/experiment-state.unit.test.ts
```diff
@@ -75,10 +68,15 @@ describe('ExperimentStateProvider', () => {
 
   it('fetches experiment state without auth when turned off', async () => {
     const experiments = new Map([[ExperimentFlag.RuntimeVersionNames, false]]);
-    colabClientStub.getExperimentState.resolves({ experiments });
+    const runGetExperimentState = new Deferred<void>();
```
mounikasomisetty: *nit* Possible to move this to a setup method or something? seems like it's duplicated a couple of times

mounikasomisetty (approved):

src/colab/experiment-state.unit.test.ts
hjjackyang: Done!

src/colab/experiment-state.ts
hjjackyang: Thanks for confirming, @mounikasomisetty! I also updated the PR description to call this out.

kevineger: The thing I'm struggling with is that while yes, in practice it's functionally the same (the 5-minute `StartMode.Scheduled` delay means `whileAuthorized()` will call `on()`/`off()` well before the first scheduled poll fires), it makes assumptions based on the only existing caller.

But what if `whileAuthorized` changes it's impl, or there's another call-path? I think it would make more sense to not schedule during construction and instead snap to the interface of reacting to on/off as the means of turning the polling on and off.

src/colab/experiment-state.unit.test.ts
```diff
@@ -87,58 +79,58 @@ describe('ExperimentStateProvider', () => {
...
+    const getExperimentStateRanPromise = stubGetExperimentStateResponse(
+      // Ensure flags are empty
+      new Map(),
+    );
 
-    expect(getFlag(ExperimentFlag.RuntimeVersionNames)).to.deep.equal([]);
-  });
+    provider.on();
 
-  it('updates flags when state changes', async () => {
```
kevineger, line 107: While individual on/off cases are covered, there's no test verifying that flags transition from one value to another within the same provider instance.

src/colab/experiment-state.unit.test.ts
```diff
@@ -87,58 +79,58 @@ describe('ExperimentStateProvider', () => {
...
-  it('handles errors when fetching experiment state', async () => {
+  it('handles errors when fetching experiment state', () => {
     colabClientStub.getExperimentState.rejects(new Error('Network error'));
 
     // Should not throw
-    await provider.turnOn(new AbortController().signal);
+    provider.on();
 
     sinon.assert.calledOnce(colabClientStub.getExperimentState);
   });
```
kevineger: Since `on()` returns before the rejection is processed (the async work happens in a floating promise via void `this.runWorker()`), the test no longer fully verifies error handling. Please use deferred async hooks to validate.

src/colab/experiment-state.ts
```diff
@@ -53,50 +69,26 @@ export class ExperimentStateProvider extends AsyncToggle implements Disposable {
     if (this.isDisposed) {
       return;
     }
-
+    this.refreshPoller.dispose();
     this.isDisposed = true;
```
kevineger, line 71: Nit: this isn't a bug, just a style thing. Consider putting `this.isDisposed = true;` right after the re-entry guard.

kevineger (approved):
Thanks fro the refactor! Sorry it took me a bit to get back to this one.

src/colab/experiment-state.ts
hjjackyang: Wouldn't it be safer if we set `this.isDisposed = true` in the end to ensure everything is actually properly disposed?

src/colab/experiment-state.ts
hjjackyang: Done. I moved `this.refreshPoller.start()` to `on()`/`off()` methods to make this change truly a no-op.

src/colab/experiment-state.unit.test.ts
hjjackyang: Oh, I didn't think it was too important to cover this since it was covered in several individual test cases. Added the test case back.

src/colab/experiment-state.unit.test.ts
hjjackyang: Good catch! Done.

src/colab/experiment-state.ts
hjjackyang: As discussed offline, moved `this.isDisposed = true` according to this [pattern](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#implement-the-dispose-pattern).

PR #548, opened 2026-04-09
fix: event when server isn't found when resolving

kevineger:
Made `resolveJupyterServer` `async` and added a try/catch when `refreshConnection` throws a `NotFoundError`, the method fires `onDidChangeServers` (to notify the Jupyter extension that the server list has changed) and returns the original server as-is (per the Jupyter API contract: _It is OK to return the given server_). Other errors are re-thrown.

### DIFF
--- a/src/jupyter/provider.ts
+++ b/src/jupyter/provider.ts
@@ -16,7 +16,7 @@ import { CancellationToken, Disposable, Event, ProviderResult } from 'vscode';
 import vscode from 'vscode';
 import { AuthChangeEvent } from '../auth/auth-provider';
 import { SubscriptionTier } from '../colab/api';
-import { ColabClient } from '../colab/client';
+import { ColabClient, NotFoundError } from '../colab/client';
 import {
   AUTO_CONNECT,
   Command,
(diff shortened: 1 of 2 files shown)


hjjackyang (approved):
Nit: Could you add the particular error you are trying to tackle in the PR description for easier reference?

kevineger:
> Nit: Could you add the particular error you are trying to tackle in the PR description for easier reference?

Here's an example stack:

```txt
Error: Server is not assigned at z1.refreshConnection (/Users/[foo]/.vscode/extensions/google.colab-0.7.3/out/extension.js:82:46618) at un.getJupyterServerUri (/Users/[foo]/.vscode/extensions/ms-toolsai.jupyter-2025.9.1-darwin-arm64/dist/extension.node.js:181:27520) at un.createConnectionInfo (/Users/[foo]/.vscode/extensions/ms-toolsai.jupyter-2025.9.1-darwin-arm64/dist/extension.node.js:181:25611) at gu.create (/Users/[foo]/.vscode/extensions/ms-toolsai.jupyter-2025.9.1-darwin-arm64
```

PR #556, opened 2026-04-14
test: use POSIX join semantics in TestUri.joinPath

dhunganapramod9:
## Summary

- update `TestUri.joinPath()` to use `path/posix` join semantics instead of OS path semantics
- normalize `\` to `/` in path segments before joining
- add focused regression tests for forward-slash joining, backslash normalization, and `..` traversal

## Context

This change is intentionally limited to the test helper and tests.

It aligns `src/test/helpers/uri.ts` with URI/POSIX path behavior and mirrors what `vscode-uri` does. It does not make any runtime/product behavior claim.

## Verification

- `npm run test:unit`

### DIFF
--- a/src/test/helpers/uri.ts
+++ b/src/test/helpers/uri.ts
@@ -6,7 +6,7 @@
 
 /* eslint-disable jsdoc/require-jsdoc */
 
-import { join } from 'path';
+import { join } from 'path/posix';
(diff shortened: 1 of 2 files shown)


kevineger (approved):
Thanks 🙂

dhunganapramod9:
Thanks @kevineger

PR #557, opened 2026-04-15
test: resource tree view e2e

hjjackyang:
**Why**? Add E2E test to cover the latest resource tree view feature before turning it on by default.


kevineger (approved):
Beautiful!

Reminds me I should add an e2e test for file uploading. I'll try and get to that this week or next.

PR #558, opened 2026-04-15
fix: preserve Request headers in colabProxyFetch

dipeshbabu:
## Summary                                                                                                             
                                                                                                                         
  Fix `colabProxyFetch(...)` so it preserves headers already present on a passed `Request` object before injecting the   
  Colab proxy headers.

  Previously, when the caller passed a `Request`, the wrapper rebuilt headers from `init.headers` only. That dropped any 
  headers already attached to the original request.

  ## What changed

  - Merge existing `Request` headers into the outgoing header set
  - Overlay any headers from `init.headers`
  - Continue setting:
    - `X-Colab-Runtime-Proxy-Token`
    - `X-Colab-Client-Agent`

  ## Why this matters

  Some Jupyter/client code paths may pass a preconstructed `Request` with important headers already attached. Dropping   
  them can cause subtle request failures or behavior changes that are hard to trace.

  ## Tests

  Added a regression test that verifies a wrapped `Request` keeps its original headers (`Accept` and a custom test       
  header) while still receiving the required Colab proxy headers.

  ## Files

  - `src/jupyter/assignments.ts`
  - `src/jupyter/assignments.unit.test.ts`

### DIFF
--- a/src/jupyter/assignments.ts
+++ b/src/jupyter/assignments.ts
@@ -828,14 +828,14 @@ function colabProxyFetch(
       info = new Request(info.url, info);
     }
 
-    init ??= {};
-    const headers = new Headers(init.headers);
-    headers.append(COLAB_RUNTIME_PROXY_TOKEN_HEADER.key, token);
-    headers.append(
+    const headers = new Headers(isRequest(info) ? info.headers : undefined);
+    new Headers(init?.headers).forEach((value, key) => headers.set(key, value));
+    headers.set(COLAB_RUNTIME_PROXY_TOKEN_HEADER.key, token);
+    headers.set(
       COLAB_CLIENT_AGENT_HEADER.key,
       COLAB_CLIENT_AGENT_HEADER.value,
     );
-    init.headers = headers;
+    init = { ...init, headers };
 
     return fetch(info, init);
   };
--- a/src/jupyter/assignments.ts
+++ b/src/jupyter/assignments.ts
@@ -815,7 +815,10 @@ class AllAcceleratorsUnavailableError extends Error {
  * To work around this, we create a new `Request` instance to ensure
  * compatibility.
  *
- * @param token - The cancellation token.
+ * Colab proxy headers always win over any caller-supplied values for the
+ * same keys.
+ *
+ * @param token - The Colab runtime proxy token.
  * @returns A fetch function that adds the Colab runtime proxy token as a
  * header.
  */
--- a/src/jupyter/assignments.ts
+++ b/src/jupyter/assignments.ts
@@ -823,18 +823,19 @@ function colabProxyFetch(
   token: string,
 ): (info: RequestInfo, init?: RequestInit) => Promise<Response> {
   return async (info: RequestInfo, init?: RequestInit) => {
+    let infoHeaders: Headers | undefined;
     if (isRequest(info)) {
       // Ensure compatibility with `node-fetch`
       info = new Request(info.url, info);
+      infoHeaders = info.headers;
     }
 
-    const headers = new Headers(isRequest(info) ? info.headers : undefined);
-    new Headers(init?.headers).forEach((value, key) => headers.set(key, value));
+    const headers = new Headers(infoHeaders);
+    new Headers(init?.headers).forEach((value, key) => {
+      headers.set(key, value);
+    });
     headers.set(COLAB_RUNTIME_PROXY_TOKEN_HEADER.key, token);
-    headers.set(
-      COLAB_CLIENT_AGENT_HEADER.key,
-      COLAB_CLIENT_AGENT_HEADER.value,
-    );
+    headers.set(COLAB_CLIENT_AGENT_HEADER.key, COLAB_CLIENT_AGENT_HEADER.value);
     init = { ...init, headers };
 
     return fetch(info, init);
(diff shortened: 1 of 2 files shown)


src/jupyter/assignments.ts
```diff
@@ -828,14 +828,14 @@ function colabProxyFetch(
...
-    const headers = new Headers(init.headers);
-    headers.append(COLAB_RUNTIME_PROXY_TOKEN_HEADER.key, token);
-    headers.append(
+    const headers = new Headers(isRequest(info) ? info.headers : undefined);
+    new Headers(init?.headers).forEach((value, key) => headers.set(key, value));
+    headers.set(COLAB_RUNTIME_PROXY_TOKEN_HEADER.key, token);
+    headers.set(
       COLAB_CLIENT_AGENT_HEADER.key,
       COLAB_CLIENT_AGENT_HEADER.value,
     );
```
kevineger: Linter (when I approve the workflow) will complain about this.

```suggestion
    const headers = new Headers(isRequest(info) ? info.headers : undefined);
    new Headers(init?.headers).forEach((value, key) => {
      headers.set(key, value);
    });
    headers.set(COLAB_RUNTIME_PROXY_TOKEN_HEADER.key, token);
    headers.set(COLAB_CLIENT_AGENT_HEADER.key, COLAB_CLIENT_AGENT_HEADER.value);
);
```

src/jupyter/assignments.ts
```diff
@@ -828,14 +828,14 @@ function colabProxyFetch(
...
-    init.headers = headers;
+    init = { ...init, headers };
 
     return fetch(info, init);
```
kevineger, line 844: Would it make sense to do the `isRequest` check once and make the intent of copying from the `infoHeaders` more clear? Something like:

```suggestion
    let infoHeaders: Headers | undefined;
    if (isRequest(info)) {
      // Ensure compatibility with `node-fetch`
      info = new Request(info.url, info);
      infoHeaders = info.headers;
    }

    const headers = new Headers(infoHeaders);
    new Headers(init?.headers).forEach((value, key) => {
      headers.set(key, value);
    });
    headers.set(COLAB_RUNTIME_PROXY_TOKEN_HEADER.key, token);
    headers.set(COLAB_CLIENT_AGENT_HEADER.key, COLAB_CLIENT_AGENT_HEADER.value);
    init = { ...init, headers };

    return fetch(info, init);
```

src/jupyter/assignments.unit.test.ts
```diff
@@ -534,6 +534,33 @@ describe('AssignmentManager', () => {
           });
         });
 
+        it('preserves request headers when wrapping a Request object', async () => {
```
kevineger, line 537: The new test only covers a `Request` with headers and no `init`. It doesn't test the more complex path where both `request.headers` and `init.headers` are provided, particularly whether `init.headers` correctly overrides a `Request` header. Could we add a test for that too?

src/jupyter/assignments.unit.test.ts
```diff
@@ -534,6 +534,33 @@ describe('AssignmentManager', () => {
...
+          sinon.assert.calledOnce(fetchStub);
+          const headers = fetchStub.firstCall.args[1]?.headers as Headers;
+          expect(headers.get('Accept')).to.equal('application/json');
+          expect(headers.get('X-Test')).to.equal('existing-value');
+          expect(headers.get(COLAB_RUNTIME_PROXY_TOKEN_HEADER.key)).to.equal(
+            server.connectionInformation.token,
+          );
+          expect(headers.get(COLAB_CLIENT_AGENT_HEADER.key)).to.equal(
+            COLAB_CLIENT_AGENT_HEADER.value,
+          );
```
kevineger: To avoid the `as` cast and the individual assertions, we could lean on a sinon matcher here, similar to what's done in the _includes a fetch implementation that attaches Colab connection info_ test case above:

```ts
sinon.assert.calledOnceWithMatch(fetchStub, 'https://example.com', {
  headers: new Headers({
    [COLAB_RUNTIME_PROXY_TOKEN_HEADER.key]:
      server.connectionInformation.token,
    [COLAB_CLIENT_AGENT_HEADER.key]: COLAB_CLIENT_AGENT_HEADER.value,
  }),
});
```

kevineger (changes requested):
Thanks 👍 

As with https://github.com/googlecolab/colab-vscode/pull/559, I'm super curious if you hit an issue which caused you to find and send this fix?

dipeshbabu:
> Thanks 👍
> 
> As with #559, I'm super curious if you hit an issue which caused you to find and send this fix?

Thanks, addressed the review comments.                                                                                              

I updated `colabProxyFetch()` to do the `isRequest()` check once, preserve headers from the `Request`, overlay `init.headers`, and then set the Colab proxy headers.                                                                                                   
                                                                                                                                      
I also updated the test to avoid the cast and added coverage for the case where both `request.headers` and `init.headers` are present, including verifying that `init.headers` overrides the request header.                                                      
                                                                                                                                      
For context: I did not hit a specific user-facing issue. I found this while tracing the proxy fetch wrapper and noticed that a      
prebuilt `Request` could lose its existing headers. So this is mainly a defensive correctness fix for request wrapping.

dipeshbabu:
@kevineger could you review the PR?

kevineger (approved):

kevineger:
Thanks for the contribution :+1:

PR #559, opened 2026-04-15
fix: revoke managed connections when assignment removals are mixed

dipeshbabu:
## Summary                                                                                                             
                                                                                                                         
  Fix `JupyterConnectionManager.revoke(...)` so a mixed assignment-removal batch does not stop revocation early when some
  removed endpoints do not have a managed local connection.                                                              
                                                                                                                         
  Previously, the method returned immediately on the first missing endpoint. If a later endpoint in the same batch did   
  have a managed connection, it would never be revoked.

  ## What changed

  - Replace the early `return` with `continue` when a removed endpoint is not present in the local connection map        
  - Track which endpoints were actually revoked
  - Emit `onDidRevokeConnections` only for endpoints that were actually revoked

  ## Why this matters

  Assignment change events can include a mix of:
  - endpoints managed by this connection manager
  - unrelated endpoints with no local connection

  The old behavior made revocation order-dependent and could leave stale managed connections alive.

  ## Tests

  Added a regression test covering this case:
  - an unrelated removed endpoint appears first
  - a managed endpoint appears later in the same removal batch
  - the managed connection is still revoked

  ## Files

  - `src/jupyter/contents/sessions.ts`
  - `src/jupyter/contents/sessions.unit.test.ts`

### DIFF
--- a/src/jupyter/contents/sessions.ts
+++ b/src/jupyter/contents/sessions.ts
@@ -235,16 +235,18 @@ export class JupyterConnectionManager implements Disposable {
     if (!endpoints.length) {
       return;
     }
+    const revoked: string[] = [];
     for (const endpoint of endpoints) {
       const promise = this.connections.get(endpoint);
       if (!promise) {
-        return;
+        continue;
       }
       bestEffortDisposeConnection(promise);
       this.connections.delete(endpoint);
+      revoked.push(endpoint);
     }
-    if (!silent) {
-      this.revokeConnectionEmitter.fire(endpoints);
+    if (!silent && revoked.length) {
+      this.revokeConnectionEmitter.fire(revoked);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


kevineger (approved):
Thanks for the contribution!

While I agree this is a bug, I'm very curious what prompted you to find & fix it. Did you hit some issue? Looking at the existing call-sites, I can't see how the method would be invoked for an _unrelated_ removal.

dipeshbabu:
> Thanks for the contribution!
> 
> While I agree this is a bug, I'm very curious what prompted you to find & fix it. Did you hit some issue? Looking at the existing call-sites, I can't see how the method would be invoked for an _unrelated_ removal.

Thanks for reviewing and merging, @kevineger!                                                                                                   
                                                                                                                                      
I did not hit a specific user-facing issue. I found this while tracing the assignment-change revocation path and noticed `revoke()` was order-dependent: one unknown endpoint could stop the rest of the batch.                                                         
                                                                                                                                      
I agree the current call sites do not seem likely to pass an unrelated removal, since removals come from extension-managed stored servers. So this was mainly a defensive correctness fix for the private batch helper, making it skip unknown endpoints and revoke/report only the connections it actually owns.

PR #567, opened 2026-04-20
Add telemetry for notebook imports, including source of import (command palette vs. URI)

mounikasomisetty:
Other half of the changes are in cl/902650826

### DIFF
--- a/src/drive/commands/import.ts
+++ b/src/drive/commands/import.ts
@@ -6,6 +6,8 @@
 
 import vscode from 'vscode';
 import { log } from '../../common/logging';
+import { telemetry } from '../../telemetry';
+import { CommandSource } from '../../telemetry/api';
 import { DriveClient } from '../client';
 import { IMPORT_DRIVE_FILE_PATH, IMPORT_NOTEBOOK_FROM_URL } from './constants';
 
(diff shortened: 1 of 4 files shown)


src/telemetry/index.ts
```diff
@@ -88,8 +94,17 @@ export const telemetry = {
   logHandleEphemeralAuth: (authType: AuthType) => {
     log({ handle_ephemeral_auth_event: { auth_type: authType } });
   },
-  logImportNotebook: (source: CommandSource) => {
-    log({ import_notebook_event: { source } });
+  logImportNotebook: (
+    source: CommandSource,
+    notebookSource?: NotebookSource,
```
kevineger: Nit: why `?:`? Can we require it?

kevineger (approved):
Thanks!

src/telemetry/index.ts
mounikasomisetty: done

src/drive/commands/import.ts
```diff
@@ -23,6 +25,13 @@ export async function importNotebookFromUrl(
   driveClient: DriveClient,
   inputUrl?: string,
 ): Promise<void> {
+  telemetry.logImportNotebook(
+    inputUrl
+      ? CommandSource.COMMAND_SOURCE_ON_URI
```
hjjackyang, line 30: What does this source mean? Does it mean it's deep linked from Colab web app?

hjjackyang (approved):
LGTM with just a question for my own understanding.

src/drive/commands/import.ts
mounikasomisetty: Yes it was activated 'onUri' i.e. via deep-linking.

PR #572, opened 2026-04-23
refactor: split extension.ts activation into per-feature modules

kevineger:
Reduce `src/extension.ts` from ~380 lines to ~100 by extracting service construction, command registration, and URI dispatch into per-feature module factories. Each new factory takes typeof vscode as a parameter, making activation logic that was previously locked in extension.ts (and therefore not unit-testable) reachable from tests.

Test notes:

- Each module factory has a test asserting shape (services present, disposables/toggles non-empty) and identity (specific service classes appear in toggles via instanceof checks; specific command IDs appear in registrations).
- `src/jupyter/module.vscode.test.ts` is an integration test (runs in the VS Code host) so the factory can take the real vscode module as input. This is what lets `ContentTreeProvider`/`ResourceTreeProvider` continue to extend concrete VS Code `TreeItem` types and import vscode values directly. The test stubs `vscode.workspace.registerFileSystemProvider`, `vscode.window.createTreeView`, and `vscode.commands.registerCommand` to avoid colliding with the activated extension's own registrations, while still asserting on the registration arguments.
- Test stub `src/test/helpers/vscode.ts` gains `commands.registerCommand`.

### DIFF
--- a/src/auth/module.ts
+++ b/src/auth/module.ts
@@ -1,0 +1,62 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { OAuth2Client } from 'google-auth-library';
+import vscode, { Disposable, ExtensionContext } from 'vscode';
+import { CONFIG } from '../colab-config';
+import { PackageInfo } from '../config/package-info';
+import { GoogleAuthProvider } from './auth-provider';
+import { getOAuth2Flows } from './flows/flows';
+import { login, LoginOptions } from './login';
+import { AuthStorage } from './storage';
+
+/** The result of activating the auth module. */
+export interface AuthModule {
+  /** The {@link GoogleAuthProvider} for downstream wiring. */
+  readonly authProvider: GoogleAuthProvider;
+  /** Disposables that should be pushed into `context.subscriptions`. */
+  readonly disposables: Disposable[];
+}
+
+/**
+ * Builds the {@link GoogleAuthProvider} and its supporting OAuth2 flow
+ * components. The returned `disposables` should be pushed into
+ * `context.subscriptions`.
+ *
+ * @param vs - The VS Code API instance.
+ * @param context - The extension context (for `secrets` access).
+ * @param packageInfo - Information about the installed extension.
+ * @returns The auth provider and its associated disposables.
+ */
+export function createAuthModule(
+  vs: typeof vscode,
+  context: ExtensionContext,
+  packageInfo: PackageInfo,
+): AuthModule {
+  const authClient = new OAuth2Client(
+    CONFIG.ClientId,
+    CONFIG.ClientNotSoSecret,
+  );
+  const authFlows = getOAuth2Flows(vs, packageInfo, authClient);
+  const authProvider = new GoogleAuthProvider(
+    vs,
+    new AuthStorage(context.secrets),
+    authClient,
+    (scopes: string[], options?: LoginOptions) =>
+      login(vs, authFlows, authClient, scopes, options),
+  );
+  const flowsDisposable: Disposable = {
+    dispose: () => {
+      for (const flow of authFlows) {
+        flow.dispose?.();
+      }
+    },
+  };
+  return {
+    authProvider,
+    disposables: [flowsDisposable, authProvider],
+  };
+}
--- a/src/jupyter/module.ts
+++ b/src/jupyter/module.ts
@@ -1,0 +1,140 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Jupyter } from '@vscode/jupyter-extension';
+import vscode, { Disposable, Extension, ExtensionContext } from 'vscode';
+import { GoogleAuthProvider } from '../auth/auth-provider';
+import { ColabClient } from '../colab/client';
+import { registerColabCommands } from '../colab/commands/register';
+import { ConnectionRefreshController } from '../colab/connection-refresher';
+import { ContentTreeProvider } from '../colab/content-browser/content-tree';
+import { registerContentBrowserCommands } from '../colab/content-browser/register';
+import { ServerKeepAliveController } from '../colab/keep-alive';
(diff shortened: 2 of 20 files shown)


src/jupyter/module.ts
```diff
@@ -0,0 +1,140 @@
...
+   */
+  readonly disposables: Disposable[];
+  /**
+   * Command-registration disposables. Push these LAST in
+   * `context.subscriptions` so they are disposed earliest, removing command
+   * handlers before the underlying services tear down.
+   */
+  readonly commandDisposables: Disposable[];
+  /** Auth-gated toggles that should be passed to `whileAuthorized`. */
+  readonly toggles: Toggleable[];
```
hjjackyang, line 44: Does it make sense to make a common `Module` interface which exposes these 3 things (`disposables`, `commandDisposables`, `toggles`)? The individual module interfaces can extend the common one. For those that may not have all the elements, they can simply return `[]`.

This ensures all module implementation are standardized (easier for future ones to follow).

Also, this allows us to keep a generic module registry (i.e. a list of modules) in `extension.ts`. All modules' `toggles` will be passed to `whileAuthorized`, and all modules' `disposables` will be pushed to `context.subscriptions` first and `commandDisposables` last. This ensures if we implement future modules, we only need to "register" them (i.e. add to the list) once, instead of needing to update multiple places to register toggles and disposables separately.

WDYT?

hjjackyang (commented):
I have a meta question before doing a more detailed review.

src/auth/module.ts
```diff
@@ -0,0 +1,62 @@
...
+import vscode, { Disposable, ExtensionContext } from 'vscode';
+import { CONFIG } from '../colab-config';
+import { PackageInfo } from '../config/package-info';
+import { GoogleAuthProvider } from './auth-provider';
+import { getOAuth2Flows } from './flows/flows';
+import { login, LoginOptions } from './login';
+import { AuthStorage } from './storage';
+
+/** The result of activating the auth module. */
+export interface AuthModule {
```
hjjackyang, line 17: I personally prefer object-oriented programming, so instead of an `interface` + a `function` to construct each module, have you considered defining a factory `class` with a static `create` method to construct a module?

Something like:
```ts
export class AuthModule implements Module {
  // readonly properties...

  static create(...): AuthModule {
    ...
  }

  private constructor(...) { ... }
}
```
where `Module` is the common interface I brought up in the other comment.

hjjackyang (commented):
Another meta question (sorry for the fragmented review comments).

src/jupyter/module.ts
kevineger: The modules don't form a uniform list, they have explicit dependency order encoded in the construction chain in `extension.ts`:

- `auth` produces `authProvider`
- `colabClient` needs `authProvider` + `packageInfo`
- `drive` needs `authProvider`
- `jupyter` needs `authProvider` + `colabClient`
- `colab` needs `colabClient` + `jupyter.assignmentManager`

A `const modules = [auth, drive, jupyter, colab]` registry would still need that explicit chain because of the cross-module data flow (`assignmentManager`, `authProvider`, `colabClient` get threaded between modules). Iterating a list to call `register()` only works for plugins that are independent; ours aren't. I think there's value in making the dependency graph visible and is the main thing I'm trying to address with this refactor.

I'm also wary of forcing a uniform `Module` shape. `AuthModule` exposes `authProvider`, `JupyterModule` exposes `assignmentManager`, etc.. These service surfaces are what the next module in the chain consumes. If every module also implements a uniform `{disposables, commandDisposables, toggles}` interface with empties for "not applicable", readers have to ask of every empty array "genuinely none, or doesn't apply?" The current per-module interfaces communicate that in the type: you can't read `auth.toggles` because it doesn't exist.

W.r.t.  _future modules_, we have four modules total in a 112-line activation file. Adding a fifth costs three lines today. I'd rather pay that cost when it's real than build a registry abstraction now.

One thing we could consider doing which maybe at least in part addresses some of the points you raise would be to add a tiny helper in `extension.ts` like `pushServiceDisposables(context, ...modules)` and `pushCommandDisposables(context, ...modules)` that takes anything matching `{ disposables?: Disposable[] }` / `{ commandDisposables?: Disposable[] }` structurally. That encodes the ordering invariant without imposing a registry or a forced interface. WDYT? Worth it, or are the comments + visible ordering enough?

src/auth/module.ts
kevineger: I'd lean against this. The modules are dataclasses, they don't have instance methods, no state mutation, no polymorphism. Converting them to classes with a static `create` and private constructor gets us a different name for `createAuthModule` and nominal typing, but there's no behavior on the instance to justify the class.

Two concrete costs:

1. **Test ergonomics.** Classes in this codebase are stubbed with `sinon.createStubInstance`. `createStubInstance` only stubs methods. On a methodless class it gives you nothing, and tests that today construct a fake `AuthModule` as `{ authProvider, disposables: [] }` would either need to call the real `create` (defeating isolation) or use `createStubInstance` for zero benefit.
2. **Convention signal.** Today in this codebase, `class` means "thing with behavior" (`GoogleAuthProvider`, `ColabClient`, `AssignmentManager`, etc.). Using `class` for a struct-shaped value muddies that signal.

Also worth noting: `colab/module.ts` exports two factories (`createColabClient` and `createColabModule`) because the client is constructed earlier in the dependency chain than the module's background services. The class-with-`create` pattern would either force these into two artificial classes or jam them awkwardly together.

Happy to revisit if a module grows real instance methods, at that point a class earns its keep.

hjjackyang: SG! Fair point with no instance methods or state mutations.

src/jupyter/module.ts
hjjackyang: I didn't think too crazy about my suggestion, and I agree that what you have now is already very clean. 1- vs 3-line cost doesn't justify the uniform abstraction I suggested if there is a risk of masking the underlying dependency chain.

Let's keep what you have, and I don't think it's worth extracting `pushServiceDisposables` and `pushCommandDisposables` either given your comments are pretty clear.

hjjackyang (approved):

PR #574, opened 2026-04-23
feat: log telemetry for opening a Colab terminal

kevineger:
Captures the source (command palette, notebook toolbar, or tree view inline
button) for every `Open Terminal` invocation. Adds a new
`COMMAND_SOURCE_TREE_VIEW_INLINE` enum value to attribute commands invoked from
inline buttons in the activity bar tree views.

The `OPEN_TERMINAL` command handler can now be invoked with either a
`CommandSource` (when invoked programmatically from the notebook toolbar) or a
`ContentItem` (when invoked from the tree view inline button); the handler
discriminates between the two and falls back to
`COMMAND_SOURCE_COMMAND_PALETTE` when neither is provided.

Proto changes: cl/903967220

### DIFF
--- a/src/colab/commands/terminal.unit.test.ts
+++ b/src/colab/commands/terminal.unit.test.ts
@@ -5,11 +5,13 @@
  */
 
 import { randomUUID } from 'crypto';
-import sinon, { SinonStubbedInstance } from 'sinon';
+import sinon, { SinonStubbedFunction, SinonStubbedInstance } from 'sinon';
 import { ExtensionTerminalOptions, QuickPick, QuickPickItem } from 'vscode';
 import { Variant } from '../../colab/api';
 import { AssignmentManager } from '../../jupyter/assignments';
 import { ColabAssignedServer } from '../../jupyter/servers';
+import { telemetry } from '../../telemetry';
+import { CommandSource } from '../../telemetry/api';
 import {
   buildQuickPickStub,
   QuickPickStub,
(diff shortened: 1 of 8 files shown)


src/colab/commands/terminal.unit.test.ts
```diff
@@ -174,6 +194,65 @@ describe('openTerminal command', () => {
...
   });
 
+  describe('telemetry', () => {
+    let logStub: SinonStubbedFunction<typeof telemetry.logOpenTerminal>;
+
+    beforeEach(() => {
+      logStub = sinon.stub(telemetry, 'logOpenTerminal');
+    });
+
+    it('logs with the default source when none is provided', async () => {
```
hjjackyang: This test case doesn't seem too different from the `logs with the provided source` test case below. Though this one is named `logs with the default source when none is provided`, it is actually providing a source too.

src/colab/commands/terminal.unit.test.ts
```diff
@@ -174,6 +194,65 @@ describe('openTerminal command', () => {
...
+        CommandSource.COMMAND_SOURCE_TREE_VIEW_INLINE,
+      );
+
+      sinon.assert.calledOnceWithExactly(
+        logStub,
+        CommandSource.COMMAND_SOURCE_TREE_VIEW_INLINE,
+      );
+    });
+
+    it('logs even when a terminal is opened with the only server', async () => {
```
hjjackyang: Is this test case testing anything different than `logs with the provided source` from the telemetry perspective? Given `logOpenTerminal` is the first line in `openTerminal`.

src/extension.ts
```diff
@@ -331,9 +331,24 @@ function registerCommands(
...
+      ) => {
+        // The command can be invoked from the command palette (no args), the
+        // notebook toolbar (passing a CommandSource), or the tree view inline
+        // button (passing a ContentItem as the first argument).
+        const source =
+          typeof sourceOrContextItem === 'number'
+            ? sourceOrContextItem
+            : sourceOrContextItem === undefined
+              ? CommandSource.COMMAND_SOURCE_COMMAND_PALETTE
+              : CommandSource.COMMAND_SOURCE_TREE_VIEW_INLINE;
```
hjjackyang: Nit: We should add a test case for this logic after you merge your module splitting PR :)

hjjackyang (commented):
I will hold off on approving this PR after the module splitting PR merge because I think it's valuable to add a test case for the `CommandSource` logic in one of my review comments.

hjjackyang (approved):

PR #578, opened 2026-04-23
feat: log telemetry for low/depleted CCU balance notifications

kevineger:
Captures the severity of the notification (`SEVERITY_LOW` warning or
`SEVERITY_DEPLETED` error), the user's subscription tier, and whether the user
clicked the call-to-action (vs. dismissed the notification).

The event is logged after the notification's promise resolves so the
`clicked_action` outcome is captured. Pairs with `upgrade_to_pro_event` to
enable funnel analysis from notification to upgrade.

`telemetry.logLowCcuNotification` accepts the Colab-level `SubscriptionTier`
directly and converts it to the proto-shaped top-level `SubscriptionTier` enum
internally, keeping callers free of the conversion.

Proto changes: cl/903985928

### DIFF
--- a/src/colab/consumption/notifier.unit.test.ts
+++ b/src/colab/consumption/notifier.unit.test.ts
@@ -5,7 +5,9 @@
  */
 
 import { assert, expect } from 'chai';
-import sinon, { SinonFakeTimers } from 'sinon';
+import sinon, { SinonFakeTimers, SinonStubbedFunction } from 'sinon';
+import { telemetry } from '../../telemetry';
+import { LowBalanceSeverity } from '../../telemetry/api';
 import { TestEventEmitter } from '../../test/helpers/events';
 import { newVsCodeStub, VsCodeStub } from '../../test/helpers/vscode';
 import { SubscriptionTier, ConsumptionUserInfo } from '../api';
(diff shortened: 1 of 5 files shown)


src/colab/consumption/notifier.unit.test.ts
```diff
@@ -426,6 +428,111 @@ describe('ConsumptionNotifier', () => {
...
+      await noOp;
+
+      sinon.assert.calledOnceWithExactly(
+        logStub,
+        LowBalanceSeverity.SEVERITY_DEPLETED,
+        SubscriptionTier.PRO_PLUS,
+        true,
+      );
+    });
+  });
```
hjjackyang: These 3 seem to be similar tests for different tiers as the only variant. Consider parameterizing with a [dynamic test](https://mochajs.org/declaring/dynamic-tests/)?

(Also, the existing test names `plumbs the <X> subscription tier through unchanged` are quite confusing to me.)

src/telemetry/telemetry.unit.test.ts
```diff
@@ -275,6 +280,52 @@ describe('Telemetry Module', () => {
...
+        sinon.assert.calledOnceWithExactly(logStub, {
+          ...baseLog,
+          low_ccu_notification_event: {
+            severity: LowBalanceSeverity.SEVERITY_LOW,
+            subscription_tier: expected,
+            clicked_action: false,
+          },
+        });
+      }
+    });
```
hjjackyang: I think this test is also better done as a [dynamic test](https://mochajs.org/declaring/dynamic-tests/) than coding a loop logic inside the test.

hjjackyang (changes requested):
Similar test-related comments as other PRs, but I have a stronger opinion to change these ones given the confusing test names and loop logic inside the test, hence requested changes.

src/telemetry/telemetry.unit.test.ts
hjjackyang: Any reason this comment is resolved but not addressed?

hjjackyang (approved):

src/telemetry/telemetry.unit.test.ts
hjjackyang: Thanks!

kevineger: Sorry had conflated it with feedback in [src/colab/consumption/notifier.unit.test.ts](https://github.com/googlecolab/colab-vscode/pull/578/files#diff-0c08d0f2b9fc3cef1cc95c5b898f828701a174d8e27ba17720f0105afc377f6e)

PR #580, opened 2026-04-24
chore: upgrade actions node24

kevineger:
Changes:

- actions/checkout@v4 → @v6
- actions/setup-node@v4 → @v6
- actions/upload-artifact@v4 → @v7
- actions/download-artifact@v4 → @v8

All of these majors run on Node.js 24, which addresses the deprecation notice.

[CI verification](https://github.com/googlecolab/colab-vscode/actions/runs/24905482360)

### DIFF
--- a/.github/actions/run-e2e/action.yml
+++ b/.github/actions/run-e2e/action.yml
@@ -5,7 +5,7 @@ runs:
   using: "composite"
   steps:
     - name: Use Node.js
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@v6
       with:
         # Assuming node version is provided via env var in the workflow
         node-version: ${{ env.NODE_VERSION }}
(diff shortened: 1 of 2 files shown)


hjjackyang (approved):
Should we also update https://github.com/googlecolab/colab-vscode/blob/main/.nvmrc to bring our local Node version to match?

kevineger:
> Should we also update https://github.com/googlecolab/colab-vscode/blob/main/.nvmrc to bring our local Node version to match?

Good catch, but let's keep this PR scoped to the deprecation fix. `.nvmrc` (and `NODE_VERSION`, and `@types/node`) describe our build/test toolchain, while the deprecated actions were about the runtime that GitHub's own JS actions run on. Bumping our dev toolchain to Node 24 is reasonable since 24 is now LTS, but it should be its own PR so we move `.nvmrc`, `NODE_VERSION`, and `@types/node` together.

I can follow up separately with this. I need to look into what's supported with our VS Code engine version target, the associated electron version, etc.

PR #581, opened 2026-04-24
test: add e2e test for Upload to Colab from explorer context menu

kevineger:
Adds an end-to-end test that exercises the full Upload to Colab flow from the VS Code Explorer context menu, verifying the file lands on the connected Colab runtime via two independent signals: it appears in the Colab Contents tree view, and a Python cell that reads the uploaded file and asserts its content executes successfully (the cell raises AssertionError on mismatch, failing the test).

Changes:

- Add `src/test/e2e/fixtures/workspace/hello-world.txt` as a shared workspace fixture, opened in VS Code by the e2e runner so all e2e tests get a populated Explorer.
- Wire `scripts/test_e2e.sh` to launch extest with the fixture folder via the `-r/--open_resource` flag, terminated by `--` so the variadic flag does not consume the test-file positional args.
- Patch `CodeUtil.open` in `test-setup.ts` to inject `--no-sandbox` into the VS Code CLI invocation it spawns. Without this, the spawned Electron process exits silently on CI runners (e.g. Ubuntu in GitHub Actions) where chrome-sandbox lacks setuid root, and the `-r/--open_resource` flag becomes a no-op. See https://github.com/redhat-developer/vscode-extension-tester/issues/2050.
- Add `src/test/e2e/upload.e2e.test.ts` that connects to a Colab server via Auto Connect, right-clicks the fixture in the Explorer to trigger the upload, waits for the file in the Colab Contents view, and runs a Python cell that asserts the uploaded file's contents match the fixture.

### DIFF
--- a/src/test/e2e/upload.e2e.test.ts
+++ b/src/test/e2e/upload.e2e.test.ts
@@ -9,8 +9,6 @@ import {
   ContextMenu,
   CustomTreeSection,
   DefaultTreeSection,
-  InputBox,
-  Key,
   SideBarView,
   Workbench,
 } from 'vscode-extension-tester';
@@ -18,8 +16,9 @@ import {
   assertAllCellsExecutedSuccessfully,
   createNotebook,
   hasQuickPickItem,
-  pushDialogButtonIfShown,
+  KERNEL_SELECT_WAIT_MS,
   selectQuickPickItem,
+  selectQuickPickItemIfShown,
   selectQuickPicksInOrder,
 } from './ui';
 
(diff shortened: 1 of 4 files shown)


src/test/e2e/upload.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+it('uploads a file from the explorer context menu', async () => {
+  const workbench = new Workbench();
+  const driver = workbench.getDriver();
+
+  // Close any leftover notebooks from prior tests; otherwise their cells
+  // would be counted by `assertAllCellsExecutedSuccessfully` below and the
+  // expected vs. actual cell-count comparison would never match.
+  await workbench.executeCommand('View: Close All Editors');
+  await pushDialogButtonIfShown(driver, "Don't Save", /* timeoutMs= */ 3000);
```
hjjackyang: I guess you can remove these lines now that you've added them to a global `afterEach` in #583 ?

src/test/e2e/upload.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  await createNotebook(workbench);
+  await workbench.executeCommand('Notebook: Select Notebook Kernel');
+  if (await hasQuickPickItem(driver, 'Select Another Kernel')) {
+    await selectQuickPickItem(driver, 'Select Another Kernel');
+  }
+  await selectQuickPicksInOrder(driver, ['Colab', 'Auto Connect']);
+  // Alias the server with the default name.
+  const inputBox = await InputBox.create();
+  await inputBox.sendKeys(Key.ENTER);
```
hjjackyang: Redundant for "Auto Connect"?

src/test/e2e/upload.e2e.test.ts
```diff
@@ -0,0 +1,127 @@
...
+          .getSection('Contents', CustomTreeSection);
+        const serverItem = await contentsSection.findItem('Colab CPU');
+        if (!serverItem) {
+          return false;
+        }
+        if (!(await serverItem.isExpanded())) {
+          await serverItem.expand();
+        }
+        const child = await serverItem.findChildItem('hello-world.txt');
+        return Boolean(child);
```
hjjackyang: Nit: `!!child` for brevity?

```suggestion
        return !!child;
```

hjjackyang (approved):
LGTM with a few minor comments!

PR #583, opened 2026-04-24
test: reduce e2e flake

kevineger:
- Bump `CELL_EXECUTION_WAIT_MS` and `CONNECT_DRIVE_DIALOG_WAIT_MS` to absorb cold-start on a freshly assigned Colab server. Most of the time its quick, sometimes it's pretty slow and hanging things up.
- connect-execute: replace racy `InputBox.create() + sendKeys(Enter)` on the alias step with `confirmInputBoxWithDefault()` that polls for the expected title, confirms, and waits for the `InputBox` to transition away.
- mount-drive, resource-view: drop the alias step entirely. These tests use Auto Connect, which reuses the existing server and never prompts for an alias; the previous Enter was operating on whatever input was open and only worked by accident.
- test-setup: tolerate transient 'Colab: Remove Server' backend errors during cleanup. Catch the picker timeout and dismiss any error modal instead of failing the whole suite.
- Add a global `afterEach` that closes all editors and dismisses any 'Don't Save' prompt so a failing test's open notebook can't pollute the next test's cell-count assertion.
- Make `hasQuickPickItem` actually non-throwing: catch the outer `driver.wait` timeout so callers using `if (await hasQuickPickItem(...))` don't blow up the suite when no QuickPick renders.

Verified at 20/20 across two consecutive 20-run dispatches.

### DIFF
--- a/src/test/e2e/test-setup.ts
+++ b/src/test/e2e/test-setup.ts
@@ -28,6 +28,8 @@ assert.equal(
   'Unexpected extension environment. Run `npm run generate:config` with COLAB_EXTENSION_ENVIRONMENT="production".',
 );
 
+const DIALOG_WAIT_MS = 3000;
+
 before(async function () {
   console.log('Starting global E2E test setup...');
   const workbench = new Workbench();
@@ -65,11 +67,7 @@ afterEach(async function () {
   try {
     await workbench.executeCommand('View: Close All Editors');
     // Close-all may surface a "Don't Save" prompt if any notebook is dirty.
-    await pushDialogButtonIfShown(
-      vsCodeDriver,
-      "Don't Save",
-      /* timeoutMs= */ 3000,
-    );
+    await pushDialogButtonIfShown(vsCodeDriver, "Don't Save", DIALOG_WAIT_MS);
   } catch (err) {
     // Best-effort cleanup; never fail the test from afterEach.
     console.warn('Best-effort editor cleanup failed in afterEach:', err);
@@ -86,11 +84,7 @@ async function signIn(
   // Dismiss the telemetry notice modal if it appears. The extension activates
   // asynchronously after notebook creation, so we poll for the dialog. This is
   // a no-op when the notice was already acknowledged (e.g. developer machine).
-  await pushDialogButtonIfShown(
-    vsCodeDriver,
-    'Acknowledge',
-    /* timeoutMs= */ 3000,
-  );
+  await pushDialogButtonIfShown(vsCodeDriver, 'Acknowledge', DIALOG_WAIT_MS);
 
   // Trigger Colab connection which will prompt for sign-in.
   await workbench.executeCommand('Notebook: Select Notebook Kernel');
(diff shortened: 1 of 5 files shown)


src/test/e2e/ui.ts
```diff
@@ -72,31 +83,40 @@ export async function hasQuickPickItem(
...
+          // looking for.
+          if (items.length !== 0) {
+            return await Promise.all(
+              items.map(async (i) => await i.getLabel()),
+            );
+          }
+          // No QuickPick items were shown, which likely means the QuickPick is
+          // still loading. Keep waiting.
+          return false;
+        } catch (_) {
```
hjjackyang: Nit: This may be simplified as:

```suggestion
        } catch {
```

src/test/e2e/ui.ts
```diff
@@ -72,31 +83,40 @@ export async function hasQuickPickItem(
...
-    `Could not find "${item}" in QuickPick`,
-  );
+      },
+      ELEMENT_WAIT_MS,
+      `Could not find "${item}" in QuickPick`,
+    );
+  } catch (_) {
+    // No QuickPick (or no items) appeared within the wait window. Treat as
+    // "item not present" rather than failing the caller.
+    return false;
```
hjjackyang: Nit: Shall we consider `console.warn` the error message here, so we don't fully swallow the `Could not find "${item}" in QuickPick` message from `driver.wait`?

Otherwise, consider omitting the optionally `Could not find "${item}" in QuickPick` message in `driver.wait` if it won't be surfaced anywhere.

src/test/e2e/ui.ts
```diff
@@ -124,6 +144,70 @@ export async function selectQuickPicksInOrder(
...
+    async () => {
+      try {
+        const inputBox = await InputBox.create();
+        const title = await inputBox.getTitle();
+        if (!title?.includes(expectedTitleSubstring)) {
+          return false;
+        }
+        await inputBox.confirm();
+        return true;
+      } catch (_) {
```
hjjackyang:
```suggestion
      } catch {
```

src/test/e2e/ui.ts
```diff
@@ -124,6 +144,70 @@ export async function selectQuickPicksInOrder(
...
+  // Phase 2: wait for the InputBox to transition away. It either closes
+  // entirely or is replaced by an unrelated InputBox/QuickPick whose title
+  // does not contain the expected substring.
+  await driver.wait(
+    async () => {
+      try {
+        const inputBox = await InputBox.create();
+        const title = await inputBox.getTitle();
+        return !title?.includes(expectedTitleSubstring);
+      } catch (_) {
```
hjjackyang:
```suggestion
      } catch {
```

src/test/e2e/test-setup.ts
```diff
@@ -55,6 +55,27 @@ before(async function () {
...
+  // from the prior notebook).
+  const workbench = new Workbench();
+  const vsCodeDriver = workbench.getDriver();
+  try {
+    await workbench.executeCommand('View: Close All Editors');
+    // Close-all may surface a "Don't Save" prompt if any notebook is dirty.
+    await pushDialogButtonIfShown(
+      vsCodeDriver,
+      "Don't Save",
+      /* timeoutMs= */ 3000,
```
hjjackyang: Nit: This timeout value (`3000`) is used twice (here and line 128). Consider extracting to a const at the top for reusability and visibility?

hjjackyang (approved):
Thanks for cleaning up and deflaking the e2e tests!

LGTM with some minor comments.

src/test/e2e/ui.ts
kevineger: Removed message.

src/test/e2e/test-setup.ts
kevineger: `DIALOG_WAIT_MS`

PR #584, opened 2026-04-25
feat: implement polling-based file system watching

lndmanh:
The `watch()` method in `ContentsFileSystemProvider` was previously a no-op with a TODO to add polling. This PR implements and hardens polling-backed watch support for mounted Colab filesystems so VS Code can detect remote file changes automatically instead of relying only on manual refreshes.

### Why this matters

During DL training on a Colab runtime through VS Code, I have to keep manually refreshing the file browser just to download the latest `best_checkpoint` and other files needed for quick local analysis, which is actually really frustrating for me overtime.

### Changes

**`ContentsFileSystemProvider` (`src/jupyter/contents/file-system.ts`)**
- Replaces the no-op `watch()` with a ref-counted, polling-based implementation backed by `SequentialTaskRunner`
- Snapshots watched files and directories on a configurable interval and emits `FileChangeEvent`s on detected differences (created / changed / deleted)
- Supports file watches, non-recursive directory watches, and recursive directory watches with configurable concurrency, depth limit, and max snapshot entries
- Coalesces duplicate events from overlapping watches while preserving create/delete edges for the same URI
- Avoids creating new Jupyter clients during background watch polling by reusing existing connections only
- Handles watch disposal, authority removal, transient poll failures, disappearing directories, and incomplete snapshots conservatively
- Cleans up watches automatically when a server authority is removed or the provider is disposed

**`package.json`**
- Declares three new VS Code settings so users can tune watch behaviour:
  - `colab.fileWatchPollIntervalMs` (default: 5000) — how frequently, in milliseconds, the extension polls watched files and folders for changes
  - `colab.fileWatchSnapshotRequestConcurrency` (default: 5) — how many directory listing requests can be in flight at once when building a recursive snapshot, useful for throttling on slow connections
  - `colab.fileWatchMaxSnapshotEntries` (default: 5000) — caps the number of entries collected in a single recursive snapshot to avoid runaway memory usage in very large directory trees

### Tests
- Expanded `ContentsFileSystemProvider` unit coverage for watch registration/ref-counting, file and directory diffs, recursive snapshot traversal, snapshot caps, event coalescing, authority cleanup, disposal during in-flight polls, transient polling failures, and configuration handling

This is my first commit for this change, so I’d really appreciate any feedback, edits, or review suggestions.

https://github.com/user-attachments/assets/4444af36-3caa-4615-8acb-d0d4c5d8d1e8


lndmanh:
This is a fairly large feature change, given its scale and lifecycle-sensitive nature. Although I have already spent a lot of time testing before opening this PR, there might still be cases that I have not covered yet. As such, I would appreciate careful testing and review before any decisions are made.

kevineger:
Hi @lndmanh, thanks for taking a swing at this! It's a real pain point and I appreciate the effort.

I want to be upfront about two things before we go further:

1. Size. A ~3k delta is too large for me to review effectively. I'd point you to Google's Small CLs (equivalent to PRs) [guide](https://google.github.io/eng-practices/review/developer/small-cls.html). To land something like this, we'd need to break it into a few independently reviewable PRs (e.g. watch plumbing, the polling engine, the recursive snapshot logic, settings/UX).
2. Design first. Polling is one of several viable approaches, and it has real tradeoffs worth discussing before we commit to an implementation:
    - **Cost against the Colab backend**. A recursive watch on a sizable workspace fans out into many directory listings every poll interval, per workspace, per user. That adds up, both for the user's runtime and for backend load. Worth thinking about scoping (do we only watch what's actually expanded in the tree? what VS Code is actively asking about?), backoff when the editor isn't focused, etc.
    - **Alternatives**. The [Jupyter contents API](https://jupyter-server.readthedocs.io/en/latest/developers/contents.html) supports content hashes in newer versions, which can help with file-level change detection. There's also the question of whether a push-style channel from the backend is feasible long-term. Polling may still win on simplicity, but I'd want that to be a decision rather than a default.
    - **Scope**. Recursive watches with a 5000-entry cap, configurable concurrency, three new user-facing settings: some of this may be premature. A first cut might just cover what `ContentTreeView` actually needs and grow from there.

---

**Suggested path forward**: would you mind opening a [feature request issue](https://github.com/googlecolab/colab-vscode/issues/new/choose) describing the user problem (the manual-refresh-during-training workflow you mentioned is great context)? We can hash out the design there, agree on a minimal first slice, and then we can carve this PR into smaller pieces against that plan. And if you don't have the bandwidth to work on a series of PRs to get this in, we can look into tackling it once we have an agreed upon implementation plan.

I'll hold off on a line-by-line review here in the meantime. Thanks again for the contribution, looking forward to working through it with you.

lndmanh:
Thank you for the review and for pointing this out clearly.

I understand the concern that the current change is too large to review comfortably, and I agree that it would be better to split the work into smaller, more focused changes.

Just to clarify the size of the diff: a large part of the changes comes from the added test coverage for the broader prototype. The actual source-code changes are only around 1/3 of the overall diff. That said, I understand that the total review burden is still high, especially because this change also involves design decisions around polling, watch scope, and backend load.

Thanks again for the kind words. I’ll restructure the work so it is easier to review and better aligned with the project direction. I have enough time to help with this feature, and I’m happy to contribute to moving it forward.

PR #586, opened 2026-04-27
chore: reduce e2e flake

kevineger:
Dispatched batches of 50 `workflow_dispatch` runs, triaged failures from logs and screenshots, applied targeted test-side fixes, and re-dispatched until two consecutive 50-run batches fully passed. Across ~13 batches (~650 runs) the failure rate dropped from ~8% on origin/main to 0% on the final two batches.

- mocha per-test timeout 120s -> 240s. Individual waits like `CELL_EXECUTION_WAIT_MS` (now 120s) and `CONNECT_DRIVE_DIALOG_WAIT_MS` (now 120s) were composing past the old budget and converting recoverable cold-starts into hard mocha timeouts.
- Cell-execution and Drive-dialog waits bumped to 120s; resource view wait bumped from 10s to 30s. On a freshly assigned Colab server the kernel cold start (server boot + kernel start + websocket connect) and the Resources view first-refresh both routinely exceeded the prior budgets.
- `KERNEL_SELECT_WAIT_MS` = 30s, plumbed through `selectQuickPickItem` and `hasQuickPickItem`. The Jupyter 'Detecting Kernels' phase can outlast the 10s default before the kernel-source picker is populated with 'Colab' or the kernel-spec picker is populated with 'Python'.
- New `selectQuickPickItemIfShown()` helper. The kernel-spec picker shown after kernel-source selection often auto-resolves and closes before the test can click when there's only one Python kernel available. Used at the 'Python' selection sites in all three tests and in setup.
- mount-drive's `authorizeDrive()` retries once on transient OAuth failures. Chrome occasionally returns ERR_FAILED loading accounts.google.com from the CI runner; a single retry after a 2s sleep clears the network blip. The failure-screenshot capture now mkdirSync's the screenshots dir first so an ENOENT can't mask the real OAuth error.
- `afterEach` now starts by dismissing any leftover OK modal before attempting 'View: Close All Editors'. A backend 504 surfaced by setup's best-effort 'Colab: Remove Server' could arrive after the inline dismissal window; left undismissed it blocks every subsequent `executeCommand()` in the suite.

### DIFF
--- a/src/test/e2e/connect-execute.e2e.test.ts
+++ b/src/test/e2e/connect-execute.e2e.test.ts
@@ -10,6 +10,7 @@ import {
   confirmInputBoxWithDefault,
   createNotebook,
   hasQuickPickItem,
+  KERNEL_SELECT_WAIT_MS,
   selectQuickPickItem,
   selectQuickPicksInOrder,
 } from './ui';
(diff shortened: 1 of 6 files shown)


hjjackyang (approved):
Thanks for the thorough deflaking and super detailed PR description!

PR #587, opened 2026-04-27
fix: tolerate orphan assignment deletion races in getServers

kevineger:
When listing servers from external/all, the unowned-assignment branch called `listSessions` for each _unowned_ assignment in parallel. If the assignment was deleted elsewhere (e.g. via Colab web or another VS Code instance sharing the account) between listing assignments and listing its sessions, the call would 404 and reject the entire `Promise.all`, surfacing as an error modal in downstream commands like "Colab: Remove Server".

Now, a 404 drops the orphan from the result (it's already gone server-side), and any other failure fails open with the placeholder label so the assignment is still surfaced to the user.


hjjackyang (approved):
Nice catch!

Issue #589, opened 2026-04-28
Add auto-refresh support for the remote file tree


lndmanh:
## Is your feature request related to a problem?

Yes. When using VS Code with a connected Colab runtime, files created or updated by long-running jobs do not reliably appear in the file explorer unless I manually refresh the tree. This is especially noticeable in training workflows where Colab writes artifacts over time, such as `checkpoints`, `logs`, or other generated outputs. While monitoring a run, I often need to see whether a new artifact has appeared, inspect it, or download it for quick local analysis. Today, that usually requires repeated manual refreshes in the file explorer.

As a result, the workflow feels more manual than expected and it is easy to miss newly written files during longer runs.

## Describe the solution you'd like to see

I would like the mounted Colab file tree in VS Code to eventually reflect remote file changes without requiring manual refresh. For example, if a training process creates, updates, or deletes files in a directory that is relevant to the file explorer, the visible tree should update automatically after a short delay. 

For an initial implementation, I think a narrow solution would be sufficient. The main goal is to improve the active file-explorer workflow while keeping backend and runtime cost bounded. Scoped metadata-assisted polling is a good solution because it can use the existing Jupyter contents API, avoid backend or runtime protocol changes, and limit refresh work to paths that VS Code is already showing or actively using. Instead of recursively scanning the whole mounted workspace, the extension could periodically compare lightweight directory metadata for relevant paths and refresh only the affected tree entries.

### Hash-assisted file detection through the Jupyter contents API

This can help answer whether a known file has changed, but it does not fully solve file-tree refresh by itself.

Benefits:

- useful for detecting changes to files that are already known
- can be more reliable than timestamp-only comparison for file contents
- aligns with newer Jupyter contents API capabilities where hash fields are available
- could be a useful follow-up optimization for known-file checks

Concerns:

- hash support depends on the Jupyter server version and contents manager implementation
- hashes mainly help with known files, not directory discovery
- hashes do not detect newly created files unless the containing directory is still checked
- hashes do not detect deleted files unless the containing directory is still checked
- hashes do not replace polling or another mechanism for observing directory state
- requesting or computing hashes may have its own backend cost
- fallback behavior is still required when hashes are unavailable

### Push-style backend events

A push-based event channel would likely be the best long-term design if the Colab or Jupyter backend can expose reliable file-change notifications. In this model, the runtime or backend would detect file changes and send events to the VS Code extension. The extension could then refresh only the affected file-tree entries instead of polling directories repeatedly.

[...]

PR #598, opened 2026-05-12
feat: add contents directory poller

lndmanh:
## Summary

Adds `DirectoryPoller`, the low-level polling unit for future Colab filesystem watch support. It provides support for working on #589 .

### Why this matters

Colab's Jupyter Contents API does not provide native filesystem watch events. To support VS Code file watching safely, we need a small, testable polling primitive before adding orchestration or provider integration.

### Changes

**`DirectoryPoller` (`src/jupyter/contents/directory-poller.ts`)**

- Adds a ref-counted poller for one watched directory URI
- Polls direct directory contents with `client.get({ path, type: 'directory' }, { signal })`
- Uses `SequentialTaskRunner` with `OverrunPolicy.AbandonAndRun`
- Emits `Created`, `Changed`, and `Deleted` events by diffing snapshots
- Treats type flips as `Deleted` followed by `Created`
- Detects changes from `mtime` or `size` differences
- Uses exponential backoff on transient failures, starting at 5 seconds and capped at 5 minutes
- Resets backoff after a successful poll
- Emits `Deleted` for the watched root and disposes on terminal `404`
- Supports `suspend()` / `resume()`, with immediate refresh on resume
- Treats missing existing clients as no-op ticks

### Tests

- Added unit coverage for initial snapshot behavior, created/changed/deleted diffs, type flips, size and mtime changes, exponential backoff, terminal 404 handling, suspend/resume, refcounting, and zero-ref state cleanup


src/jupyter/contents/directory-poller.ts
```diff
@@ -0,0 +1,368 @@
...
+   * Handles a failed poll, including terminal deletion and backoff.
+   *
+   * @param error - Error thrown while polling.
+   */
+  private handlePollError(error: unknown): void {
+    if (error instanceof ResponseError && error.response.status === 404) {
+      this.options.onDidChangeFile([
+        { type: this.options.vs.FileChangeType.Deleted, uri: this.options.uri },
+      ]);
+      this.dispose();
```
kevineger: Self-disposing here makes the ref-counted API awkward for callers. After this runs, any subsequent `release()`, `addRef()`, `start()`, `suspend()`, or `resume()` call from the owner throws via `guardDisposed()`. The orchestration layer that holds the `addRef()` will, in the natural cleanup path, want to call `release()` on the watch handle and instead has to special-case `isDisposed`.

Two options worth considering:
1. Don't auto-dispose. Set an internal `isTerminated` flag, stop the runner, and let `release()` continue to be safe (no-op past 0).
2. (Preferred) Make `release()`/`addRef()` tolerant of the disposed state (return current `refCount` instead of throwing), and clearly document the lifecycle.

src/jupyter/contents/directory-poller.ts
```diff
@@ -0,0 +1,368 @@
...
+
+  /**
+   * Runs one scheduled poll when not suspended or backed off.
+   *
+   * @param signal - Abort signal for the current scheduled poll.
+   */
+  private async poll(signal: AbortSignal): Promise<void> {
+    if (this.isSuspended || Date.now() < this.nextPollTimeMs) {
+      return;
+    }
```
kevineger, line 148: The backoff is implemented by short-circuiting the poll method, but the runner is still ticking at `intervalMs` and invoking `poll()` on every interval. A couple of consequences:

- The effective wake-up cadence stays at `intervalMs` even when backoff is e.g. 5 minutes. The runner does work (albeit, cheap) every interval, ticks the abort signal lifecycle, etc.
- The actual delay between retries is `intervalMs`-rounded. With `intervalMs=5000` and first-failure `currentBackoffMs=5000`, the next retry happens at the next interval tick (5s) regardless. The exponential growth only manifests at intervals of `intervalMs`.

For real exponential backoff with quiet idle periods, consider stopping the runner on failure and using a `setTimeout` to restart it after `currentBackoffMs`.

src/jupyter/contents/directory-poller.ts
```diff
@@ -0,0 +1,368 @@
...
+    if (error instanceof ResponseError && error.response.status === 404) {
+      this.options.onDidChangeFile([
+        { type: this.options.vs.FileChangeType.Deleted, uri: this.options.uri },
+      ]);
+      this.dispose();
+      return;
+    }
+
+    log.warn(`Unable to poll ${this.options.uri.toString()}`, error);
+    this.nextPollTimeMs = Date.now() + this.currentBackoffMs;
```
kevineger: Most exponential backoff implementations double *before* scheduling so the first retry already has a longer delay than the baseline. It's also probably overkill here, but most have _jitter_.

src/jupyter/contents/directory-poller.ts
```diff
@@ -0,0 +1,368 @@
...
+   * @returns The updated reference count.
+   */
+  release(): number {
+    this.guardDisposed();
+    if (this.refCountValue === 0) {
+      return 0;
+    }
+    this.refCountValue -= 1;
+    if (this.refCountValue === 0) {
+      this.runner.stop();
```
kevineger: If a poll is in flight when `release()` brings the count to 0, the abort propagates to the request and the task rejects with an abort-type error. That error is **not** a `ResponseError` 404 so it hits the warn/backoff branch in `handlePollError`. The result: after the user has released all watches, the poller still mutates `nextPollTimeMs` and `currentBackoffMs` via the in-flight task that resolves after the synchronous reset on lines 131-132. If `start()` is called again later, the poller starts in a backed-off state with no obvious cause.

Severity is low (the abort error path runs only once and the values get re-clobbered on the next successful poll), but the snapshot/backoff reset here isn't actually authoritative.

src/jupyter/contents/directory-poller.ts
```diff
@@ -0,0 +1,368 @@
...
+function coalesceEvents(events: readonly FileChangeEvent[]): FileChangeEvent[] {
+  const seen = new Set<string>();
+  const coalesced: FileChangeEvent[] = [];
+  for (const event of events) {
+    const key = `${String(event.type)}:${event.uri.toString()}`;
+    if (seen.has(key)) {
+      continue;
+    }
+    seen.add(key);
+    coalesced.push(event);
```
kevineger: Sorry if I'm tracing this wrong, but what case does this defend against? I'm strugling to see how duplicates can be produced. Each path appears at most once in `previous` and at most once in `current`, and the type-flip branch produces (Deleted, Created) pairs which differ by `type`.

src/jupyter/contents/directory-poller.ts
```diff
@@ -0,0 +1,368 @@
...
+   * @returns The comparable snapshot entry.
+   */
+  private toSnapshotEntry(contents: Contents): SnapshotEntry {
+    return {
+      uri: this.options.vs.Uri.joinPath(this.options.uri, contents.name),
+      type: toFileType(this.options.vs, contents.type),
+      mtime: contents.lastModified
+        ? new Date(contents.lastModified).getTime()
+        : 0,
+      size: contents.size ?? 0,
```
kevineger, line 213: Should we key on the URI string (after normalization)?

src/jupyter/contents/directory-poller.ts
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  /** Resumes polling and refreshes immediately. */
+  resume(): void {
+    this.guardDisposed();
+    if (!this.isStarted || !this.isSuspended || this.refCountValue === 0) {
+      return;
+    }
+    this.isSuspended = false;
+    this.nextPollTimeMs = 0;
```
kevineger: `resume()` unconditionally resets `nextPollTimeMs = 0`, which means a resume during a long backoff window discards the backoff. If `suspend()` was called while we were backing off after repeated failures, `resume()` triggers an immediate poll attempt instead of respecting the remaining backoff. Probably fine for a user-initiated resume ("I want to see the latest now"), but is this the intended behavior?

If resume is ever triggered programmatically in response to a network/state event, you could thrash a failing endpoint. No action required if this is by design, if so just leave a comment in the docblock.

src/jupyter/contents/directory-poller.ts
```diff
@@ -0,0 +1,368 @@
...
+
+  /** Starts polling if there is at least one active registration. */
+  start(): void {
+    this.guardDisposed();
+    if (this.isStarted || this.refCountValue === 0) {
+      return;
+    }
+    this.isStarted = true;
+    this.runner.start(StartMode.Immediately);
+  }
```
kevineger, line 95: Nit: `start()` silently no-ops when `refCountValue === 0`. That's reasonable, but consider whether `start()` should call `addRef()` implicitly, or whether the contract is that `addRef()` is always called by an outer factory. If the latter, document on the class that `start()` requires at least one `addRef()` first. Otherwise debugging "why is my poller not polling?" can get confusing.

src/jupyter/contents/directory-poller.unit.test.ts
```diff
@@ -0,0 +1,291 @@
...
+
+    await advance();
+    sinon.assert.calledThrice(client.get);
+
+    await advance();
+    sinon.assert.callCount(client.get, 4);
+
+    await advance();
+    sinon.assert.callCount(client.get, 5);
+  });
```
kevineger, line 277: This test confirms that the backoff *throttles* polls but doesn't make obvious the timing-vs-interval interaction. With `intervalMs=5000` and the first failure setting `nextPollTimeMs` to `t+5000`, the next runner tick at `t+5000` actually executes the poll (because `Date.now() < nextPollTimeMs` is false at equality). The doubling kicks in only after the *second* failure. The expectations match the implementation, but a quick comment in the test labeling the wall-clock state at each `advance()` (e.g., `// t=10000: nextPoll=10000, executes`) would make this much easier to maintain when the backoff math changes.

src/jupyter/contents/directory-poller.unit.test.ts
```diff
@@ -0,0 +1,291 @@
...
+  afterEach(() => {
+    poller.dispose();
+    sinon.restore();
+  });
+
+  it('takes an initial snapshot without firing file events', async () => {
+    client.get.resolves(listing([file('a.txt')]));
+
+    poller.start();
+    await advance(0);
```
kevineger: This is effectively just flushign the micro-task queue. Are there any proper async mechanisms we can block on instead?

kevineger:
Thanks for sending this first PR!

Sorry it's taken me a while to get to, been pretty busy with some other work and its taken me a while to tease apart the lifecycle here.

---

I think we could simplify things a bit if the `DirectoryPoller` wasn't having to do three things at once:

1. A polling engine (snapshot --> diff --> events, with backoff)
2. A ref-counted shared resource (multiple `watch()` calls coalesce into one poll loop)
3. Its own lifecycle controller (start/suspend/resume/dispose)

What do you think of this split? I think it may align a bit closer along the VS Code seams we need to integrate with:

- `ContentsFileSystemProvider.watch(uri, opts)`: VS Code contract, returns `Disposable`
- `DirectoryWatcherRegistry`: owns `Map<authority+path, DirectoryPoller>`
  - `watch(uri): Disposable`: does the ref-counting; consumer never sees it
  - on auth/assignment events: suspend/resume/drop the entire map
- `DirectoryPoller`: just polls. Start in constructor, stop in `dispose()`.
  - no addRef/release
  - no start/suspend/resume on the public surface (or keep suspend/resume but no ref logic)
  - one job: tick, diff, emit, backoff

---

Sorry if this back and forth is more churn than you'd like to take on. If at any point you'd prefer I push commits and make the changes directly let me know.

src/jupyter/contents/directory-poller.ts
lndmanh: I don’t think _jitter_ is necessary for this single-user VS Code polling path, but I’ll make the retry sequence explicit in the tests and use the revised scheduling model consistently.

src/jupyter/contents/directory-poller.ts
lndmanh: I verified this race. An in-flight request can still complete after `release()`/stop resets local state. I’ll add an in-flight dispose/abort test and guard state mutation after awaits with `signal.aborted` / disposed checks.

src/jupyter/contents/directory-poller.ts
lndmanh: In my opinion, the current Jupyter path key is already sufficient for one poller. But keying by normalized emitted URI string should be clearer and better aligned with the events. I am considering to switch the snapshot map key to the child URI string. Do you have any thoughts on this?

src/jupyter/contents/directory-poller.ts
lndmanh: Current behavior intentionally probes immediately on resume, but that is not documented. Since focus regain is user-visible and the user likely wants a fresh check, I’d keep the immediate refresh behavior and document that it bypasses the pending retry delay while preserving the current backoff level for subsequent failures.

src/jupyter/contents/directory-poller.unit.test.ts
lndmanh: I’ll replace raw `advance(0)` calls with named helpers such as `flushInitialPoll()` / `flushMicrotasks()` so the async intent is explicit in each test.

src/jupyter/contents/directory-poller.ts
lndmanh: No you are not wrong. I don’t see a real duplicate case inside one directory diff either. I’ll remove `coalesceEvents()` from `DirectoryPoller`.

lndmanh:
I reviewed the lifecycle feedback against the current implementation. The main point is valid: `DirectoryPoller` is carrying polling, refcount sharing, and lifecycle ownership at once, and that makes the future provider wiring more awkward than it needs to be.

I revised this PR so `DirectoryPoller` only owns polling one directory and emitting diffs. Refcounting and map ownership would move to a follow-up `DirectoryWatcherRegistry` layer. For backoff, I’ll switch away from interval short-circuiting so failure backoff creates quiet idle periods instead of waking every base interval.

Kindly take a look at my newly changes and if anytime you want, just make the changes directly. Thanks for your kindly support again!

kevineger:
Please feel free to _re-request_ review (under the _Reviewers_ section) when I should take another look!

src/jupyter/contents/directory-poller.ts
```diff
@@ -0,0 +1,351 @@
...
+   *
+   * @param signal - Abort signal for the current scheduled poll.
+   */
+  private async poll(signal: AbortSignal): Promise<void> {
+    if (this.shouldIgnorePoll(signal)) {
+      return;
+    }
+
+    const client = await this.options.getClient();
+    if (!client || this.shouldIgnorePoll(signal)) {
```
kevineger, line 151: The "missing client is a no-op tick" behavior from the PR description doesn't look covered by a test. Worth one where `getClient` resolves `undefined` and we assert no events fire and the backoff stays put? It's easy to silently regress this into a spurious backoff otherwise.

src/jupyter/contents/directory-poller.ts
```diff
@@ -0,0 +1,351 @@
...
+   */
+  private async readSnapshot(
+    client: ContentsApi,
+    signal: AbortSignal,
+  ): Promise<DirectorySnapshot> {
+    const contents = await client.get(
+      { path: this.options.uri.path, type: ContentsGetTypeEnum.Directory },
+      { signal },
+    );
+    if (!isDirectoryContents(contents)) {
```
kevineger, line 184: A non-directory response throws here and lands in the backoff branch. In that case (e.g. a path that's actually a file or a transient wrong-type response), would we retry forever (capped at 5 min) and warn each cycle?

src/jupyter/contents/directory-poller.unit.test.ts
```diff
@@ -0,0 +1,345 @@
...
+    await clock.tickAsync(ms);
+  }
+
+  async function flushInitialPoll(): Promise<void> {
+    await clock.tickAsync(0);
+  }
+
+  async function flushMicrotasks(): Promise<void> {
+    await clock.tickAsync(0);
+  }
```
kevineger: Do you see any way around needing this microtask flushing? It can be very brittle and it's way better to rely on proper async mechanisms (e.g. deferred promises we control) where possible.

kevineger (changes requested):

src/jupyter/contents/directory-poller.ts
lndmanh: Agreed. I traced it: when `getClient()` resolves `undefined`, `poll()` returns before the try/catch, so nothing emits, `currentBackoffMs` is untouched, no retry is armed, and the base interval keeps ticking - but nothing locks that in. Adding two tests that prove it via cadence (matching the existing backoff tests rather than reaching into private state):
  - a no-op tick followed by a successful poll on the **next base interval** (proves no spurious backoff delay)
  - a no-op tick **after** a prior failure, asserting the elevated backoff is preserved (not reset, not re-armed).

src/jupyter/contents/directory-poller.unit.test.ts
lndmanh: `ConsumptionPoller`'s test already has the pattern I should be using here (same `SequentialTaskRunner` + immediate poll). I'll drop the `tickAsync(0)` pumping and synchronize on observable side-effects with our `Deferred` helper: resolve a deferred at the top of the `client.get` `callsFake` (poll *started*) and inside the change listener (diff *emitted*), then `await` those. Genuine time advancement stays as `tickAsync(INTERVAL)` (which also drains the synchronous tail), so there's no bare `tick(0)` left as a load-bearing synchronizer.

src/jupyter/contents/directory-poller.ts
lndmanh: Yes, and it's actually reachable via two paths, only one of which is this line:

  1. **Stock Jupyter (the common case):** `get({ type: 'directory' })` on a path that is now a file rejects with `ResponseError` **400, reason `"bad type"`** *before* this `isDirectoryContents` guard is even  reached. `handlePollError` only treats `404` as terminal, so the 400 falls through to `log.warn` + `scheduleRetry` → infinite retry capped at 5 min, warning every cycle.
  2. **This line:** a non-compliant custom `ContentsManager` could instead *resolve* a coerced file model, which throws the generic `Error` and lands in the same backoff branch.

  A file-where-a-directory-was is a stable state, not a transient one, so backoff can never recover. I'll treat it like the existing 404 terminal path: emit `Deleted(rootUri)`, fire `onDidTerminate()`, and stop (renaming `handleTerminalDelete` → `handleTerminalGone` since it now covers "deleted **or** no longer a directory"). I'll route the 400 in `handlePollError` and the coerced-model case here. Since this poller only ever sends `type=directory` with no `format`, the only 400 it can produce is `"bad type"`, so I'll scope the terminal check to `400`/`404` and leave `401/403/5xx` on the transient backoff path. Adding tests for both routes mirrored on the existing 404 test.

PR #613, opened 2026-06-11
test: update e2e test email selector

amtoney524:
This change updates the email selector in the e2e test to search for the string "email" in the aria-label (case insensitive)

The input type has changed recently, breaking the existing selector


kevineger (approved):
Thanks! Looks like the current failure (_TimeoutError: hello-world.txt did not appear in the Colab Contents view_) is an unrelated flake 🙁 .

PR #622, opened 2026-06-23
fix: failing `upload file` E2E test

hjjackyang:
**Why**? The `upload file` E2E test currently fails due to test setup constantly failing to clean up the server. This resulted in a dirty state when `upload file` test runs, causing test to assert the incorrect server.

After debugging extensively, I found the cause of the server cleanup failure was because the `listSessions` calls in `getServers` were stuck for ~3 min and then timed out for unowned servers. I don't know yet why they were stuck (more investigation is needed), but we only needed these `listSessions` calls for fetching the labels for unowned servers, so it's okay to force a failure for each call after a much shorter timeout (I made it 3s for now). For this purpose, I refactored and reused the existing `waitForTimeout` method from `code-manager.ts`.

---

In addition to the above change, I made 2 more improvements that don't directly contribute to the current E2E test failures:
1. Set `git.enabled` to false in E2E tests to get rid of the annoying Git notification dialog.
2. Made `sessions.list` and `sessions.delete` in `unassignServer` best-effort since it's not really mandatory to delete the session(s) before unassigning the server.

### DIFF
--- a/src/common/async.ts
+++ b/src/common/async.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import vscode from 'vscode';
 import { log } from './logging';
 
 /**
@@ -90,4 +91,40 @@ export function isPromiseLike(value: unknown): value is PromiseLike<unknown> {
     (typeof value === 'object' || typeof value === 'function') &&
     typeof (value as { then?: unknown }).then === 'function'
   );
+}
+
+/**
+ * A promise that can have its underlying resources (like timers or listeners)
+ * cleaned up.
+ */
+export interface DisposablePromise<T> extends vscode.Disposable {
+  /** The underlying promise that can be awaited. */
+  promise: Promise<T>;
+}
+
+/**
+ * Creates a promise that rejects after a specified timeout.
+ * Returns a disposable to clear the timer.
+ *
+ * @param ms - The delay duration in milliseconds.
+ * @param message - The error message for the timeout.
+ * @returns A disposable promise that rejects after the specified timeout.
+ */
+export function waitForTimeout(
+  ms: number,
+  message: string,
+): DisposablePromise<never> {
+  let timeoutId: NodeJS.Timeout;
+  const promise = new Promise<never>((_, reject) => {
+    timeoutId = setTimeout(() => {
+      reject(new Error(message));
+    }, ms);
+  });
+
+  return {
+    promise,
+    dispose: () => {
+      clearTimeout(timeoutId);
+    },
+  };
 }
--- a/src/common/async.unit.test.ts
+++ b/src/common/async.unit.test.ts
@@ -5,11 +5,11 @@
  */
 
 import { expect } from 'chai';
-import sinon from 'sinon';
+import sinon, { SinonFakeTimers } from 'sinon';
 import { Deferred } from '../test/helpers/async';
 import { ColabLogWatcher } from '../test/helpers/logging';
 import { newVsCodeStub } from '../test/helpers/vscode';
-import { LatestCancelable } from './async';
+import { LatestCancelable, waitForTimeout } from './async';
 import { LogLevel } from './logging';
 
 describe('LatestCancelable', () => {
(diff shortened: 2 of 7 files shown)


src/common/async.ts
kevineger, line 1: The public exported `waitForTimeout` should probably have tests.

src/jupyter/assignments.ts
```diff
@@ -500,13 +501,30 @@ export class AssignmentManager implements Disposable {
       return;
     }
     const client = ProxiedJupyterClient.withStaticConnection(server);
+    // Best-effort clean up sessions before unassigning the server. Without
+    // this, session won't immediately disconnect if there is a notebook
+    // attached in VS Code. However, we don't want to fail the entire
+    // unassignServer call because the unassign API call will eventually garbage
+    // collect and clean up the session(s) too.
```
```suggestion
    // Best-effort clean up sessions before unassigning the server. Without
    // this, sessions won't immediately disconnect if there are notebooks
    // attached in VS Code. However, we don't want to fail the entire
    // unassignServer call because the unassign API call will eventually garbage
    // collect and clean up the session(s) too.
```

src/jupyter/assignments.ts
```diff
@@ -500,13 +501,30 @@ export class AssignmentManager implements Disposable {
...
     await Promise.all(
```
kevineger: Consider abstracting promise chain to a function and use `async`/`await`.

src/jupyter/assignments.ts
```diff
@@ -500,13 +501,30 @@ export class AssignmentManager implements Disposable {
...
+        .list({ signal })
+        .catch((err: unknown) => {
+          // Swallow the sessions.list error as this is best-effort.
+          log.warn('Error occurred while listing sessions:', err);
+          return [];
+        })
+        .then((sessions) =>
+          sessions.map((session) =>
+            session.id
+              ? client.sessions.delete({ session: session.id }, { signal })
```
kevineger: Shouldn't deleting also be best effort?

src/jupyter/assignments.ts
```diff
@@ -937,3 +963,7 @@ function errorToAssignmentOutcome(error: unknown): AssignmentOutcome {
   }
   return AssignmentOutcome.ASSIGNMENT_OUTCOME_OTHER_FAILURE;
 }
+
+export const TEST_ONLY = {
```
kevineger: Nit: I'd prefer we optionally configure the class to accept this option, dup the timeout in the test file or export the underlying `const` (in that order).

src/jupyter/assignments.ts
```diff
@@ -500,13 +501,30 @@ export class AssignmentManager implements Disposable {
...
+    // collect and clean up the session(s) too.
     await Promise.all(
-      (await client.sessions.list({ signal })).map((session) =>
-        session.id
-          ? client.sessions.delete({ session: session.id }, { signal })
-          : Promise.resolve(),
-      ),
-    );
+      await client.sessions
+        .list({ signal })
```
kevineger: Should this also be promise raced against the 3s timeout?

src/jupyter/assignments.ts
```diff
@@ -729,11 +747,15 @@ export class AssignmentManager implements Disposable {
...
+              LIST_UNOWNED_SESSIONS_TIMEOUT_MS,
+              `Listing sessions timeout exceeded for endpoint ${a.endpoint}`,
+            );
             try {
-              const sessions = await this.client.listSessions(
-                a.endpoint,
-                signal,
-              );
+              const sessions = await Promise.race([
+                this.client.listSessions(a.endpoint, signal),
```
kevineger, line 732: This would be out of scope for this PR, but we should probably migrate all of these TFE based calls. Only mentioning it here since we use the `ProxiedJupyterClient` above in `unassignServer`.

kevineger (commented):
Thanks for tackling this!

src/jupyter/assignments.ts
hjjackyang: For now, this is not causing any issues because this `sessions.list` call only applies on VS Code owned servers. I'd prefer to leave it like this for now until it's causing issue.

src/jupyter/assignments.ts
hjjackyang: Yes, it is. The `catch` block on `Promise.all` will catch all `delete` errors.

src/jupyter/assignments.ts
hjjackyang: Ack. I can look into this as part of the public API migration later.

src/jupyter/assignments.ts
kevineger: I think it's problematic for a couple reasons: 

1. Dangling Promises: The remaining delete requests are still floating in the background, but the application has already moved on.
2. Unhandled Promise Rejections: If a second or third delete fails a few milliseconds later, their errors are no longer caught by the outer `.catch` (because the `Promise.all` has already settled). This will trigger `UnhandledPromiseRejection` warnings, which will lead to the extension having unhandled errors.
---

We should either catch individually or use `Promise.allSettled`.

src/jupyter/assignments.ts
kevineger: Fair enough.

src/jupyter/assignments.ts
hjjackyang: Went with duping the timeout in the test file.

src/jupyter/assignments.ts
hjjackyang: Oh, I did not know that! Changed to `Promise.allSettled`.

src/common/async.ts
hjjackyang: Done!

src/common/async.unit.test.ts
```diff
@@ -195,3 +195,42 @@ describe('LatestCancelable', () => {
     });
   });
 });
+
+describe('waitForTimeout', () => {
```
kevineger, line 199: Consider a happy path case (promise completes within timeout).

kevineger (approved):

src/common/async.unit.test.ts
hjjackyang: Is this necessary? `waitForTimeout` on its own just returns a promise that will always reject after the timeout. The happy path is only applicable when used in a `Promise.race` with another promise, which is already covered by tests in call sites. LMK if this makes sense.

PR #623, opened 2026-06-24
feat: generate Colab OpenAPI schema

hjjackyang:
Setting up VS Code extension to generate OpenAPI TS schema for Colab public API (still in beta). `openapi-typescript` is used for schema generation per https://openapi-ts.dev/introduction.

Opted to check in the `api-*.json` files to the repo for easier regeneration because the hosted OpenAPI schema doc is still restricted (not yet launched publicly).

Generated code + OpenAPI json files are exempted from lint, spell and format checks.

---

In the next PR, I will write the lightweight client code using `openapi-fetch` per https://openapi-ts.dev/openapi-fetch.

PR #624, opened 2026-06-26
feat: new ColabApiClient using openapi-fetch

hjjackyang:
Implemented new lightweight `ColabApiClient` using `openapi-fetch` and schemas generated in https://github.com/googlecolab/colab-vscode/pull/623.

See https://openapi-ts.dev/openapi-fetch for more details on `openapi-fetch`.

---

This PR is a no-op as the new `ColabApiClient` isn't called from anywhere yet.


src/colab/client/v2/index.ts
kevineger, line 1: Consider we're going to have a pretty long tail of private APIs (e.g. user info) around, the `src/colab/client.ts` and `src/colab/client/index.ts` distinction isn't very obvious. Should we be explicit? Considering they'll live alongside each other for a while too, maybe versioning our impl. makes sense?

`src/colab/client/v1/index.ts` and ``src/colab/client/v2/index.ts`?

Also open to explicit `src/colab/google_api_client.ts`.

src/colab/client/v2/index.ts
```diff
@@ -0,0 +1,219 @@
...
+import type {
+  paths as operationsPaths,
+  components as operationsComponents,
+} from './generated/schema-v1';
+import type {
+  paths as colabPaths,
+  components as colabComponents,
+} from './generated/schema-v1beta';
+
+/** A client for interacting with the Colab API. */
```
kevineger, line 52:
```suggestion
/** A client for interacting with the public Colab API. */
```

kevineger: Apply globally.

src/colab/client/index.ts
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  /**
+   * Gets the subscription for the authenticated user.
+   *
+   * @returns A promise that resolves to the subscription.
+   */
+  async getSubscription() {
+    const { data } = await this.colabClient.GET('/v1beta/subscription');
+    return data as unknown as colabComponents['schemas']['Subscription'];
```
kevineger: Is there no way around the `as unknown as` casts? I would have expected the generated client to be fully typed? Something feels off if not. Also surprised to see us rolling all these methods and constructing payloads manually then calling `.GET`.

src/colab/client/v2/index.ts
kevineger, line 1: Could we consider exposing [as we did for Jupyter](https://github.com/googlecolab/colab-vscode/blob/main/src/jupyter/client/index.ts)?

I really like the call-sites (e.g. `client.contents.get`).

kevineger (commented):
😆 can't wait to flatten `colabApiDomain`, ` colabGapiDomain` and `colabPublicApiDomain` one day...

src/colab/client/index.ts
hjjackyang: > Is there no way around the `as unknown as` casts? I would have expected the generated client to be fully typed? Something feels off if not.

Ya, I would expect that too, but for some reason, the `data` type is inferred as `undefined`, so I kinda have to explicitly cast it for my `ColabApiClient` methods to infer the correct type.

Per https://openapi-ts.dev/openapi-fetch:
> `data` and `error` are typechecked and expose their shapes to Intellisense in VS Code (and any other IDE with TypeScript support).

I will need to investigate further why it's not working correctly in my VS Code.

---

> Also surprised to see us rolling all these methods and constructing payloads manually then calling `.GET`.

Right. I did some research. There are some community libraries out there that can generate full OpenAPI SDK. However, https://openapi-ts.dev appears to be the most official and recommended one to use, with `openapi-typescript` to generate types and `openapi-fetch` to implement the client.

hjjackyang:
Abandoning this PR in favor of https://github.com/googlecolab/colab-vscode/pull/626.

PR #639, opened 2026-07-09
feat: migrate `assignServer` to `CreateRuntime`

hjjackyang:
Migrated `assignServer` to the new public `CreateRuntime` and `WaitOperation` APIs behind the `EnablePublicApi` feature gate.

[Local test screencast](https://screencast.googleplex.com/cast/NjI3MzczMzQ3OTEwNDUxMnw3NjYxZjI0NC1mYw)

A few things to note:
* The new `CreateRuntime` API generates a runtime ID for us. In order to use it, I had to loosen the `ColabJupyterServer.id` type from `UUID` to `string`. This change should be backward compatible when we switch on the `EnablePublicApi` feature gate. When I migrate `GetRuntime` and `ListRuntimes` in the future PRs, I can use `isUUID(ColabJupyterServer.id)` to differentiate servers assigned by old private API vs new public API.
* I added a new progress notification of `Assigning server...` while waiting for the `CreateRuntime` long-running operation for better UX in case long-running operation runs long 😉 .
* I installed a new `@grpc/grpc-js` package to avoid hardcoding the error code for `FAILED_PRECONDITION` (9) in our code. Happy to revert this bit and define our own constant if we feel it's unnecessary to import a new package for a constant.


src/colab/client/v2/index.ts
```diff
@@ -116,6 +143,66 @@ export function normalizeShape(shape: Shape): CommonShape {
...
+              'You have insufficient quota to assign this server.',
+            );
+          default:
+            if (
+              code === status.FAILED_PRECONDITION &&
+              accelerator &&
+              accelerator !== 'NONE'
+            ) {
+              throw new AcceleratorUnavailableError(accelerator);
+            }
```
kevineger, line 199: Can we doc the thrown errors in the doc block?

src/colab/client/v2/index.ts
```diff
@@ -189,3 +276,12 @@ class ErrorMiddleware implements Middleware {
     return Promise.resolve();
   }
 }
+
+function isErrorInfo(obj: unknown): obj is ErrorInfo {
```
kevineger, line 287: Any idea what's going on [here](https://github.com/googlecolab/colab-vscode/blob/a52b48f2a077e78b049ef6de8a0584f31172c227/src/colab/client/v2/generated/colab/models/ErrorInfo.ts#L83-L88)?

Are we supposed to override there? Is there a way to have it generate a proper type assertion instead of `return true;`?

src/colab/client/v2/index.unit.test.ts
```diff
@@ -1280,3 +1292,222 @@ describe('normalizeShape', () => {
...
+      name: OPERATION_NAME,
+      done: true,
+      response: {
+        name: 'runtimes/1',
+      },
+    };
+
+    throwIfOperationError(nonErrorOperation);
+
+    // No error is thrown. We are good!
```
kevineger: Use `not.to.throw`.

src/colab/client/v2/index.unit.test.ts
```diff
@@ -1280,3 +1292,222 @@ describe('normalizeShape', () => {
...
+    const errorOperation: Operation = {
+      name: OPERATION_NAME,
+      done: true,
+      error: {
+        code: status.FAILED_PRECONDITION,
+        message: ERROR_MESSAGE,
+        details: [
+          // Intentionally add an unrelated detail here to ensure that it will
+          // be ignored
+          { unrelatedKey: 'unrelated value' },
```
kevineger: Instead of duplicating this across all error type tests, could we just have one test which asserts unrelated details are ignored?

src/colab/errors.ts
```diff
@@ -52,3 +52,41 @@ export class InsufficientQuotaError extends Error {}
...
+    readonly reason = 'UNKNOWN',
+  ) {
+    super(
+      `Operation ${operationName} failed with error ${String(code)}: ${message} (reason: ${reason})`,
+    );
+  }
+}
+
+/** Error thrown when WaitOperation times out. */
+export class WaitOperationTimeoutError extends Error {
```
kevineger, line 79: Nit: should this extend `LongRunningOperationError` considering it must follow the same semantics (error info w/ same details)?

src/jupyter/assignments.ts
```diff
@@ -389,17 +413,39 @@ export class AssignmentManager implements Disposable {
...
-        assignment.runtimeProxyInfo,
-        new Date(),
-      );
+
+      let server: ColabAssignedServer;
+      if (instanceOfRuntime(assignmentOrRuntime)) {
+        assert(assignmentOrRuntime.name);
+        const runtimeId = trimPrefix(assignmentOrRuntime.name, 'runtimes/');
+        const c = assignmentOrRuntime.connectionInfo;
+        assert(c, `ConnectionInfo is missing in runtime: ${runtimeId}}`);
```
kevineger: [`instanceOfRuntime`](https://github.com/googlecolab/colab-vscode/blob/a52b48f2a077e78b049ef6de8a0584f31172c227/src/colab/client/v2/generated/colab/models/Runtime.ts#L75) is a type assertion (`value is Runtime`). Why is [`connectionInfo`](https://github.com/googlecolab/colab-vscode/blob/a52b48f2a077e78b049ef6de8a0584f31172c227/src/colab/client/v2/generated/colab/models/Runtime.ts#L43) `?:`?

Is it ever the case it's undefined? API missing an annotation? Or is it because the same object is used for req/resp?

src/jupyter/assignments.ts
```diff
@@ -906,12 +960,66 @@ export class AssignmentManager implements Disposable {
...
+        title: 'Assigning server...',
+        cancellable: false,
+      },
+      () => {
+        return this.colabApiClient.operations.waitOperation(
+          { operationsId: operationId, timeout: WAIT_OPERATION_TIMEOUT },
+          { signal },
+        );
+      },
+    )) as CreateRuntimeOperation;
```
kevineger: Shouldn't we deserialize instead of cast? I think there's a bug where in the LRO path `connectionInfo.expireTime` stays a `string`, not a `Date`.

`operations.waitOperation` returns the generic `Operation` type, whose `response` is deserialized as raw `json['response']` (see the generated `OperationFromJSON` in `operations/models/Operation.ts`), it doesn't go through `RuntimeFromJSON`. I think the `as CreateRuntimeOperation` cast masks this and `operation.response` is typed as `Runtime`, but the nested `connectionInfo.expireTime` is still the raw ISO string from the wire.

In general, let's not do any casts unless absolutely necessary.

src/jupyter/assignments.ts
```diff
@@ -906,12 +960,66 @@ export class AssignmentManager implements Disposable {
...
+    if (!operation.done) {
+      throw new WaitOperationTimeoutError(operationId, WAIT_OPERATION_TIMEOUT);
```
kevineger: `waitOperation` is called once with a 120s server-side timeout and no polling loop. If the runtime isn't ready within 120s, this throws `WaitOperationTimeoutError` and the whole assignment fails. Two concerns:

1. Is 120s a safe upper bound for GPU/TPU assignment latency? What was the "old" path limit?
2. On timeout, the runtime may still be created server-side afterward, leaving an orphaned runtime that counts toward `TOO_MANY_ACTIVE_RUNTIMES` and future assigns.

Should we poll `waitOperation` until `done` (with an overall deadline), and/or best-effort clean up the runtime on timeout (would like to avoid since I don't know if we support this in the API)?

src/jupyter/assignments.unit.test.ts
```diff
@@ -1136,484 +1172,1222 @@ describe('AssignmentManager', () => {
...
+                (req: WaitOperationRequest) =>
+                  req.operationsId === OPERATION_ID &&
+                  req.timeout === WAIT_OPERATION_TIMEOUT,
+              ),
+            )
+            .resolves({
+              name: `operations/${OPERATION_ID}`,
+              done: true,
+              response: defaultRuntime,
+            });
```
kevineger: This stub resolves `waitOperation` with `response: defaultRuntime`, whose `connectionInfo.expireTime` is already a `Date` (L129). The real `operations.waitOperation` returns `Operation` with `response` left as raw JSON, so `expireTime` arrives as a string. Because the stub hands back a pre-parsed `Date`, this test can't catch the missing `RuntimeFromJSON` deserialization in `createRuntime`'s LRO path (see my comment there). Consider returning wire-shaped JSON (string `expireTime`) here so the test exercises the real deserialization.

kevineger (changes requested):

src/colab/client/v2/index.ts
hjjackyang: I think this is because all fields in `ErrorInfo` are identified as optional fields, and I don't think there is anything we can do about it.

Taking `Runtime` as an example, since it contains a mandatory `runtimeSpec` field, `instanceOfRuntime` is generated properly by checking if `runtimeSpec` field is present:

https://github.com/googlecolab/colab-vscode/blob/aa4d563c1f9d326132829d9df365adcde8a2f0b2/src/colab/client/v2/generated/colab/models/Runtime.ts#L72-L78

src/jupyter/assignments.ts
hjjackyang: Yes, the same object is used for both req and resp, hence `connectionInfo` is generated as optional. We expect `connectionInfo` to be always present in a response, so can simply `assert` it.

src/colab/errors.ts
hjjackyang: `WaitOperationTimeoutError` indicates the operation is still *not* done after waiting for X amount of time. In this case, there is no error code or error info directly returned in the `Operation`.

I can probably fabricate one with `status.DEADLINE_EXCEEDS` to fit the `LongRunningOperationError` semantics, but I don't feel it's necessary.

src/colab/client/v2/index.ts
hjjackyang: Done! Doc block added.

src/colab/client/v2/index.unit.test.ts
hjjackyang: Done! Good catch! I could never remember `not.to.throw` exists.

src/colab/client/v2/index.unit.test.ts
hjjackyang: Done! Removed the duplicate test data and added a dedicated test case for it.

src/jupyter/assignments.ts
hjjackyang: Great catch!

I incorrectly assumed `Operation` and `CreateRuntimeOperation` were the same in this case and could be safely casted. Changed this to convert with `CreateRuntimeOperationFromJSON` instead.

src/jupyter/assignments.unit.test.ts
hjjackyang: Updated the test data with a raw `expireTime` string in this case.

src/jupyter/assignments.ts
hjjackyang: Based on the existing Prod metrics:
* TFE VM Assignment Latency: http://screen/3dPeiGSWTFPboxF
* OnePlat `CreateAssignment` RPC p99 Latency: http://screen/7RynhSKVfsUj8vB

The p99 latency are mostly below 2m (120s).

Looking further, I found that the deadline currently defined on our internal `CreateAssignment` RPC is [180s](http://google3/google/internal/colab/v1/runtime_service.proto;l=23;rcl=926279189). So I think we can align the FE wait to that internal deadline with some network buffer, e.g. `200s`.

If an operation hasn't finished within 200s, the operation is likely going to fail and we shouldn't need to worry about orphaned runtime problem.

WDYT?

kevineger: SG

hjjackyang: Updated timeout to 200s.

src/colab/client/v2/index.unit.test.ts
```diff
@@ -1280,3 +1292,214 @@ describe('normalizeShape', () => {
...
+      name: OPERATION_NAME,
+      done: true,
+      response: {
+        name: 'runtimes/1',
+      },
+    };
+
+    expect(() => {
+      throwIfOperationError(nonErrorOperation);
+    }).to.not.throw;
```
kevineger: `.to.not.throw` without the parens is just a property access, need to invoke with `.to.not.throw()`.

src/jupyter/assignments.ts
kevineger: Typo: `}}`.

src/jupyter/assignments.ts
```diff
@@ -923,12 +977,69 @@ export class AssignmentManager implements Disposable {
...
+        return this.colabApiClient.operations.waitOperation(
+          { operationsId: operationId, timeout: WAIT_OPERATION_TIMEOUT },
+          { signal },
+        );
+      },
+    );
+
+    createRuntimeOperation = CreateRuntimeOperationFromJSON(operation);
+    if (!createRuntimeOperation.done) {
+      throw new WaitOperationTimeoutError(operationId, WAIT_OPERATION_TIMEOUT);
```
kevineger, line 1025: Is this timeout path covered? The LRO test only exercises `waitOperation` resolving `done: true`. Worth a case where it resolves `done: false` so we assert this throws `WaitOperationTimeoutError`. It's a new error type and the "never completes" case should be covered.

src/jupyter/assignments.ts
```diff
@@ -923,12 +977,69 @@ export class AssignmentManager implements Disposable {
...
+      return createRuntimeOperation.response;
+    }
+
+    assert(createRuntimeOperation.name);
+    const operationId = trimPrefix(createRuntimeOperation.name, 'operations/');
+    const operation = await this.vs.window.withProgress(
+      {
+        location: this.vs.ProgressLocation.Notification,
+        title: 'Assigning server...',
+        cancellable: false,
```
kevineger, line 1013: Not actually sure what's right here, but we set `cancellable: false` yet pass the `signal`. Should we be wiring up cancellation w/ the signal?

kevineger (approved):

src/jupyter/assignments.ts
hjjackyang: Wow, good eye!

src/jupyter/assignments.ts
hjjackyang: Test case added.

src/jupyter/assignments.ts
hjjackyang: We definitely shouldn't make the progress `cancellable` until we can support a proper `CancelOperation` call. Otherwise, accidentally cancelling will result in orphaned runtime.

Will keep it as is because the `signal` is currently passed down from top of the call stack and is always `undefined` for the time being.

PR #641, opened 2026-07-10
fix: make Prettier checks work on Windows

dipeshbabu:
## Summary

- enforce LF checkouts for detected text files with a repository-level `.gitattributes` policy
- keep Prettier's default LF enforcement consistent on Windows and other platforms

## Why

With `core.autocrlf=true`, Windows checks out text files as CRLF when the repository has no line-ending policy. Prettier defaults to LF, so `npm run format:check` reports nearly every tracked text file even when the Git worktree is clean.

Adding `* text=auto eol=lf` makes the repository policy explicit: Git checks out detected text files as LF on every platform while leaving binary files untouched.

## Validation

- fresh Windows clone with global `core.autocrlf=true`: 295 text files checked out as LF; 4 binary files remained binary
- `npm run format:check`
- `npm run typecheck`
- `npm run lint` (0 errors; one warning in generated `src/colab-config.ts`)
- `npm run test:unit` (1,156 passing)
- CSpell check used by CI (199 files, 0 issues)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


kevineger:
The default is `lf` ([docs](https://prettier.io/docs/options#end-of-line)). If we switch to `auto` then won't the repo develop inconsistent line endings?

Would the proper fix be step 4 from the [docs](https://prettier.io/docs/options#end-of-line)? _Add * text=auto eol=lf to the repo’s .gitattributes file. You may need to ask Windows users to re-clone your repo after this change to ensure git has not converted LF to CRLF on checkout._

dipeshbabu:
@kevineger Good point, thanks. I replaced `endOfLine: auto` with a root `.gitattributes` containing `* text=auto eol=lf`, so Prettier continues to enforce consistent LF endings.

I validated the amended commit from a fresh Windows clone with global `core.autocrlf=true`: all 295 detected text files checked out as LF, the 4 binary files remained binary, and `npm run format:check` passed.

kevineger (approved):
Right on, thanks for the contribution 🤙

PR #646, opened 2026-07-20
fix: `removeServer` to call Jupyter `sessions.list` instead of TFE

hjjackyang:
**Why**? During public API migration, I found that the current `listSessions` call via TFE endpoint was consistently hanging / failing (likely started since cl/922783892). Since TFE is on the deprecation path and this API is also available as part of the standard [Jupyter REST API](https://jupyter-server.readthedocs.io/en/latest/developers/rest-api.html#get--api-sessions), I figure it is best to migrate.

This change is safe because the existing call is already consistently hanging/failing and this call in `removeServer` to get unowned servers is considered best-effort, hence doing it without a feature flag.

Screencasts:
* [Before](https://screencast.googleplex.com/cast/NjU0ODQyNTQ5NDQ5NTIzMnwzYzFjZGY3MC03OA)
* [After](https://screencast.googleplex.com/cast/NjcyNDEwNDY1NTczMjczNnxjMDFhYjAxNS1hNg)

---

As part of this PR, I also:
* Cleaned up the old `listSessions` implementation since `removeServer` was the only call site.
* Changed `ProxiedJupyterClient.withStaticConnection` signature to be more generic to accept connection info from different source shapes.
* Added a trace log in `removeServer` to log all resolved servers for future tracing and debugging purposes.

### DIFF
--- a/src/colab/commands/server.ts
+++ b/src/colab/commands/server.ts
@@ -5,6 +5,7 @@
  */
 
 import vscode, { QuickPickItem } from 'vscode';
+import { log } from '../../common/logging';
 import { MultiStepInput } from '../../common/multi-step-quickpick';
 import { AssignmentManager } from '../../jupyter/assignments';
 import { ContentsFileSystemProvider } from '../../jupyter/contents/file-system';
@@ -143,6 +144,7 @@ export async function removeServer(
     source ?? CommandSource.COMMAND_SOURCE_COMMAND_PALETTE,
   );
   const allServers = await assignmentManager.getServers('all');
+  log.trace(`removeServer: getServers('all') resolved:`, allServers);
   const vsCodeServers = allServers.assigned;
   const nonVsCodeServers = allServers.unowned;
   if (vsCodeServers.length === 0 && nonVsCodeServers.length === 0) {
--- a/src/jupyter/assignments.ts
+++ b/src/jupyter/assignments.ts
@@ -30,7 +30,6 @@ import {
 import { REMOVE_SERVER } from '../colab/commands/constants';
 import {
   AcceleratorUnavailableError,
-  ColabRequestError,
   DenylistedError,
   InsufficientQuotaError,
   NotFoundError,
@@ -44,6 +43,7 @@ import {
 import { Shape, SubscriptionTier, Variant } from '../colab/types';
 import { waitForTimeout } from '../common/async';
 import { log } from '../common/logging';
+import { FetchError as JupyterFetchError } from '../jupyter/client/generated';
 import { telemetry } from '../telemetry';
 import { AssignmentOutcome, CommandSource } from '../telemetry/api';
 import { ProxiedJupyterClient } from './client';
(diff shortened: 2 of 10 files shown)


src/colab/commands/server.ts
```diff
@@ -143,6 +144,7 @@ export async function removeServer(
     source ?? CommandSource.COMMAND_SOURCE_COMMAND_PALETTE,
   );
   const allServers = await assignmentManager.getServers('all');
+  log.trace(`removeServer: getServers('all') resolved:`, allServers);
```
kevineger, line 147: Del?

src/jupyter/assignments.ts
```diff
@@ -755,31 +755,35 @@ export class AssignmentManager implements Disposable {
...
-              const sessions = await Promise.race([
-                this.colabClient.listSessions(a.endpoint, signal),
-                timeout.promise,
-              ]);
-              label =
-                sessions.length === 1 && sessions[0].name?.length
-                  ? sessions[0].name
-                  : UNKNOWN_REMOTE_SERVER_NAME;
+              const rp = a.runtimeProxyInfo;
+              if (rp) {
```
kevineger: Nit: consider inverting conditional to reduce nesting.

src/jupyter/assignments.unit.test.ts
```diff
@@ -545,6 +555,96 @@ describe('AssignmentManager', () => {
...
   describe('getServers', () => {
+    const ENDPOINT_WITH_SESSION_NAME = 'test-endpoint-with-name';
+    const ENDPOINT_WITHOUT_SESSION_NAME = 'test-endpoint-without-name';
+    const ENDPOINT_WITHOUT_SESSION = 'test-endpoint-without-session';
+    const URL_WITH_SESSION_NAME = 'https://test.url.with.session.name';
+    const URL_WITHOUT_SESSION_NAME = 'https://test.url.without.session.name';
+    const URL_WITHOUT_SESSION = 'https://test.url.without.session';
+    const PROXY_TOKEN = 'test-proxy-token';
+    const TEST_SESSION_NAME = 'test-session-name';
+    const UNKNOWN_REMOTE_SERVER_NAME = 'Untitled';
```
kevineger: Nit: remove these `const`-s and embed them in the objects below (e.g. `const assignmentWithName`). Then the other callsites will read cleanly as `assignmentWithName.endpoint`.

src/jupyter/assignments.unit.test.ts
```diff
@@ -545,6 +555,96 @@ describe('AssignmentManager', () => {
...
+      path: '',
+      type: '',
+      kernel: {
+        lastActivity: '',
+        executionState: '',
+        id: '',
+        name: '',
+        connections: 1,
+      },
+    };
```
kevineger, line 596: Nit: reduce LOC by object destructuring and expanding with what makes the assignment unique. Always helps with at-a-glance readability to understand the variants.

kevineger (approved):

src/colab/commands/server.ts
hjjackyang: I actually intentionally added and kept this (also called out specifically in the PR description) because I kinda find this trace log valuable to keep for future tracing and debugging purposes.

Let me know if you think otherwise.

src/jupyter/assignments.ts
hjjackyang: I initially did the other way around to avoid duplicating the `return`. Done!

src/jupyter/assignments.unit.test.ts
hjjackyang: Done. Good suggestion!

src/jupyter/assignments.unit.test.ts
hjjackyang: Done! Reused the existing `defaultAssignment` object to fill the new objects to reduce LOC.

PR #653, opened 2026-07-25
feat: migrate `refreshConnection` to `GetRuntime`

hjjackyang:
Migrated `AssignmentManager.refreshConnection` to the new public `GetRuntime` API.

Unlike the other API migrations, this one is *not directly* gated by the `EnablePublicApi` feature flag, but relies on the stored `server.id` format to differentiate whether the server was assigned by the old private API (v1 client) or the new public API (v2 client). We only use the new public `GetRuntime` API to refresh the connection info if the server was assigned by the new public `CreateRuntime` API to reduce complexity and ensure consistency.

---

Local test screenshot:
GetRuntime_local_test


src/jupyter/assignments.ts
```diff
@@ -564,10 +565,25 @@ export class AssignmentManager implements Disposable {
...
     }
-    const newConnectionInfo = await this.colabClient.refreshConnection(
-      server.endpoint,
-      signal,
-    );
+
+    let newConnectionInfo: RuntimeProxyToken | ConnectionInfo;
+    // If the id is in UUID format, it means the server was assigned by the old
+    // v1 client. Use the v1 client to refresh the connection in this case.
+    if (isUUID(id)) {
```
kevineger, line 551: Does this work?

`createRuntime` doesn't set `runtimeId`, and the public API docs say an unset `runtime_id` gets "a random UUID" generated for it. Wouldn't `isUUID` be true for both?

src/jupyter/assignments.ts
```diff
@@ -564,10 +565,25 @@ export class AssignmentManager implements Disposable {
...
+        server.endpoint,
+        signal,
+      );
+    } else {
+      // Otherwise, use the new v2 client to get a fresh connection info.
+      const runtime = await this.colabApiClient.colab.getRuntime(
+        { runtime: id },
+        { signal },
+      );
+      assert(runtime.connectionInfo);
```
kevineger: A message here would help. This fires inside the background `ConnectionRefresher`, where the user sees nothing, so an `AssertionError` with no context is rough to debug.

src/jupyter/assignments.unit.test.ts
```diff
@@ -3189,82 +3190,127 @@ describe('AssignmentManager', () => {
...
+        name: 'with Public API disabled',
+        enablePublicApi: false,
+        server: defaultServer,
+      },
+      {
+        name: 'with Public API enabled',
+        enablePublicApi: true,
+        server: defaultServerV2,
+      },
+    ];
```
kevineger, line 3107: The flag doesn't affect the branch under test. Dispatch is purely on id format, and the flag only changes which list API `reconcileAssignedServers` calls. These effectively read as "v1 id" vs "v2 id".

Flag off with an `r-` prefixed id isn't covered.

src/jupyter/assignments.unit.test.ts
```diff
@@ -3189,82 +3190,127 @@ describe('AssignmentManager', () => {
...
+              (
+                colabApiClientStub.colab.listRuntimes as sinon.SinonStub
+              ).resolves({ runtimes: [defaultRuntime] });
+              (colabApiClientStub.colab.getRuntime as sinon.SinonStub)
+                .withArgs(
+                  sinon.match(
+                    (req: GetRuntimeRequest) => req.runtime === server.id,
+                  ),
+                  sinon.match.any,
+                )
```
kevineger: `sinon.match({ runtime: server.id })` does the same thing and reads more concisely

src/jupyter/assignments.unit.test.ts
```diff
@@ -3189,82 +3190,127 @@ describe('AssignmentManager', () => {
...
+              connectionInformation: {
+                ...server.connectionInformation,
+                headers: {
+                  [COLAB_RUNTIME_PROXY_TOKEN_HEADER.key]: newToken,
+                  [COLAB_CLIENT_AGENT_HEADER.key]:
+                    COLAB_CLIENT_AGENT_HEADER.value,
+                },
+                token: newToken,
+              },
+            };
```
kevineger, line 3191: `expectedServer` reuses `server.connectionInformation.tokenExpiry`, and the stubbed `expireTime` is the same `NOW + TOKEN_EXPIRY_MS`, so a regression that dropped the refreshed expiry still passes. Bumping the stub's `expireTime` would better protect `refreshConnection`.

src/jupyter/assignments.ts
hjjackyang: The public API doc is a bit outdated on this one. I'm actually fixing it to say "random ID" instead of "random UUID" in cl/953970389.

The new `runtime_id` is generated as `r-${UUID}`, so should fail the `isUUID` test.

src/jupyter/assignments.ts
hjjackyang: Added a message.

src/jupyter/assignments.unit.test.ts
hjjackyang: Added 2 more test variables to cover [flag ON, v1 server] and [flag OFF, v2 server] cases.

src/jupyter/assignments.unit.test.ts
hjjackyang: TIL! Done.

src/jupyter/assignments.unit.test.ts
hjjackyang: Done. Bumped `expireTime` in test case.

src/jupyter/assignments.ts
kevineger: Couldn't a user specify a UUID in the public API tho?

hjjackyang: Yes, they could in public API, but my VS Code implementation to create runtime does not pass in a user-specified `runtimeId`, so the assigned servers will all have `r-${UUID}` runtime IDs.

https://github.com/googlecolab/colab-vscode/blob/af9192a1b3283d8ffb9ba44c10af1caf7f3d6727/src/jupyter/assignments.ts#L986-L999

src/jupyter/assignments.ts
```diff
@@ -542,10 +543,28 @@ export class AssignmentManager implements Disposable {
...
     }
-    const newConnectionInfo = await this.colabClient.refreshConnection(
-      server.endpoint,
-      signal,
-    );
+
+    let newConnectionInfo: RuntimeProxyToken | ConnectionInfo;
+    // If the id is in UUID format, it means the server was assigned by the old
+    // v1 client. Use the v1 client to refresh the connection in this case.
+    if (isUUID(id)) {
```
kevineger, line 551: Sorry if I'm misreading, but is `resolveJupyterServer` inconsistent? E.g. see [test](https://github.com/googlecolab/colab-vscode/blob/47ef0891ecfc01ea2430094b391b74c91b2a9704/src/jupyter/provider.unit.test.ts#L268-L274).

Resolve is defensively implemented since it _shouldn't_ be called if `provideJupyterServers` returns the connection info, but the call sites should agree.

src/jupyter/assignments.ts
```diff
@@ -542,10 +543,28 @@ export class AssignmentManager implements Disposable {
...
       throw new NotFoundError('Server is not assigned');
     }
-    const newConnectionInfo = await this.colabClient.refreshConnection(
-      server.endpoint,
-      signal,
-    );
+
+    let newConnectionInfo: RuntimeProxyToken | ConnectionInfo;
+    // If the id is in UUID format, it means the server was assigned by the old
+    // v1 client. Use the v1 client to refresh the connection in this case.
```
kevineger, line 550: Worth a short comment next to `createRuntime` warning not to set `runtimeId` since our detection relies on it?

src/jupyter/assignments.ts
```diff
@@ -542,10 +543,28 @@ export class AssignmentManager implements Disposable {
...
+        server.endpoint,
+        signal,
+      );
+    } else {
+      // Otherwise, use the new v2 client to get a fresh connection info.
+      const runtime = await this.colabApiClient.colab.getRuntime(
+        { runtime: id },
+        { signal },
+      );
+      assert(
```
kevineger, line 562: There's already a `runtimeWithoutConnectionInfo` fixture covering the equivalent `getServers` paths. `listRuntimes` needs to keep returning `defaultRuntime` (add test?) so reconcile doesn't prune the server and throw `NotFoundError` before we ever reach here.

src/jupyter/assignments.ts
```diff
@@ -542,10 +543,28 @@ export class AssignmentManager implements Disposable {
...
+        runtime.connectionInfo,
+        `ConnectionInfo missing in runtime: ${id}`,
```
kevineger: Re-use `MISSING_CONNECTION_INFO_ERR_MSG`?

src/jupyter/assignments.unit.test.ts
```diff
@@ -3021,82 +3021,153 @@ describe('AssignmentManager', () => {
...
+              sinon.assert.calledOnce(colabClientStub.refreshConnection);
+              sinon.assert.notCalled(
+                colabApiClientStub.colab.getRuntime as sinon.SinonStub,
+              );
+            } else {
+              sinon.assert.notCalled(colabClientStub.refreshConnection);
+              sinon.assert.calledOnce(
+                colabApiClientStub.colab.getRuntime as sinon.SinonStub,
+              );
+            }
```
kevineger: Can we make these their own tests?

kevineger (approved):

src/jupyter/assignments.ts
hjjackyang: Done. Removed `isUUID` check in `provideJupyterServers`.

src/jupyter/assignments.ts
hjjackyang: Done.

src/jupyter/assignments.ts
hjjackyang: I'm not fully understanding this comment since the `refreshConnection` -> `with a refreshed connection` tests already return a `defaultRuntime` from `listRuntimesStub`.

Resolving this for now and will follow up with @kevineger after he is back from OOO.

src/jupyter/assignments.ts

src/jupyter/assignments.unit.test.ts
hjjackyang: Moved these to a more global `afterEach`, and asserted `listRuntimesStub`/`colabClientStub.listAssignments` as well.

PR #660, opened 2026-08-03
feat: migrate `unassignServer` to `DeleteRuntime`

hjjackyang:
Migrated `AssignmentManager.unassignServer` to the new public `DeleteRuntime` API.

Similar to `GetRuntime` in https://github.com/googlecolab/colab-vscode/pull/653, for a VS Code assigned server, we call the new `DeleteRuntime` API if the server was originally assigned by the new public API (v2 client) based on the `server.id` format.

For an unowned server (assigned elsewhere), we call the new `DeleteRuntime` API based on the `EnablePublicApi` feature flag. In order for `DeleteRuntime` API to work for the unowned servers, we had to update the old `ListAssignments` API to return `notebookIdHash` and store it in `UnownedServer.id`. This is safe because `notebookIdHash` is already always returned in `ListAssignments` API. We just never read it on client side.

---

### Local test screenshots

Deleting a VS Code assigned server:
DeleteRuntime_assigned_local_test

Deleting a web assigned server (can tell from notebook ID hash format):
DeleteRuntime_unowned_local_test


src/jupyter/assignments.ts
```diff
@@ -610,7 +618,19 @@ export class AssignmentManager implements Disposable {
...
     await this.deleteSessions(server, signal);
-    await this.colabClient.unassign(server.endpoint, signal);
+
+    // If the id is in UUID format, it means the server was assigned by the old
+    // v1 client. Use the v1 client to unassign in this case.
+    if (isUUID(server.id)) {
+      await this.colabClient.unassign(server.endpoint, signal);
+    } else {
+      // Otherwise, use the new v2 client to delete the runtime.
+      await this.colabApiClient.colab.deleteRuntime(
```
amtoney524, line 628: nit: should this be flag guarded too?

amtoney524 (approved):

src/jupyter/assignments.ts
hjjackyang: Not really. This is *indirectly* guarded by the flag because `CreateRuntime` is guarded by the flag already.

As I mentioned in the PR description, similar to https://github.com/googlecolab/colab-vscode/pull/653, I did this because servers assigned by the old `assign` API vs new `CreateRuntime` API have different ID formats. It would be unnecessarily complex (4 code branches) to rely on the flag and then code different delete logic for the old vs new ID formats, whereas the current implementation (relying on just the ID formats) has only 2 code branches.

Hope this makes sense!

PR #669, opened 2026-08-17
feat: log `experiment_ids` in all telemetry events

hjjackyang:
* **Why**? With selected `experiment_ids` logged, when I start ramping `EnablePublicApi` feature, I can easily monitor error rates between the *treatment* and *control* arms.
* This PR depends on cl/966174941 and cl/966077262 to be rolled out to Prod first.

---

Internal tracking bug: b/523943653

### DIFF
--- a/src/colab/client/v1/api.ts
+++ b/src/colab/client/v1/api.ts
@@ -599,6 +599,7 @@ export const ExperimentStateSchema = z.object({
       return new Map(entries) as Map<ExperimentFlag, ExperimentFlagValue>;
     })
     .optional(),
+  selectedIds: z.array(z.int32()).optional(),
 });
 /** The experiment state response. */
 export type ExperimentState = z.infer<typeof ExperimentStateSchema>;
--- a/src/colab/client/v1/api.ts
+++ b/src/colab/client/v1/api.ts
@@ -599,7 +599,8 @@ export const ExperimentStateSchema = z.object({
       return new Map(entries) as Map<ExperimentFlag, ExperimentFlagValue>;
     })
     .optional(),
-  selectedIds: z.array(z.int32()).optional(),
+  /** The selected experiment IDs. */
+  selectedIds: z.array(z.number()).optional(),
 });
 /** The experiment state response. */
 export type ExperimentState = z.infer<typeof ExperimentStateSchema>;
--- a/src/colab/experiment-state.ts
+++ b/src/colab/experiment-state.ts
@@ -156,9 +156,7 @@ function setFlagForTest(
  *
  * @param ids - The experiment IDs to override.
  */
-function setExperimentIdsForTest(
-  ids: readonly number[],
-): void {
+function setExperimentIdsForTest(ids: readonly number[]): void {
   experimentIds = ids;
 }
 
--- a/src/colab/experiment-state.ts
+++ b/src/colab/experiment-state.ts
@@ -25,6 +25,15 @@ export function getFlag(flag: ExperimentFlag): ExperimentFlagValue {
   return flags.get(flag) ?? EXPERIMENT_FLAG_DEFAULT_VALUES[flag];
 }
 
+/**
+ * Gets the list of selected experiment IDs.
+ *
+ * @returns A list of experiment IDs.
+ */
+export function getExperimentIds(): readonly number[] {
+  return experimentIds;
+}
+
 const REFRESH_INTERVAL_MS = 5 * 60 * 1000; // 5 minutes.
 const REFRESH_TIMEOUT_MS = 30 * 1000; // 30 seconds.
 
@@ -44,6 +53,7 @@ export class ExperimentStateProvider implements Toggleable, Disposable {
   constructor(private readonly client: ColabClient) {
     // Reset experiment flags.
     flags = new Map<ExperimentFlag, ExperimentFlagValue>();
+    experimentIds = [];
     this.refreshPoller = new SequentialTaskRunner(
       {
         intervalTimeoutMs: REFRESH_INTERVAL_MS,
@@ -114,6 +124,7 @@ export class ExperimentStateProvider implements Toggleable, Disposable {
       );
       if (result.experiments) {
         flags = result.experiments;
+        experimentIds = result.selectedIds ?? [];
         log.trace(
           `Experiment state updated while ${requireAccessToken ? 'authorized' : 'not authorized'}:`,
           Object.fromEntries(flags),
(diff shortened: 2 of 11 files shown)


src/colab/experiment-state.ts
```diff
@@ -114,6 +124,7 @@ export class ExperimentStateProvider implements Toggleable, Disposable {
       );
       if (result.experiments) {
         flags = result.experiments;
+        experimentIds = result.selectedIds ?? [];
```
kevineger: Should this move outside the `if (result.experiments)`?

`experiments` is optional, so a response that has `selectedIds` but no flag overrides would leave `experimentIds` at its previous value.

src/telemetry/index.ts
```diff
@@ -217,6 +218,7 @@ function log(event: ColabEvent) {
     ...baseLog,
     ...event,
     timestamp: new Date().toISOString(),
+    experiment_ids: getExperimentIds(),
```
kevineger, line 221: Should we set `experiment_ids` at flush time so startup errors flow?

[`log` stringifies the event at queue time](https://github.com/googlecolab/colab-vscode/blob/50a436261bee680b972efd81aa0448200669264f/src/telemetry/client.ts#L113) and [`flush` is gated on `EnableTelemetry`](https://github.com/googlecolab/colab-vscode/blob/50a436261bee680b972efd81aa0448200669264f/src/telemetry/client.ts#L134-L137), which [defaults to false](https://github.com/googlecolab/colab-vscode/blob/50a436261bee680b972efd81aa0448200669264f/src/colab/client/v1/api.ts#L569) until the first experiment fetch lands. Telemetry also initializes at [extension.ts#L40](https://github.com/googlecolab/colab-vscode/blob/50a436261bee680b972efd81aa0448200669264f/src/extension.ts#L40), before the Colab client exists, so activation events get queued with an empty array baked into the JSON and delivered once the arm is already known.

src/colab/client/v1/api.ts
```diff
@@ -599,6 +599,7 @@ export const ExperimentStateSchema = z.object({
       return new Map(entries) as Map<ExperimentFlag, ExperimentFlagValue>;
     })
     .optional(),
+  selectedIds: z.array(z.int32()).optional(),
```
kevineger: We use `z.number()` everywhere else. Does this need to be `z.int32()` / should it be?

Also, please docblock the field.

src/colab/experiment-state.unit.test.ts
```diff
@@ -105,6 +116,17 @@ describe('ExperimentStateProvider', () => {
     expect(getFlag(ExperimentFlag.RuntimeVersionNames)).to.deep.equal([]);
   });
 
+  it('returns empty experiment IDs when no IDs are returned', async () => {
+    const getExperimentStateRanPromise = stubGetExperimentStateResponse(
+      new Map(),
+    );
```
kevineger: This is the same case as above. Should we instead have a case with `experiments` undefined? Guards `selectedIds` when the backend sends no flag overrides.

src/telemetry/telemetry.unit.test.ts
```diff
@@ -133,12 +135,17 @@ describe('Telemetry Module', () => {
...
+    let baseLog: ColabLogEventBase & {
+      timestamp: string;
+      experiment_ids: readonly number[];
+    };
     let logStub: SinonSpy;
 
     beforeEach(() => {
       logStub = sinon.stub(ClearcutClient.prototype, 'log');
       sinon.stub(process, 'platform').get(() => PLATFORM);
+      EXPERIMENT_TEST_ONLY.setExperimentIdsForTest(EXPERIMENT_IDS);
```
kevineger: Should we change them between two `log` calls and assert each event separately?

These are set before `initializeTelemetry`, so an implementation that snapshotted the IDs once at init would still pass.

kevineger (changes requested):

src/colab/experiment-state.ts
hjjackyang: Oh good catch! Fixed.

src/colab/client/v1/api.ts
hjjackyang: No, I didn't think too much when I wrote `z.int32()` because the data type in proto was `int32`. Changed to `z.number()` for consistency.

Also added missing docblock.

src/colab/experiment-state.unit.test.ts
hjjackyang: Refactored tests to be parameterized to cover all scenario variations.

src/telemetry/index.ts
hjjackyang: I have a different opinion in this. IMO, we should log the experiment IDs that are evaluated at the time of the event. Given `flush` occurs at a later time, there is a chance that the selected experiment IDs are different at `flush` time from when they were evaluated at `log` time.

In terms of startup errors, I think it's fine to never attribute them to any experiment IDs because experiment state isn't available at the startup time.

WDYT?

src/telemetry/telemetry.unit.test.ts
hjjackyang: Added a separate `logs experiment IDs when available` test case to cover this.

For simplicity, I reverted other test cases to simply log a default empty `experiment_ids` array.

src/colab/experiment-state.unit.test.ts
```diff
@@ -47,36 +48,86 @@ describe('ExperimentStateProvider', () => {
...
+  });
 
-    await expect(getExperimentStateRanPromise).to.eventually.be.fulfilled;
-    sinon.assert.calledOnceWithExactly(
-      colabClientStub.getExperimentState,
-      false,
-      sinon.match.any,
-    );
-    expect(getFlag(ExperimentFlag.RuntimeVersionNames)).to.be.false;
+  describe('when auth turned off', () => {
```
kevineger: Should we fold the auth dimension into the table too?

[`index.unit.test.ts` already does that for this endpoint](https://github.com/googlecolab/colab-vscode/blob/72ad6abb3af25fa536428cfdd76af63d0c2bb563/src/colab/client/v1/index.unit.test.ts#L977-L986), and it would drop the second copy of the body.

src/telemetry/index.ts
kevineger: Ah, SG.

Anything gated on the flag can only run after the fetch, since `getFlag` returns the default until then. A pre-fetch event can't have been affected by the arm, so there is nothing to attribute.

kevineger (approved):

src/colab/experiment-state.unit.test.ts
hjjackyang: I'm not sure how for this one. This `ExperimentStateProvider` is implemented as a `Toggleable`, so the auth on/off is controlled via the `on()`/`off()` method call.

kevineger: Fine either way. But I think just branching logic in the body would suffice.

hjjackyang: Oh, I see what you mean. Done!

PR #681, opened 2026-08-31
refactor: clean up `EnablePublicApi` feature flag + unused private API calls

hjjackyang:
Post public API migration rollout, this PR cleans up the fully launched `EnablePublicApi` flag and unused code.

For a peace of mind, I went through each and every call site, and manually cleaned up without the help of agents.

---

**Validations**:
* ✅ All tests and pre-merge checks are green.
* ✅ GitHub Copilot agent review came back promising: https://paste.googleplex.com/4803984506880000.
* ✅ Sanity tested multiple scenarios locally with no issues.

---

Internal tracking bug: b/555768009

### DIFF
--- a/src/colab/client/v1/api.ts
+++ b/src/colab/client/v1/api.ts
@@ -24,16 +24,7 @@
  */
 
 import { z } from 'zod';
-import { Kernel as GeneratedKernel } from '../../../jupyter/client/generated';
 import { Shape, SubscriptionTier, Variant } from '../../types';
-
-export enum SubscriptionState {
-  UNSUBSCRIBED = 1,
-  RECURRING = 2,
-  NON_RECURRING = 3,
-  PENDING_ACTIVATION = 4,
-  DECLINED = 5,
-}
 
 enum ColabSubscriptionTier {
   UNKNOWN = 0,
@@ -48,25 +39,10 @@ enum ColabGapiSubscriptionTier {
   PRO_PLUS = 'SUBSCRIPTION_TIER_PRO_PLUS',
 }
 
-export enum Outcome {
-  UNDEFINED_OUTCOME = 0,
-  QUOTA_DENIED_REQUESTED_VARIANTS = 1,
-  QUOTA_EXCEEDED_USAGE_TIME = 2,
-  // QUOTA_EXCEEDED_USAGE_TIME_REFUND_MIGHT_UNBLOCK (3) is deprecated.
-  SUCCESS = 4,
-  DENYLISTED = 5,
-}
-
 enum ColabGapiVariant {
   UNSPECIFIED = 'VARIANT_UNSPECIFIED',
   GPU = 'VARIANT_GPU',
   TPU = 'VARIANT_TPU',
-}
-
-enum ColabGapiShape {
-  UNSPECIFIED = 'SHAPE_UNSPECIFIED',
-  STANDARD = 'SHAPE_DEFAULT',
-  HIGHMEM = 'SHAPE_HIGH_MEM',
 }
 
 /** Colab supported auth types. */
@@ -113,21 +89,6 @@ function normalizeVariant(variant: ColabGapiVariant): Variant {
   }
 }
 
-/**
- * Normalize the similar but different GAPI representation for the shape.
- *
- * @param shape - the machine shape as represented by the Colab Google API.
- * @returns the normalized machine shape.
- */
-function normalizeShape(shape: ColabGapiShape): Shape {
-  switch (shape) {
-    case ColabGapiShape.HIGHMEM:
-      return Shape.HIGHMEM;
-    default:
-      return Shape.STANDARD;
-  }
-}
-
 export const Accelerator = z.object({
   /** The variant of the assignment. */
   variant: z.enum(ColabGapiVariant).transform(normalizeVariant),
@@ -140,31 +101,20 @@ export const Accelerator = z.object({
 
 /**
  * The schema for top level information about a user's tier, usage and
- * availability in Colab.
+ * availability in Colab when CCU consumption info is requested (consumption
+ * fields are required).
  */
-export const UserInfoSchema = z.object({
+export const ConsumptionUserInfoSchema = z.object({
   /** The subscription tier. */
   subscriptionTier: z
     .enum(ColabGapiSubscriptionTier)
     .transform(normalizeSubTier),
   /** The paid Colab Compute Units balance. */
-  paidComputeUnitsBalance: z.number().optional(),
+  paidComputeUnitsBalance: z.number(),
   /** The eligible machine accelerators. */
   eligibleAccelerators: z.array(Accelerator),
   /** The ineligible machine accelerators. */
   ineligibleAccelerators: z.array(Accelerator),
-});
-/** Colab user information. */
-export type UserInfo = z.infer<typeof UserInfoSchema>;
-
-/**
- * The schema for top level information about a user's tier, usage and
- * availability in Colab when CCU consumption info is requested (consumption
- * fields are required).
- */
-export const ConsumptionUserInfoSchema = UserInfoSchema.required({
-  paidComputeUnitsBalance: true,
-}).extend({
   /**
    * The current rate of consumption of the user's CCUs (paid or free) based on
    * all assigned VMs.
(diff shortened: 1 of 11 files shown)


kevineger:
Nice! :rocket:  I'll wait a couple weeks so we can monitor stability of the API switch before reviewing and merging this.

src/jupyter/assignments.ts
```diff
@@ -545,31 +473,16 @@ export class AssignmentManager implements Disposable {
...
-        { signal },
-      );
-      assert(
-        runtime.connectionInfo,
-        `${MISSING_CONNECTION_INFO_ERR_MSG}: ${id}`,
-      );
-      newConnectionInfo = runtime.connectionInfo;
-    }
+    const runtime = await this.colabApiClient.colab.getRuntime(
+      { runtime: id },
```
kevineger, line 475: Should a stored server with a legacy v1 UUID id still resolve here?

`main` routed those to [`colabClient.refreshConnection`](https://github.com/googlecolab/colab-vscode/blob/2ead262407943772103b0230285c0906aabc7672/src/jupyter/assignments.ts#L551-L556) and, in `unassignServer`, to [`colabClient.unassign`](https://github.com/googlecolab/colab-vscode/blob/2ead262407943772103b0230285c0906aabc7672/src/jupyter/assignments.ts#L621-L622), and this drops both with no storage migration. A pre-launch entry that survives `reconcileStoredServers` would now hit `getRuntime({ runtime: <uuid> })`, and `isUUID` is left behind with only test callers.

src/jupyter/assignments.ts
```diff
@@ -369,36 +331,19 @@ export class AssignmentManager implements Disposable {
...
-              await this.colabClient.assign(
-                id,
-                { variant, accelerator, shape, version },
-                signal,
-              ));
-          }
+          runtime = await this.createRuntime(descriptor, signal);
         }
       } catch (error) {
         log.trace(`Failed assigning server ${id}`, error);
```
kevineger: Should we drop `id` now that it only feeds this log line?

It used to be the notebook hash passed to v1 `assign`, and the stored server is now keyed off `runtimeId` from `runtime.name`, so this logs a UUID that matches nothing.

src/jupyter/assignments.unit.test.ts
```diff
@@ -1454,1855 +1164,1174 @@ describe('AssignmentManager', () => {
...
+          },
+        });
 
-        expect(
-          assignmentManager.assignServer(defaultAssignmentDescriptor),
-        ).to.be.rejectedWith(/connection info/);
-      });
+      expect(
+        assignmentManager.assignServer(defaultServerDescriptor),
+      ).to.be.rejectedWith(/connection info/);
```
kevineger: Should we `await` these (and the `token: ''` twin below) and set `connectionInfo: undefined`?

Neither `it` is `async` and neither `expect` is awaited (the disposed-manager test above does `await expect(...)`), so `rejectedWith` never runs. `{ ...defaultRuntime.connectionInfo, url: '' }` is truthy anyway, so [`assert(c, ...)`](https://github.com/googlecolab/colab-vscode/blob/948ec562d278e7a8ab57b4a34fa88096f691a3ed/src/jupyter/assignments.ts#L374) never fires.

src/jupyter/provider.ts
```diff
@@ -188,15 +183,8 @@ export class ColabJupyterServerProvider
...
-        let tier: SubscriptionTier;
-        if (enablePublicApi) {
-          const subs = await this.colabApiClient.colab.getSubscription();
-          tier = normalizeSubscriptionTier(subs.tier);
-        } else {
-          tier = (await this.colabClient.getUserInfo()).subscriptionTier;
-        }
-        if (tier === SubscriptionTier.NONE) {
+        const subs = await this.colabApiClient.colab.getSubscription();
+        if (subs.tier === SubscriptionTier.SubscriptionTierFree) {
```
kevineger, line 187: Should we keep `normalizeSubscriptionTier` and compare against the common `SubscriptionTier`?

It's the only one of the four v2 normalizers this PR removes, and `assignments.ts` still [imports](https://github.com/googlecolab/colab-vscode/blob/948ec562d278e7a8ab57b4a34fa88096f691a3ed/src/jupyter/assignments.ts#L22-L25) `normalizeVariant` / `normalizeShape`. Its unit test went with it, including the `SUBSCRIPTION_TIER_UNSPECIFIED` case.

src/colab/client/v1/api.ts
```diff
@@ -24,17 +24,8 @@
...
 
-export enum SubscriptionState {
-  UNSUBSCRIBED = 1,
-  RECURRING = 2,
-  NON_RECURRING = 3,
-  PENDING_ACTIVATION = 4,
-  DECLINED = 5,
-}
-
 enum ColabSubscriptionTier {
```
kevineger: Should `ColabSubscriptionTier` go too?

src/jupyter/provider.ts
```diff
@@ -69,7 +66,6 @@ export class ColabJupyterServerProvider
    * @param vs - The VS Code API instance.
    * @param authEvent - The authentication event emitter.
    * @param assignmentManager - The assignment manager instance.
-   * @param colabClient - The old Colab private API client instance.
    * @param colabApiClient - The new Colab public API client instance.
```
kevineger: Nit, drop "new" now that v1 is gone. Do global pass throughout.

src/jupyter/assignments.unit.test.ts
kevineger: You need to `await` these assertions or they won't fire. Sorry this was missed, please do full scan.

kevineger (changes requested):

src/jupyter/assignments.ts
kevineger: On second thought... since runtimes only last up to a day, is this even possible?

We've had the extension on the public API / new client for > 1 day, so if any 1+ plus runtimes are persisted they'll get pruned when they're read from storage. In other words, a v1 UUID ID can't make it here, can it?

kevineger: Yeah, pretty sure it's not. Can ignore my comment.

src/colab/client/v1/api.ts
hjjackyang: No, this is still referenced by `ConsumptionUserInfoSchema`.

https://github.com/googlecolab/colab-vscode/blob/948ec562d278e7a8ab57b4a34fa88096f691a3ed/src/colab/client/v1/api.ts#L113-L117

src/jupyter/assignments.ts
hjjackyang: Done. Good catch!

src/jupyter/provider.ts
hjjackyang: For this one, since the result isn't propagated anywhere outside of the method, I would prefer to avoid normalizing the value and simply check the raw value returned for simplicity. This is simpler code and saves us one unnecessary value conversion.

For variant/shape, since their corresponding normalized values are referenced in many places across the codebase, it's better to keep them as common API abstractions to avoid massive updates.

src/jupyter/provider.ts
hjjackyang: Done. Turns out this is the only place that needs this update. 2 more other places still reference both old and new client instances.

src/jupyter/assignments.unit.test.ts
hjjackyang: Good eye!

Added `.eventually.` to many existing assertions which already `await`ed.

For this particular one, the new client code actually never checks for URL or token individually. After adding `await` and `.eventually.`, the 2 tests started failing. Instead, I consolidated the 2 tests to one which simply checks if the entire `ConnectionInfo` object is missing (with the proper await).

src/colab/client/v1/api.ts
kevineger: Should this still go? My last comment was too terse, sorry.

`ConsumptionUserInfoSchema` does reach `normalizeSubTier`, but it feeds it `z.enum(ColabGapiSubscriptionTier)` at [L115-L117](https://github.com/googlecolab/colab-vscode/blob/714a19c998e2214d3d5cd6a6e9432d597f412e07/src/colab/client/v1/api.ts#L115-L117), and that is the only caller. With `PostAssignmentResponseSchema` gone nothing can produce a `ColabSubscriptionTier` value, so this enum and the `L70` / `L73` cases are unreachable.

src/colab/client/v1/index.ts
```diff
@@ -185,147 +130,6 @@ export class ColabClient {
...
-          ) {
-            throw new AcceleratorUnavailableError(
-              params.accelerator ?? 'default',
-            );
-          }
-          throw error;
-        }
-
-        switch (res.outcome) {
-          case Outcome.QUOTA_DENIED_REQUESTED_VARIANTS:
```
kevineger, line 247: Does the public API surface an equivalent reason for this one?

[`throwIfOperationError`](https://github.com/googlecolab/colab-vscode/blob/714a19c998e2214d3d5cd6a6e9432d597f412e07/src/colab/client/v2/index.ts#L169-L186) maps `QUOTA_EXCEEDED_USAGE_TIME` but has no `QUOTA_DENIED_REQUESTED_VARIANTS` case, so that denial would fall through to `LongRunningOperationError` and `notifyInsufficientQuota` never fires.

src/jupyter/assignments.unit.test.ts
```diff
@@ -1454,1855 +1164,1140 @@ describe('AssignmentManager', () => {
...
+              spec.accelerator === defaultRuntime.runtimeSpec.accelerator &&
+              req.runtime?.version === defaultRuntime.version
+            );
+          }),
+        )
+        .resolves({
+          done: true,
+          response: {
+            ...defaultRuntime,
+            connectionInfo: undefined,
```
kevineger, line 1219: Should we add the `name: undefined` sibling while we're here?

`assert(runtime.name, MISSING_RUNTIME_NAME_ERR_MSG)` sits one line above the `connectionInfo` assert and has no test (`runtimeWithoutName` only ever reaches `listRuntimesStub`, never `createRuntimeStub`).

src/jupyter/assignments.unit.test.ts
```diff
@@ -1454,1855 +1164,1140 @@ describe('AssignmentManager', () => {
...
+        await assignmentManager.assignServer(defaultServerDescriptor);
 
-          await assignmentManager.assignServer(defaultAssignmentDescriptor);
+        sinon.assert.calledOnceWithExactly(
+          logStub,
+          AssignmentOutcome.ASSIGNMENT_OUTCOME_SUCCEEDED,
+          {
+            variant: defaultServerDescriptor.variant,
+            accelerator: defaultServerDescriptor.accelerator ?? '',
+            shape: 'STANDARD',
```
kevineger, line 1684: Should we port `logs the requested shape and version when present` into this suite?

It was the only test asserting `shape: 'HIGHMEM'` and it lived in the deleted flag-off block, so `Shape[shape]` is now exercised only for `STANDARD` and `''`.

src/jupyter/provider.unit.test.ts
```diff
@@ -320,116 +316,69 @@ describe('ColabJupyterServerProvider', () => {
...
+
+          assert.isDefined(commands);
+          expect(commands.map((c) => c.label)).to.deep.equal([
+            buildIconLabel(AUTO_CONNECT),
+            buildIconLabel(NEW_SERVER),
+            buildIconLabel(OPEN_COLAB_WEB),
+          ]);
+        });
+
+        it('returns commands to auto-connect, create a server, open Colab web and upgrade to pro for free users', async () => {
```
kevineger, line 358: Should we add a fourth case for `SUBSCRIPTION_TIER_UNSPECIFIED`?

Its only coverage was the `normalizeSubscriptionTier` block this PR deletes, which pinned that unspecified is not treated as free.

src/colab/client/v1/api.ts
kevineger: Sorry if missing something.

`ConsumptionUserInfoSchema` does reach `normalizeSubTier`, but it feeds it `z.enum(ColabGapiSubscriptionTier)` at [L115-L117](https://github.com/googlecolab/colab-vscode/blob/714a19c998e2214d3d5cd6a6e9432d597f412e07/src/colab/client/v1/api.ts#L115-L117), and that is the only caller. With `PostAssignmentResponseSchema` gone nothing can produce a `ColabSubscriptionTier` value, so this enum and the `L70` / `L73` cases are unreachable.

src/colab/client/v1/index.ts
hjjackyang: Looks like the new private API does not surface a `QUOTA_DENIED_REQUESTED_VARIANTS` reason anymore.

Perhaps it is already folded into `QUOTA_EXCEEDED_USAGE_TIME` reason?

@peap - Is this something that you could help confirm?

src/jupyter/assignments.unit.test.ts
hjjackyang: Done! Added a sibling name missing test.

src/jupyter/assignments.unit.test.ts
hjjackyang: Great catch! I must've accidentally deleted this test. Added back.

src/jupyter/provider.unit.test.ts
hjjackyang: Done! Added a fourth case for `SUBSCRIPTION_TIER_UNSPECIFIED` with a comment to note that it should not occur in reality, but covered for completeness.

src/colab/client/v1/api.ts
hjjackyang: Yet again, you are right! I did not pay attention to the actual usage and implementation of `normalizeSubTier`, and was just blindly trimming based on code references.

`ColabSubscriptionTier` is now gone!

src/jupyter/provider.unit.test.ts
hjjackyang: Actually, I just refactored these 3 tests to a single dynamic test to cover the 3 non-free tiers, and omitted the extra comment on the `SUBSCRIPTION_TIER_UNSPECIFIED` case.

src/jupyter/assignments.ts
```diff
@@ -153,56 +143,28 @@ export class AssignmentManager implements Disposable {
...
-      if (!isHighMemOnlyAccelerator(descriptor.accelerator)) {
-        proDescriptors.push({ ...descriptor, shape: Shape.STANDARD });
-      }
-      proDescriptors.push({ ...descriptor, shape: Shape.HIGHMEM });
-    }
-    return proDescriptors;
+    // The new ListRuntimeSpecs API already takes user's subscription tier
+    // into account, returning with the correct eligibility info. The new API
+    // also returns additional high-memory shapes for the Pro users, so we
+    // don't need to manually add them.
```
kevineger: Can we reword the comment now that the old API path is gone?

kevineger (approved):
Thanks for tackling this! Huge LOC so hopefully I didn't miss anything. If we want to defer figuring out the API error thing we can, since that path is already live. But would be nice to double check all that.

src/jupyter/assignments.ts
hjjackyang: Done. Thanks again for the thorough review! This is a huge PR and super tough to review!

src/colab/client/v1/index.ts
peap: The `QUOTA_DENIED_REQUESTED_VARIANTS` reason was removed internally several months ago, as it was related to a long-gone feature. I don't think the RuntimeService ever returned it.

hjjackyang: Thanks for confirming!

PR #686, opened 2026-09-02
fix: redact and truncate request errors

kevineger:
Error messages embedded the full request URL and the entire response body, so telemetry received query param values and whole HTML error pages (some quite large).

### DIFF
--- a/src/colab/errors.ts
+++ b/src/colab/errors.ts
@@ -7,6 +7,36 @@
 import { Request, Response } from 'node-fetch';
 
 /**
+ * Cap on the response body carried in an error message.
+ *
+ * Bodies can be entire HTML error pages, and the message is reported verbatim
+ * to telemetry.
+ */
+const MAX_BODY_CHARS = 512;
+
+/**
+ * Drops the query string, which can carry notebook hashes and auth hints.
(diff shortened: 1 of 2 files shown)


src/colab/errors.ts
```diff
@@ -6,6 +6,36 @@
...
+const MAX_BODY_CHARS = 512;
+
+/**
+ * Drops the query string, which can carry notebook hashes and auth hints.
+ *
+ * @param url - The URL to redact.
+ * @returns The URL without its query string.
+ */
+function redactUrl(url: string): string {
+  return url.split(/[?#]/)[0];
```
hjjackyang: Nit: It's probably more readable to parse the url as `URL`, clear its `search`, and return the `url.toString()`.

hjjackyang (approved):

src/colab/errors.ts
kevineger: Re-_hashed_ to set `REDACTED` for the URL search param values, and fragment
value. If unparseable, splits everything after the first `?` or `#` as orignally
done.

hjjackyang: LGTM!

PR #687, opened 2026-09-02
fix: stop reporting cancellations as errors

kevineger:
Aborted in-flight work and user-dismissed prompts are normal control flow.

### DIFF
--- a/src/telemetry/index.ts
+++ b/src/telemetry/index.ts
@@ -9,6 +9,7 @@ import { Disposable } from 'vscode';
 import { AuthType } from '../colab/client/v1/api';
 import { getExperimentIds } from '../colab/experiment-state';
 import { SubscriptionTier as ColabSubscriptionTier } from '../colab/types';
+import { InputFlowAction } from '../common/multi-step-quickpick';
 import { COLAB_EXT_IDENTIFIER } from '../config/constants';
 import { getPackageInfo } from '../config/package-info';
 import { JUPYTER_EXT_IDENTIFIER } from '../jupyter/jupyter-extension';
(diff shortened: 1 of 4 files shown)


src/telemetry/index.ts
```diff
@@ -133,6 +147,9 @@ export const telemetry = {
     });
   },
   logError: (e: unknown) => {
+    if (isCancellation(e)) {
```
hjjackyang: Nit: Should we put this gate in `createProcessErrorHandler` where the uncaught errors are logged?

https://github.com/googlecolab/colab-vscode/blob/ab4d682f6e86d8d59ae2ea7a36b00dc4107f1616/src/telemetry/process-errors.ts#L9-L28

I think we should still allow such errors to be logged if we ever intentionally want to log one of these errors from somewhere else. Thoughts?

src/telemetry/index.ts
```diff
@@ -33,6 +34,19 @@ let baseLog: ColabLogEventBase;
...
 let isTelemetryEnabled: () => boolean;
 
+/**
+ * A set of error names that represent cancelled or abandoned work.
+ *
+ * These are normal control flow. E.g. a user dismissing a picker or a poll
+ * being superseded. Errors with these names are not reported to the telemetry
+ * service as they are not indicative of a problem with the extension.
+ */
+const CANCELLATION_ERROR_NAMES = new Set([
```
hjjackyang: Nit `ReadonlySet` for deep immutability?

hjjackyang (approved):

src/telemetry/index.ts
kevineger: Good point, done.

src/telemetry/index.ts
kevineger: Done.

PR #691, opened 2026-09-08
fix: don't resurrect a cleared session after a token refresh

kevineger:
`refreshSessionIfNeeded` guarded on `this.session` and then wrote back `{ ...this.session, accessToken }` on the far side of an `await`. If the session was cleared while the refresh was in flight (e.g. a sign out racing `getSessions`) the spread ran against `undefined`. Spreading `undefined` yields `{}`, so the write-back resurrected the session as a partial object holding only `accessToken`.

That object is truthy, so it survives the `!this.session` guard in `getSessions`, and the next scoped lookup evaluates `this.session?.scopes.includes(r)` against an undefined `scopes`, raising "Cannot read properties of undefined (reading 'includes')". The corruption is sticky: the partial session also has no `id`, so `removeSession` and `signOut` no longer match it and cannot clear it. Every subsequent `getSessions` call throws until the window reloads, which is why a handful of affected users produced a large, sustained volume of this error.

TypeScript did not catch it because narrowing a field survives an `await` in the type system but not at runtime. Snapshot the session into a local instead of relying on that narrowing, abandon the write-back when the field no longer holds the session we set out to refresh, and re-read it after the refresh before returning.

### DIFF
--- a/src/auth/auth-provider.ts
+++ b/src/auth/auth-provider.ts
@@ -218,11 +218,15 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
   ): Promise<AuthenticationSession[]> {
     this.guardDisposed();
     this.assertReady();
+    // Snapshot the session rather than narrowing `this.session` directly:
+    // narrowing a field survives the `await` below in the type system, but not
+    // at runtime.
+    const current = this.session;
     if (
-      !this.session ||
+      !current ||
       !areScopesAllowed(scopes) ||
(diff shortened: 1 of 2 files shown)


hjjackyang (approved):
I checked out the branch locally and tested the following 2 scenarios:
1. Signed it when running from `main` branch, switched to this branch and verified the signed-in session was retained.
2. Performed regular sign-out and sign-in when running locally from this branch.

PR #692, opened 2026-09-09
feat: migrate `propagateCredentials` to OnePlatform API behind flag

hjjackyang:
Migrate `propagateCredentials` to call the new OnePlatform API behind the `enable_op_credprop_api_vscode` feature flag.

I intentionally isolated the changes and branching logic in `colab/client/v1/index.ts` to avoid touching too many places. Unlike the old API, the new API returns the OAuth redirect URL in `ErrorInfo` with a 400 status instead of in a 200 success response.

See http://go/paste/4895389396893696 for a sample error response containing the OAuth redirect URL.

---

**Local test screenshot**:

credprop_local_test

---

Internal tracking bug: b/556251396

### DIFF
--- a/src/colab/client/v1/api.ts
+++ b/src/colab/client/v1/api.ts
@@ -75,7 +75,7 @@ export enum AuthType {
   AUTH_USER_EPHEMERAL = 'auth_user_ephemeral',
 }
 
-/** Colab supported auth types. */
+/** Colab supported access token types. */
 export enum AccessTokenType {
   DFS_EPHEMERAL = 'ACCESS_TOKEN_TYPE_DFS_EPHEMERAL',
   AUTH_USER_EPHEMERAL = 'ACCESS_TOKEN_TYPE_AUTH_USER_EPHEMERAL',
--- a/src/colab/client/v1/api.ts
+++ b/src/colab/client/v1/api.ts
@@ -604,7 +604,7 @@ export const EXPERIMENT_FLAG_DEFAULT_VALUES: Record<
   ExperimentFlag,
   ExperimentFlagValue
 > = {
-  [ExperimentFlag.EnableOpCredentialPropagationApi]: true,
+  [ExperimentFlag.EnableOpCredentialPropagationApi]: false,
   [ExperimentFlag.EnablePublicApi]: false,
   [ExperimentFlag.EnableTelemetry]: false,
   [ExperimentFlag.ResourcePollIntervalMs]: 10000,
--- a/src/colab/client/v1/api.ts
+++ b/src/colab/client/v1/api.ts
@@ -75,6 +75,12 @@ export enum AuthType {
   AUTH_USER_EPHEMERAL = 'auth_user_ephemeral',
 }
 
+/** Colab supported auth types. */
+export enum AccessTokenType {
+  DFS_EPHEMERAL = 'ACCESS_TOKEN_TYPE_DFS_EPHEMERAL',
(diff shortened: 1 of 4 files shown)


src/colab/client/v1/api.ts
```diff
@@ -496,10 +502,42 @@ export const CredentialsPropagationResultSchema = z
     ...rest,
     unauthorizedRedirectUri: unauthorized_redirect_uri,
   }));
+/** Type of the Colab credentials propagation API result. */
 export type CredentialsPropagationResult = z.infer<
   typeof CredentialsPropagationResultSchema
 >;
 
+/** OnePlatform API error schema. */
+export const OnePlatformErrorSchema = z.object({
```
kevineger, line 511: Is there a generated model we can use instead of redefining this? Or is it a
unique object to the private OP API?

E.g. `Status`?

src/colab/client/v1/index.ts
```diff
@@ -395,6 +428,66 @@ export class ColabClient {
...
+        success: true,
+        unauthorizedRedirectUri: undefined,
+      };
+    } catch (e: unknown) {
+      if (
+        e instanceof ColabRequestError &&
+        e.response.status === 400 &&
+        e.responseBody
+      ) {
+        const errorBody = JSON.parse(e.responseBody) as unknown;
```
kevineger:
```suggestion
        const errorBody: unknown = JSON.parse(e.responseBody);
```

src/colab/client/v1/index.ts
```diff
@@ -395,6 +428,66 @@ export class ColabClient {
...
+            return {
+              success: false,
+              unauthorizedRedirectUri:
+                detail.metadata?.unauthorized_redirect_uri,
+            };
+          }
+        }
+      }
+      throw e;
+    }
```
kevineger: Is there ever a valid case where multiple error reasons would be returned and
**one** of them would be `OAUTH_CONSENT_REQUIRED`? If so, should we even extract
the URI? I think we can probably switch to look for a single error detail. I
think the code would also benefit from some factoring. Something like this:

```ts
function redirectUri(e: unknown): string | undefined {
  if (
    !(e instanceof ColabRequestError) ||
    e.response.status !== 400 ||
    !e.responseBody
  ) {
    return undefined;
  }
  const errorBody = JSON.parse(e.responseBody) as unknown;
  if (!isOnePlatformError(errorBody)) {
    return undefined;
  }
  const details = errorBody.error.details ?? [];
  if (
    details.length !== 1 ||
    !isErrorInfo(details[0]) ||
    details[0].reason !== 'OAUTH_CONSENT_REQUIRED'
  ) {
    return undefined;
  }
  return details[0].metadata?.unauthorized_redirect_uri;
}
```

and 

```suggestion
    } catch (e: unknown) {
      const uri = redirectUri(e);
      if (!uri) {
        throw e;
      }
      return {
        success: false,
        unauthorizedRedirectUri: uri,
      };
  }
```

src/colab/client/v1/api.ts
hjjackyang: Thanks for bringing this up!

I didn't bother doing extensive research since it's just 2 very lightweight Zod schemas. A quick agent research suggests nothing easily reusable out there:

agent_response

src/colab/client/v1/index.ts
hjjackyang: Done!

src/colab/client/v1/index.ts
hjjackyang: Done!

Also updated tests to cover the single-error-detail scenario.

kevineger (approved):

PR #698, opened 2026-09-10
fix: back off consumption polling after repeated failures

kevineger:
The poll retried every 60s indefinitely, so one badly-networked user generated thousands of identical failures a day. Consecutive failures now skip an exponentially growing number of intervals, capped at 15 and jittered; success, an assignment change, or toggling off clears it.

Aborts from the runner superseding or disposing a poll are not counted. They don't  change the network conditions, and counting them swallowed the replacement poll. A timeout still counts, since a hanging request is the case this exists for.

### DIFF
--- a/src/colab/consumption/poller.ts
+++ b/src/colab/consumption/poller.ts
@@ -9,6 +9,7 @@ import {
   OverrunPolicy,
   SequentialTaskRunner,
   StartMode,
+  TimeoutError,
 } from '../../common/task-runner';
 import { Toggleable } from '../../common/toggleable';
 import { AssignmentChangeEvent } from '../../jupyter/assignments';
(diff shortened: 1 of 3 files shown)


src/colab/consumption/poller.ts
```diff
@@ -17,6 +18,36 @@ import { ConsumptionUserInfo } from '../client/v1/api';
 
 const POLL_INTERVAL_MS = 1000 * 60; // 1 minute.
 const TASK_TIMEOUT_MS = 1000 * 10; // 10 seconds.
+/**
+ * Ceiling on the consecutive intervals skipped while backing off, so a user who
+ * is simply offline issues ~4 requests an hour rather than 60.
+ */
+const MAX_BACKOFF_INTERVALS = 15;
```
hjjackyang: Nit: Just to make it clear in the name that this is 15 minutes.

```suggestion
const MAX_BACKOFF_INTERVALS_MIN = 15;
```

src/colab/consumption/poller.ts
```diff
@@ -17,6 +18,36 @@ import { ConsumptionUserInfo } from '../client/v1/api';
...
+/**
+ * Reports whether a failed poll says anything about the network.
+ *
+ * The runner aborts polls it has superseded or disposed, which is routine. A
+ * timeout is different: the request really is hanging and we should back off.
+ *
+ * @param signal - The signal the poll ran under.
+ * @returns True when the failure should count towards the backoff.
+ */
+function countsAsFailure(signal?: AbortSignal): boolean {
```
hjjackyang: Nit: Make it a bit more explicit in the function name that we are trying to determine a network failure.

```suggestion
function countsAsNetworkFailure(signal?: AbortSignal): boolean {
```

src/colab/consumption/poller.ts
```diff
@@ -17,6 +18,36 @@ import { ConsumptionUserInfo } from '../client/v1/api';
...
+ */
+const MAX_BACKOFF_INTERVALS = 15;
+
+/**
+ * Number of poll intervals to sit out after a run of failures.
+ *
+ * @param failures - Consecutive failures so far, at least one.
+ * @returns The number of intervals to skip before trying again.
+ */
+function backoffIntervals(failures: number): number {
```
hjjackyang: Nit: Unlike Go, I believe TypeScript still prefers naming function with an action verb prefix?

```suggestion
function calculateBackoffIntervals(failures: number): number {
```

src/colab/consumption/poller.ts
hjjackyang, line 26: Nit: The number `15` threw me off a bit when I was initially reading this, thinking this means 15 minutes. After reading the implementation further, I then realize this is actually 15 intervals, which gets translated to 15 minutes because our interval happens to be 1 minute.

I wonder if it's worth making this a bit more clear in the TSDoc with some verbiage of: `Given our 1-minute poll interval, max backoff intervals of 15 means a max 15-minute backoff.`

WDYT?

hjjackyang (approved):
LGTM with just a few naming and documentation nits!

src/colab/consumption/poller.ts
kevineger: Done.

src/colab/consumption/poller.ts
kevineger: Done.

src/colab/consumption/poller.ts
kevineger: Added paren to hopefully make this clearer.

PR #702, opened 2026-09-11
feat: concurrent files upload to colab

luqmansen:
Hi Colab Team,

I happen to use this feature quite a lot and notice that uploading an entire folder took some amount of time. I made a simple change to allow concurrent upload using simple `Promise.all` to allow interleaving upload without waiting prior file upload to be completed.

Let me know if I missed something from this PR. This is my first contribution to the repo. 

Thanks,

### DIFF
--- a/src/colab/commands/files.ts
+++ b/src/colab/commands/files.ts
@@ -70,23 +70,45 @@ export async function upload(
           cancellable: false,
         },
         async (progress) => {
-          for (const op of operations) {
+
+          const dirOps = operations.filter((op) => op.type === 'directory');
+          const fileOps = operations.filter((op) => op.type === 'file');
+
+          for (const op of dirOps) {
             try {
-              if (op.type === 'directory') {
-                await createDirectory(vs, op.dest);
-              } else {
-                const fileName = op.source.path.split('/').pop() ?? '';
-                progress.report({
-                  message: `Importing ${fileName}...`,
-                  increment: incrementPerFile,
-                });
-                const content = await vs.workspace.fs.readFile(op.source);
-                await vs.workspace.fs.writeFile(op.dest, content);
-                successCount++;
-                uploadedBytes += content.byteLength;
-              }
+              await createDirectory(vs, op.dest);
             } catch (err) {
-              log.error(`Failed to process ${op.dest.toString()}`, err);
+              log.error(`Failed to create directory ${op.dest.toString()}`, err);
+              failCount++;
+            }
+          }
+
+          const uploadResults = await Promise.all(
+            fileOps.map(
+              async (op) => {
+                try {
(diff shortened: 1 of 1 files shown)


luqmansen:
Signed the CLA 🙏

src/colab/commands/files.ts
```diff
@@ -70,23 +70,45 @@ export async function upload(
...
+                  });
+                  const content = await vs.workspace.fs.readFile(op.source);
+                  await vs.workspace.fs.writeFile(op.dest, content);
+                  return { success: true, bytes: content.byteLength };
+
+                } catch (err) {
+                  log.error(`Failed to process ${op.dest.toString()}`, err);
+                  return { success: false, bytes: 0 };
+                }
+              }
```
hjjackyang: Consider extracting this into an `uploadFile` helper function to reduce nesting.

hjjackyang (changes requested):
I see the Pre-Merge check failed. Please run `npm run format` to fix the format.

src/colab/commands/files.ts
luqmansen: Thanks for the review!

updated in [ee2d71f](https://github.com/googlecolab/colab-vscode/pull/702/commits/ee2d71f3915b565299d1f974cdd59b7dfb65d8fc)

src/colab/commands/files.ts
hjjackyang, line 1: Would you mind adding unit test coverage to verify the new parallel behavior?

src/colab/commands/files.ts
```diff
@@ -70,23 +70,38 @@ export async function upload(
...
-              log.error(`Failed to process ${op.dest.toString()}`, err);
+              log.error(
+                `Failed to create directory ${op.dest.toString()}`,
+                err,
+              );
+              failCount++;
+            }
+          }
+
+          const uploadResults = await Promise.all(
```
hjjackyang, line 88: I think there may be a potential issue that uploading thousands of files all in parallel might make VS Code hang (@kevineger to correct me if I'm wrong).

I wonder if it makes sense to divide files into reasonable-sized (e.g. 50) batches, and upload files within a batch in parallel while sequencing the batches?

@kevineger - WDYT?

hjjackyang (changes requested):
Thanks for addressing my previous comments! Added a couple more which I missed previously.

src/colab/commands/files.ts
kevineger: Yeah, a simple `Promise.all` will issue the all concurrently (as I understand), which'll likely cause issues (as I read the code). Have you tried running this locally? There's a chance VS Code meters the `workspace.fs.writeFile` but I'd be surprised if that was the case.

PR #703, opened 2026-09-14
fix: stop reporting abandoned tasks with a zero grace period

kevineger:
`NonGracefulAbandonError` is the noisiest error event of late. This came to light after  #688, which started sending the error to telemetry. Until #688 it never reached telemetry, which is why the events begin in 0.9.4.

Two things were wrong.

- A zero grace period cannot be met. Every runner configures `abandonGraceMs: 0`, meaning "nothing to clean up, abandon immediately", yet the abort handler still armed a 0ms timer against the task settling. Measured against the runner: with a zero grace, a task that settles 50ms after its abort is reported as a defect; the same task under a 100ms grace is not. Only a task settling inside the abort's own microtask drain escapes, and the polling tasks cannot rely on that.
- The report was not gated on why the abort happened, so aborting a task via `runNow()`, `stop()` or `dispose()` reported it as a defect too.


src/common/task-runner.ts
```diff
@@ -274,6 +278,38 @@ export class SequentialTaskRunner implements Disposable {
...
+      telemetry.logError(err);
+      return;
+    }
+    // A zero grace means "nothing to clean up, abandon immediately". The task
+    // was never given a window it could have met, so overrunning it says
+    // nothing. Report why we aborted instead. The timeout itself is already
+    // logged by the abort listener in `runWorker`.
+    if (abortReason instanceof TimeoutError) {
+      telemetry.logError(abortReason);
+    }
```
hjjackyang, line 310: Nit: The comment below is a bit confusing to me:

> The timeout itself is already logged by the abort listener in `runWorker`.

Making me think that you only intend to log non-`TimeoutError` here because `TimeoutError` is already logged elsewhere. Suggest to align the comment and implementation to make it more clear.

Also, I wonder if the "logged by the abort listener in `runWorker`" in the comment means the following code:

https://github.com/googlecolab/colab-vscode/blob/36666819c3cc0369af34c9ea9843a3ac40a6a7a6/src/common/task-runner.ts#L207-L216

If so, perhaps we should just add `telemetry.logError(reason);` in the same abort listener, and remove it from elsewhere. WDYT?

hjjackyang (approved):

src/common/task-runner.ts
kevineger: Refactored the code to hopefully make this all more clear.

hjjackyang: LGTM!

PR #704, opened 2026-09-14
fix: drop stored servers this version cannot read

kevineger:
`ServerStorage` parsed the whole persisted array with a single `AssignedServers.parse`, so one entry the current schema cannot read threw out of `list()` and out of `store()` and `remove()` via `mapServersById`. The entry stays in secret storage, so every later call fails the same way and there is no way to delete the offending server from inside the product.

Parse entries individually and prune the ones that fail, so an unreadable entry costs that server rather than the whole list, and is gone rather than re-read and re-reported on every call. An assigned server is ephemeral (idle ones reclaimed and none outliving 24 hours) so an entry written by a schema this version predates refers to a runtime that is long expired, and recovering it would only resurrect a dead endpoint.

An unrecognized `shape` still degrades to undefined rather than dropping the server, since it is presentational and a live server should survive the API gaining a shape this version predates.


src/jupyter/storage.ts
```diff
@@ -7,30 +7,72 @@
...
+
+const AssignedServer = z.object({
+  id: z.string(),
+  label: z.string().nonempty(),
+  variant: z.enum(Variant),
+  accelerator: z.string().optional(),
+  // The shape is presentational. Degrading an unrecognized one keeps a live
+  // server usable if the API gains a shape this version predates, rather than
+  // discarding it over a label.
+  shape: z.enum(Shape).optional().catch(undefined),
```
hjjackyang, line 24: Just curious, did my public API change somehow break this?

hjjackyang (approved):
Great improvement!

src/jupyter/storage.ts
kevineger: I came across the issue while looking through telemetry errors. Found a handful where somehow `0` was persisted instead of the variant string name. Before we shipped (while we were doing internal testing), we had that. I suspect the errors are from Googlers that had a pre-production version of the extension and forever have a corrupt server stored.

So unless we have some other path where `0` is somehow making it to disk (which I can't identify), no.

PR #709, opened 2026-09-17
refactor: `refreshConnection` to throw `NotFoundError` on 404

hjjackyang:
**Why**? Noticed many 404 `NOT_FOUND` errors from `GetRuntime` calls on server:

image

I suspect the cause is due to the servers not garbage collected in `ConnectionRefresher` properly after they are gone. This fix may not fully address the issue, but is low risk and nice to have.

---

While I was here, also added a missing test case to cover `GetRuntime` returning a runtime without `ConnectionInfo`.

---

Internal tracking bug: b/562976948


src/jupyter/assignments.ts
```diff
@@ -468,15 +469,25 @@ export class AssignmentManager implements Disposable {
...
+    let runtime: Runtime;
+    try {
+      runtime = await this.colabApiClient.colab.getRuntime(
+        { runtime: id },
+        { signal },
+      );
+    } catch (e: unknown) {
+      // Rethrow 404 as NotFoundError so the server is garbage collected in
+      // ConnectionRefresher right away to avoid another refresh.
+      if (e instanceof ResponseError && e.response.status === 404) {
```
kevineger, line 484: Consider a shared type guard function with the other 404 response check we do [here](https://github.com/googlecolab/colab-vscode/blob/83cfd1302aecca966440d6bf5f35b07ef8551621/src/jupyter/contents/file-system.ts#L481).

src/jupyter/assignments.unit.test.ts
```diff
@@ -2274,12 +2274,51 @@ describe('AssignmentManager', () => {
...
+      getRuntimeStub
+        .withArgs(sinon.match({ runtime: defaultServer.id }), sinon.match.any)
+        .rejects(new ResponseError(new Response(null, { status: 500 })));
+
+      await expect(
+        assignmentManager.refreshConnection(defaultServer.id),
+      ).to.eventually.be.rejectedWith(ResponseError);
+    });
+
+    it('throws an error if refreshed runtime is missing connection info', async () => {
```
kevineger, line 2307: Thanks for adding missing coverage

kevineger (approved):

src/jupyter/assignments.ts
hjjackyang: As discussed offline, refactoring this isn't as straightforward because the underlying `ResponseError` are 2 different classes (semantically the same, but generated in 2 different files).

I will leave this as is until there is a 3rd occurrence of the duplication - https://wiki.c2.com/?ThreeStrikesAndYouRefactor.

PR #710, opened 2026-09-17
refactor: clean up `EnableOpCredentialPropagationApi` feature flag and usage

hjjackyang:
With `enable_op_credprop_api_vscode` fully launched for almost a week (cl/981189018) with no major issues, we are ready to clean up the feature flag and the TFE `/credentials-propagation` call from the codebase.

✅ Sanity tested locally with no problem.

---

Internal tracking bug: b/556251396


src/colab/client/v1/index.ts
```diff
@@ -17,32 +17,28 @@ import {
...
   COLAB_VS_CODE_EXTENSION_VERSION,
-  COLAB_XSRF_TOKEN_HEADER,
   CONTENT_TYPE_JSON_HEADER,
 } from '../../headers';
 import {
+  AccessTokenType,
   AuthType,
   ConsumptionUserInfo,
   ConsumptionUserInfoSchema,
   CredentialsPropagationResult,
```
kevineger, line 33: Should the schema go too, and `CredentialsPropagationResult` become a plain type?

Nothing parses [`CredentialsPropagationResultSchema`](https://github.com/googlecolab/colab-vscode/blob/effec6b4cd0b0a4c170838c0c66d3d01e1f194b4/src/colab/client/v1/api.ts#L245-L260) any more once `propagateCredentialsV1` is gone, and its `unauthorized_redirect_uri` transform is the V1 response shape. The OP path builds the result by hand and reads the redirect URI out of [`ErrorInfo.metadata`](https://github.com/googlecolab/colab-vscode/blob/effec6b4cd0b0a4c170838c0c66d3d01e1f194b4/src/colab/client/v1/index.ts#L348) instead.

project-words.txt
```diff
@@ -5,7 +5,6 @@ chromedriver
 colab
 colabclient
 colaboratory
-credprop
 dryrun
```
kevineger: Mind doing a quick audit? I think `dryrun` and `authtype` might no longer be used?

kevineger: Curious actually if the eslint warns in this case.

kevineger (approved):

src/colab/client/v1/index.ts
hjjackyang: Good suggestion!

Got rid of `CredentialsPropagationResultSchema` and changed `CredentialsPropagationResult` to an interface instead. Another upside of doing this is that I can now define `unauthorizedRedirectUri` as `?: string`, so I don't have to explicitly set it to `undefined` when constructing a success result!

project-words.txt
hjjackyang: Good catch! Removed both `dryrun` and `authtype`.

> Curious actually if the eslint warns in this case.

If you meant if eslint warns if project words become unused, the answer is no.

PR #711, opened 2026-09-18
refactor: `AssignmentManager` to key servers off runtime ID instead of endpoint

hjjackyang:
* Mainly updated `reconcileStoredServers` and `getUnownedServers` to use server `id` instead of `endpoint` as unique key to reconcile and filter servers.
* **Why**? Post https://github.com/googlecolab/colab-vscode/pull/681, stored Colab server ID will fully match runtime ID returned from the public API. This refactoring will reduce dependency on the internal `endpoint` field in the open source world, and treat `id` as the unique key for Colab server wherever possible.
* ✅ Sanity tested locally with no obvious issues.


src/jupyter/assignments.ts
```diff
@@ -589,13 +583,15 @@ export class AssignmentManager implements Disposable {
 
   private async reconcileStoredServers(
     storedServers: ColabAssignedServer[],
-    liveEndpoints: string[],
+    liveRuntimes: AssertedRuntime[],
   ): Promise<ColabAssignedServer[]> {
-    const liveEndpointSet = new Set(liveEndpoints);
+    const liveServerIdSet = new Set(
```
kevineger: Nit: (existing...) `...Set` suffix unnecessary.

src/jupyter/assignments.ts
```diff
@@ -589,13 +583,15 @@ export class AssignmentManager implements Disposable {
...
+      liveRuntimes.map((r) => trimPrefix(r.name, RUNTIME_NAME_PREFIX)),
+    );
     const removed: ColabAssignedServer[] = [];
     const reconciled: ColabAssignedServer[] = [];
     for (const s of storedServers) {
-      if (liveEndpointSet.has(s.endpoint)) {
+      if (liveServerIdSet.has(s.id)) {
```
kevineger: Should `getUnownedServers` key off id too?

It still [filters by `storedEndpointSet`](https://github.com/googlecolab/colab-vscode/blob/78ac155e81412ea0eeadc15f40729d646c363aa8/src/jupyter/assignments.ts#L721-L726), and `getServers('all')` derives both lists from the same `allAssignedRuntimes`, so a stored server whose endpoint drifted would reconcile as assigned and come back as unowned. [`refreshConnection` re-pins the stored endpoint](https://github.com/googlecolab/colab-vscode/blob/78ac155e81412ea0eeadc15f40729d646c363aa8/src/jupyter/assignments.ts#L474-L479) instead of taking the fresh one.

Also, `logPruneServers` below still reports endpoint, worth switching to id now that's the match key.

src/jupyter/assignments.unit.test.ts
```diff
@@ -997,10 +1002,7 @@ describe('AssignmentManager', () => {
           ],
         });
         // One of the assignments was assigned within VS Code extension
-        const assignedServer = {
-          ...defaultServer,
-          endpoint: runtimeWithoutSessionName.connectionInfo.endpoint,
-        };
+        const assignedServer = { ...defaultServer, ...serverWithoutName };
```
kevineger, line 999: Should one of these cases have a stored server whose id matches a live runtime but whose endpoint doesn't?

Every fixture here spreads id and endpoint from the same `serverWith*`, so `reconcileStoredServers` matching on id and `getUnownedServers` matching on endpoint can never disagree. The [`olderActiveServer` / `olderActiveRuntime` pair](https://github.com/googlecolab/colab-vscode/blob/78ac155e81412ea0eeadc15f40729d646c363aa8/src/jupyter/assignments.unit.test.ts#L2186-L2199) is already that shape, it just never reaches `getServers('all')`.

src/jupyter/assignments.ts

src/jupyter/assignments.ts
hjjackyang: Great catch!

Fixed:
1. `getUnownedServers` to rely on `id` instead of `endpoint`.
2. `refreshConnection` to take in the fresh `endpoint`.
3. `logPruneServers` to report on `id` instead of `endpoint` - This also requires a separate proto comment update to align the behavior.

src/jupyter/assignments.unit.test.ts
hjjackyang: I updated the test data for these tests to *not* supply different endpoints for different runtimes (retaining the same default `endpoint` value from `defaultServer`).

This way it would hopefully catch future code changes when logic across methods don't agree.

src/jupyter/assignments.ts
```diff
@@ -490,7 +484,7 @@ export class AssignmentManager implements Disposable {
     assert(runtime.connectionInfo, `${MISSING_CONNECTION_INFO_ERR_MSG}: ${id}`);
     const updatedServer = this.toAssignedServer(
       server,
-      server.endpoint,
+      runtime.connectionInfo.endpoint,
```
kevineger: Should `RefreshingClient` match on `id` too, now that a refresh can change `endpoint`?

[It filters `changed` by the endpoint it captured at construction](https://github.com/googlecolab/colab-vscode/blob/98d9d629f22b713b6ee6acfc64cbc008624fa4bd/src/jupyter/client/index.ts#L157-L169), so a drifted endpoint would leave it on the old token. [`ConnectionRefresher` already keys `changed` by id](https://github.com/googlecolab/colab-vscode/blob/98d9d629f22b713b6ee6acfc64cbc008624fa4bd/src/colab/connection-refresher.ts#L224-L225).

src/jupyter/assignments.ts
```diff
@@ -490,7 +484,7 @@ export class AssignmentManager implements Disposable {
     assert(runtime.connectionInfo, `${MISSING_CONNECTION_INFO_ERR_MSG}: ${id}`);
     const updatedServer = this.toAssignedServer(
```
kevineger: Both callers now pass `connectionInfo.endpoint`. Should `toAssignedServer` read it off `connectionInfo` and drop the `endpoint` param?

src/jupyter/assignments.ts
```diff
@@ -733,20 +729,21 @@ export class AssignmentManager implements Disposable {
     storedServers: ColabAssignedServer[],
     signal?: AbortSignal,
   ): Promise<UnownedServer[]> {
-    const storedEndpointSet = new Set(storedServers.map((s) => s.endpoint));
+    const storedRuntimeNames = new Set(
+      storedServers.map((s) => RUNTIME_NAME_PREFIX + s.id),
+    );
```
kevineger: Should this compare trimmed ids like `reconcileStoredServers` does?

The two have to agree, and `trimPrefix` tolerates a name without the prefix where this concatenation doesn't.

src/jupyter/assignments.unit.test.ts
```diff
@@ -727,7 +735,7 @@ describe('AssignmentManager', () => {
           listRuntimesStub.resolves({ runtimes: [defaultRuntime] });
           const noLongerAssignedServer = {
             ...defaultServer,
-            endpoint: 'no-longer-assigned',
+            id: 'no-longer-assigned',
```
kevineger: Should we also assert `logPruneServers` gets `['no-longer-assigned']` here?

Nothing pins the switch from endpoint to id, and the [`logAssignServer` tests already stub telemetry](https://github.com/googlecolab/colab-vscode/blob/98d9d629f22b713b6ee6acfc64cbc008624fa4bd/src/jupyter/assignments.unit.test.ts#L1639-L1643).

PR #712, opened 2026-09-18
fix: disambiguate a cancelled sign-in apart from a failed one.

kevineger:
Sign-in failures were reported with nothing to diagnose them by. `login` caught each flow's error, showed it in a toast, and then threw a bare `All authentication methods failed.` that referenced none of them, so every report logged without an indication of what, if anything, broke. This extends to users who simply hit `Cancel`. Those reject on cancellation just like they do on failure, so a deliberate "no thanks" looks the same.

This change effectively disambiguates the types of cancellation. `login` now throws an `AggregateError` carrying every attempt's error when anything genuinely failed, and a cancellation only when nothing did. So cancelling out of both flows is quiet while a failure followed by a cancellation is not.

`resolveCode` threw when no exchange awaited a nonce, which happens routinely: a browser replaying the redirect, or a code pasted after the 60s exchange timeout. From the loopback handler that throw escaped the request listener as an uncaught exception and left the tab waiting on a response that never came; from the paste prompt it became an unhandled rejection and the sign-in died silently. It now reports the outcome and both callers answer for themselves.

Finally, the code promise in each flow is only awaited on the happy path, so a failure during setup orphaned it to reject unhandled a minute later as an exchange timeout. Its rejection is claimed up front.


src/auth/flows/loopback.ts
```diff
@@ -160,7 +164,17 @@ class Handler implements LoopbackHandler {
         if (!nonce || !code) {
           throw new Error('Missing nonce or code in redirect URI');
```
kevineger: Should we answer the deny path here too?

Google redirects a denied consent to `/?error=access_denied&state=nonce=...`, so `state` is present, `code` is absent, and this throws. [`handleRequest` is called bare](https://github.com/googlecolab/colab-vscode/blob/99a2ccd0ecadbca055988e9868d43a1ecfe4bb7c/src/common/loopback-server.ts#L62-L64), so it escapes the listener exactly as the 409 case eleven lines below would have.

src/auth/flows/loopback.ts
```diff
@@ -93,6 +93,10 @@ export class LocalServerFlow implements OAuth2Flow, vscode.Disposable {
     this.activeServers.add(server);
     try {
       const code = this.codeManager.waitForCode(options.nonce, options.cancel);
+      // `code` is only awaited on the happy path below. Claim its rejection now
+      // so bailing out during setup doesn't leave it to reject unhandled once
+      // the exchange later times out.
+      void code.catch(() => undefined);
```
kevineger: Should we cover the unclaimed rejection, e.g. reject `server.start()` and assert no `unhandledRejection` fires?

Deleting this line, or its twin in `proxied.ts`, leaves the whole suite green. [`colab-proxy-websocket.unit.test.ts`](https://github.com/googlecolab/colab-vscode/blob/99a2ccd0ecadbca055988e9868d43a1ecfe4bb7c/src/jupyter/colab-proxy-websocket.unit.test.ts#L279-L283) already has the `process.on('unhandledRejection')` idiom.

src/auth/login.ts
```diff
@@ -70,12 +72,13 @@ export async function login(
     throw new Error('No authentication flows available.');
   }
 
+  const attemptErrors: Error[] = [];
   for (const flow of flows) {
+    if (flow !== flows[0] && !(await promptIfFallback(vs))) {
```
kevineger: Should the fallback prompt stop claiming a failure when the previous attempt was cancelled?

[Line 115](https://github.com/googlecolab/colab-vscode/blob/99a2ccd0ecadbca055988e9868d43a1ecfe4bb7c/src/auth/login.ts#L115-L121) suppresses the failure toast for a cancellation, then this reaches `promptIfFallback`, which opens with `Failed to authenticate with Google.`. The new `reports cancelling out of every flow as a cancellation` test asserts that toast fires.

src/auth/login.ts
```diff
@@ -106,21 +109,50 @@ export async function login(
...
+      cause: attemptErrors[0],
+    });
+  }
   const msg =
-    flows.length > 1
+    attemptErrors.length > 1
       ? 'All authentication methods failed.'
       : 'Authentication failed.';
-  throw new Error(msg);
+  return new AggregateError(attemptErrors, msg);
```
kevineger: Should we update the comment in `error-description.ts`?

[`nextLinks` says "An AggregateError's own message is empty"](https://github.com/googlecolab/colab-vscode/blob/99a2ccd0ecadbca055988e9868d43a1ecfe4bb7c/src/common/error-description.ts#L40), which this makes false. The telemetry output is still correct.

src/auth/auth-provider.ts
```diff
@@ -326,8 +327,11 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
       this.vs.window.showInformationMessage('Signed in to Google!');
       return this.session;
     } catch (err: unknown) {
-      const msg = err instanceof Error ? err.message : 'unknown error';
-      this.vs.window.showErrorMessage(`Sign in failed: ${msg}`);
+      // A user who abandoned the sign-in does not need to be told it failed.
+      if (!isCancellation(err)) {
+        const msg = err instanceof Error ? err.message : 'unknown error';
+        this.vs.window.showErrorMessage(`Sign in failed: ${msg}`);
```
kevineger: Should this surface the aggregated reasons rather than `.message`?

`err.message` is `Authentication failed.` and drops `.errors`, so the toast reads `Sign in failed: Authentication failed.` right after [each attempt already toasted its own reason](https://github.com/googlecolab/colab-vscode/blob/99a2ccd0ecadbca055988e9868d43a1ecfe4bb7c/src/auth/login.ts#L118-L120).

src/auth/flows/proxied.ts
```diff
@@ -122,7 +126,14 @@ export class ProxiedRedirectFlow implements OAuth2Flow, vscode.Disposable {
           },
           value: '',
         });
-        this.codeManager.resolveCode(nonce, pastedCode);
+        if (!this.codeManager.resolveCode(nonce, pastedCode)) {
```
kevineger: Should the disposed case be covered here as well?

`resolveCode` still throws from [`guardDisposed`](https://github.com/googlecolab/colab-vscode/blob/99a2ccd0ecadbca055988e9868d43a1ecfe4bb7c/src/auth/code-manager.ts#L93), and `dispose()` does not close the open input box, so a paste after dispose escapes the `void MultiStepInput.run(...)` above as an unhandled rejection.

src/auth/flows/proxied.ts
```diff
@@ -122,7 +126,14 @@ export class ProxiedRedirectFlow implements OAuth2Flow, vscode.Disposable {
...
+          // The input box outlives the exchange, so a code pasted after the
+          // exchange timed out or was cancelled arrives with nobody waiting.
+          // Say so instead of failing silently.
+          this.vs.window.showWarningMessage(
+            'That authorization code arrived too late. Please sign-in again.',
```
kevineger: `sign in` is the verb here. The repo uses `Sign in to Google` for the verb and `Sign-in attempt failed` for the noun.

src/auth/login.unit.test.ts
```diff
@@ -245,5 +309,71 @@ describe('login', () => {
...
+    it('still reports a failure when the user declines another method', async () => {
+      flow1.trigger.rejects(new Error('Barf'));
+      // `showErrorMessage` resolves undefined by default, e.g. "No".
+
+      const err = await rejectionOf(
+        login(vs.asVsCode(), [flow1, flow2], oauth2Client, SCOPES),
+      );
+
+      expect(isCancellation(err)).to.be.false;
+      expect(aggregatedMessages(err)).to.deep.equal(['Barf']);
```
kevineger: Should this also assert the thrown message is `Authentication failed.`?

This is the only test where `attemptErrors.length` and `flows.length` differ, and reverting [the condition](https://github.com/googlecolab/colab-vscode/blob/99a2ccd0ecadbca055988e9868d43a1ecfe4bb7c/src/auth/login.ts#L147-L150) to `flows.length > 1` leaves the whole suite green.

src/jupyter/colab-proxy-websocket.unit.test.ts
```diff
@@ -232,6 +233,38 @@ describe('colabProxyWebSocket', () => {
...
+        sendSpy,
+        sinon.match((data: string) => {
+          const message = JSON.parse(data) as unknown;
+          return (
+            isColabInputReplyMessage(message) &&
+            message.content.value.error === errMsg
+          );
+        }),
+      );
+      sinon.assert.notCalled(logErrorStub);
```
kevineger: Should the failure test above also assert `logError` was called?

Deleting [`telemetry.logError(err)`](https://github.com/googlecolab/colab-vscode/blob/99a2ccd0ecadbca055988e9868d43a1ecfe4bb7c/src/jupyter/colab-proxy-websocket.ts#L162) leaves the whole suite green, so this assertion passes for the wrong reason.

src/auth/flows/proxied.unit.test.ts
```diff
@@ -134,4 +135,28 @@ describe('ProxiedRedirectFlow', () => {
...
+    await expect(trigger).to.eventually.be.rejectedWith(/timeout/);
+
+    inputBoxStub.value = CODE;
+    inputBoxStub.onDidChangeValue.yield(CODE);
+    inputBoxStub.onDidAccept.yield();
+
+    await warned.promise;
+    sinon.assert.calledOnceWithMatch(
+      vs.window.showWarningMessage,
+      sinon.match(/too late/),
```
kevineger: Should the happy-path test also assert `notCalled(showWarningMessage)`?

Forcing [the new condition](https://github.com/googlecolab/colab-vscode/blob/99a2ccd0ecadbca055988e9868d43a1ecfe4bb7c/src/auth/flows/proxied.ts#L129) to always warn leaves the whole suite green.

src/auth/flows/loopback.ts
```diff
@@ -156,11 +161,38 @@ class Handler implements LoopbackHandler {
         }
         const parsedState = new URLSearchParams(state);
         const nonce = parsedState.get('nonce');
+        const error = url.searchParams.get('error');
+        if (error) {
```
kevineger: Should we gate this on `error === 'access_denied'`?

[RFC 6749 4.1.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) defines seven codes and only that one is a refusal, so `server_error` or `invalid_scope` becomes a `UserCancelledError` that [`createSession` deliberately never toasts](https://github.com/googlecolab/colab-vscode/blob/1be3b59fc299c239a1d4f7c67ad406eed4ec7959/src/auth/auth-provider.ts#L331) and [`process-errors` never reports](https://github.com/googlecolab/colab-vscode/blob/1be3b59fc299c239a1d4f7c67ad406eed4ec7959/src/telemetry/process-errors.ts#L26). Narrowing the branch to `access_denied` leaves the suite green, so only the refusal case is covered.

Also please carry `error` into the thrown error. The log line is the only place it survives today.

src/auth/flows/loopback.ts
```diff
@@ -156,11 +161,38 @@ class Handler implements LoopbackHandler {
...
+          // A refused consent comes back as `?error=access_denied&state=...`:
+          // `state` is present but no code is coming. Fail the exchange now so
+          // the user is not left watching a picker until it times out, and
+          // answer the tab rather than throwing past the request listener.
+          log.info(`Sign-in was refused at the consent screen: ${error}`);
+          if (nonce) {
```
kevineger: Should a missing nonce still answer the waiter?

The comment four lines above says this branch exists so the user is not left watching the picker, but when `state` parses no nonce nothing is rejected and they wait the full [60s](https://github.com/googlecolab/colab-vscode/blob/1be3b59fc299c239a1d4f7c67ad406eed4ec7959/src/auth/code-manager.ts#L11). Injecting a throw before the guard leaves the suite green.

src/auth/flows/loopback.ts
```diff
@@ -156,11 +161,38 @@ class Handler implements LoopbackHandler {
...
+              new UserCancelledError('Sign-in was refused.'),
+            );
+          }
+          res.writeHead(200, { 'Content-Type': 'text/plain' });
+          res.end('Sign-in was cancelled. You can close this tab.');
+          break;
+        }
         const code = url.searchParams.get('code');
         if (!nonce || !code) {
           throw new Error('Missing nonce or code in redirect URI');
```
kevineger: Should `handleRequest` be wrapped so a throw answers the socket instead?

[The listener calls it bare](https://github.com/googlecolab/colab-vscode/blob/1be3b59fc299c239a1d4f7c67ad406eed4ec7959/src/common/loopback-server.ts#L62-L64), so this, the `state` throw above it, and both `assert`s still escape uncaught and leave the tab hanging. A plain `GET /` with no query reaches one of them, and [the table at line 130](https://github.com/googlecolab/colab-vscode/blob/1be3b59fc299c239a1d4f7c67ad406eed4ec7959/src/auth/flows/loopback.unit.test.ts#L130-L134) pins that as expected.

src/auth/auth-provider.ts
```diff
@@ -489,6 +494,28 @@ export class GoogleAuthProvider implements AuthenticationProvider, Disposable {
...
+ * Renders a sign-in failure for a toast.
+ *
+ * `login` throws an `AggregateError` whose own message only names the umbrella
+ * failure, so reporting it verbatim says nothing the `Sign in failed` prefix
+ * has not already said. The reasons are in `errors`.
+ *
+ * @param err - The thrown value.
+ * @returns A message naming why sign-in failed.
+ */
+function getErrorDescription(err: unknown): string {
```
kevineger: Should this reuse `describeCauses`?

[`nextLinks` already flattens `AggregateError.errors`](https://github.com/googlecolab/colab-vscode/blob/1be3b59fc299c239a1d4f7c67ad406eed4ec7959/src/common/error-description.ts#L42-L46), and `describeCauses` also walks `cause` four levels deep, which this ignores. Telemetry renders through `describeCauses`, so a flow error wrapping its reason in `cause` shows the full chain there and a bare wrapper in the toast.

src/auth/login.ts
```diff
@@ -70,12 +76,14 @@ export async function login(
     throw new Error('No authentication flows available.');
   }
 
+  const attemptErrors: Error[] = [];
   for (const flow of flows) {
+    const previous = attemptErrors.at(-1);
```
kevineger: Should we add a third flow so `at(-1)` is pinned?

Changing this to `attemptErrors[0]` leaves the suite green. Every multi-flow test uses exactly two flows, where the two agree.

src/auth/flows/loopback.unit.test.ts
```diff
@@ -192,6 +193,66 @@ describe('LocalServerFlow', () => {
...
+      // The abandoned exchange rejects on its own timeout, long after `trigger`
+      // gave up on it.
+      clock.tick(60_001);
+      await new Promise((resolve) => setImmediate(resolve));
+    } finally {
+      process.off('unhandledRejection', capture);
+      clock.restore();
+    }
+
+    expect(unhandled).to.be.empty;
```
kevineger: Should the capture assert the rejection came from this exchange?

`to.be.empty` ignores origin, so a leak from another file fails this test, and one run in roughly ten did exactly that. Deleting the production line together with line 247 also leaves the suite green, so detection rests on that one `setImmediate`.

PR #713, opened 2026-09-18
fix: report a real Error for non-ok v2 API responses

kevineger:
`ErrorMiddleware.post` passed the undici `Response` object straight to `logError`. It has no enumerable own properties, so `JSON.stringify` renders it `{}` and every non-ok v2 response arrived as `Error: {}` carrying neither the status nor the path.

It also reported and then let the request continue, so the generated `request()` threw its own `ResponseError` a moment later at an error boundary. That means we had two events for one failure, the second saying only "Response returned an error code".

The middleware now throws instead of reporting, with the status and the path in the message. That is one event, raised wherever the call stack is.


src/colab/client/v2/index.ts
```diff
@@ -271,6 +279,18 @@ class ErrorMiddleware implements Middleware {
...
+ * Drops the query string, which can carry notebook hashes and auth hints.
+ *
+ * @param url - The URL to redact.
+ * @returns The URL without its query string.
+ */
+function redactUrl(url: string): string {
+  // Deliberately textual: parsing and falling back on failure leaks the query
+  // of any URL that does not parse, which is the opposite of what we want.
+  return url.split(/[?#]/)[0];
+}
```
kevineger: Should we export and reuse [`redactUrl` from `colab/errors.ts`](https://github.com/googlecolab/colab-vscode/blob/ded05bab1206a01a5847bfc7ef03c3eb6f2de368/src/colab/errors.ts#L30-L42) instead?

Its `catch` branch is this exact `url.split(/[?#]/)[0]`, and [there is a test for it](https://github.com/googlecolab/colab-vscode/blob/ded05bab1206a01a5847bfc7ef03c3eb6f2de368/src/colab/errors.unit.test.ts#L50-L55), so the premise in the comment does not hold. It also redacts values rather than dropping keys, so `requestId=REDACTED` still tells you which params were sent.

src/colab/client/v2/index.ts
```diff
@@ -249,16 +250,23 @@ class ErrorMiddleware implements Middleware {
...
+    if (context.response.status === 401 && this.onAuthError) {
+      await this.onAuthError();
     }
+    log.warn(
+      `Error response received by ${context.init.method ?? ''} ${context.url}:`,
+      context.response,
+    );
+    // Replaces the identical throw in the generated `request()`, whose
+    // ResponseError carries neither the status nor the path.
+    throw new ResponseError(
```
kevineger: Are we ok that operations calls now reject with the `colab` `ResponseError`?

This middleware instance is shared by both APIs, and `generated/operations` declares [its own `ResponseError`](https://github.com/googlecolab/colab-vscode/blob/ded05bab1206a01a5847bfc7ef03c3eb6f2de368/src/colab/client/v2/generated/operations/runtime.ts#L260) distinct from [the one imported here](https://github.com/googlecolab/colab-vscode/blob/ded05bab1206a01a5847bfc7ef03c3eb6f2de368/src/colab/client/v2/generated/colab/runtime.ts#L260). Nothing `instanceof`-checks the operations one today, but a future 404 branch around `waitOperation` would silently never match.

src/colab/client/v2/index.ts
```diff
@@ -249,16 +250,23 @@ class ErrorMiddleware implements Middleware {
...
-      );
+    if (context.response.ok) {
+      return;
+    }
+
+    if (context.response.status === 401 && this.onAuthError) {
+      await this.onAuthError();
     }
+    log.warn(
+      `Error response received by ${context.init.method ?? ''} ${context.url}:`,
```
kevineger: Should this use `redactUrl` too, or are logs deliberately kept raw?

src/colab/client/v2/index.unit.test.ts
```diff
@@ -170,10 +176,25 @@ describe('ColabApiClient', () => {
...
+            () => expect.fail('expected the request to reject'),
+            (e: unknown) => e,
+          );
+
+          if (!(error instanceof ResponseError)) {
+            expect.fail(`expected a ResponseError, got ${typeof error}`);
+          }
+          expect(error.message)
+            .to.contain(String(status))
+            .and.to.contain('/v1beta/subscription');
```
kevineger: Should one of these assert the query string is dropped?

`/v1beta/subscription` has no query, so swapping `redactUrl(context.url)` back to `context.url` leaves every test green. [`createRuntime` puts `requestId` and `runtimeId` in the query](https://github.com/googlecolab/colab-vscode/blob/ded05bab1206a01a5847bfc7ef03c3eb6f2de368/src/colab/client/v2/generated/colab/apis/ColaboratoryApi.ts#L420-L427) and already has the same status block.

src/colab/client/v2/index.ts
```diff
@@ -271,6 +279,18 @@ class ErrorMiddleware implements Middleware {
   }
 }
 
+/**
+ * Drops the query string, which can carry notebook hashes and auth hints.
+ *
+ * @param url - The URL to redact.
+ * @returns The URL without its query string.
+ */
+function redactUrl(url: string): string {
```
hjjackyang: Consider exporting and reusing this function?

https://github.com/googlecolab/colab-vscode/blob/ddd48e63c10caa7dfe9fdd4fe0ed5ca320ef2a0b/src/colab/errors.ts#L30

src/colab/client/v2/index.ts
hjjackyang: Nit: Redact this URL too?

hjjackyang (commented):
Can't help but notice this PR before you send it ;)

src/colab/client/v2/index.ts
kevineger: Done — dropped the local copy and exported the one in `errors.ts`. Its `catch` branch was already doing the same thing the local one did, so there was nothing to reconcile. Added direct tests for it while it was becoming public.

src/colab/client/v2/index.ts
kevineger: Done, and the `log.warn` above it too. Worth noting it was not purely cosmetic: that `log.warn` was also passing the `Response` as a second argument, and `OutputChannelLogger` `JSON.stringify`s object args, so it was printing the same useless `{}` this PR is about. It now shares the thrown message instead.

src/colab/client/v2/index.ts
```diff
@@ -51,6 +53,10 @@ export interface ColabApiClient {
 
   /**
    * A client instance to access the Operations APIs.
+   *
+   * Non-ok responses reject with the {@link ResponseError } declared by
```
hjjackyang, line 57: Nit:

```suggestion
   * Non-ok responses reject with the {@link ResponseError} declared by
```

hjjackyang (approved):

PR #714, opened 2026-09-18
fix: stop reporting a declined consent as an extension error

kevineger:
`process-errors` already knew that a dismissed picker is normal control flow, but that was in a private helper. A user declining the ephemeral auth prompt threw a plain `Error` that `colabProxyWebSocket` logged and reported like a defect.

This lifts the classification into a standalone module, alongside a `UserCancelledError` for code that needs to raise one. `handleEphemeralAuth` throws that when the user declines and `CodeManager` when the token is cancelled, and the websocket logs those as _info_ and not to telemetry. The kernel still hears that no credentials are coming.

Propagation failures also carried no reason at all. The result has no message to report, but whether the backend handed back another authorization URL separates a consent that did not take effect from a failure with some other cause, so the thrown error now says which.


src/auth/ephemeral.ts
```diff
@@ -124,6 +126,14 @@ async function propagateCredentials(
   log.trace(`[${authType}] credentials propagation:`, propagationResult);
 
   if (!propagationResult.success) {
-    throw new Error(`[${authType}] Credentials propagation unsuccessful`);
+    // The result carries no failure message, but whether the backend handed
+    // back another authorization URL separates "the consent we just collected
+    // did not take effect" from "propagation failed for some other reason".
+    const reason = propagationResult.unauthorizedRedirectUri
+      ? 'the server still reports the credentials as unauthorized'
+      : 'the server reported no reason';
```
hjjackyang, line 134: With the new OnePlat API, this `the server reported no reason` path will never trigger.

On server side, errors are [directly returned](http://google3/research/colab/boq/oneplatform/runtime/service/credprop.go;l=206-208;rcl=929368756), and on the client side, they are [thrown](https://github.com/googlecolab/colab-vscode/blob/ddd48e63c10caa7dfe9fdd4fe0ed5ca320ef2a0b/src/colab/client/v1/index.ts#L209) rather than returned as `success: false`.

I think one way we can refactor this is to augment `CredentialsPropagationResult` to contain the error thrown as `reason`, so we can catch and report them here properly. Thoughts?

kevineger: Pushed [765d7c4](https://github.com/googlecolab/colab-vscode/pull/714/commits/765d7c42de640f5b41287f93f1d78c14db4a85db) simplifying the cred prop API.

hjjackyang: LGTM!

hjjackyang (approved):
LGTM!

PR #715, opened 2026-09-18
fix: answer a malformed redirect instead of throwing

kevineger:
`LoopbackServer` calls its handler straight out of Node's `request` listener, so anything the handler throws goes uncaught. The loopback flow throws on four shapes of malformed redirect, and a plain `GET /` with no query reaches one of them, so a stray request left the browser tab waiting on a response that never came.

The listener now catches, logs, and answers 500 (when it can).


src/common/loopback-server.ts
```diff
@@ -60,7 +60,20 @@ export class LoopbackServer implements vscode.Disposable {
...
+        // Node calls this listener bare, so a throw would go uncaught and
+        // leave the browser tab waiting
+        log.error('Loopback request handler failed', err);
+        if (res.writableEnded) {
+          return;
+        }
+        if (!res.headersSent) {
+          res.writeHead(500, { 'Content-Type': 'text/plain' });
+        }
+        res.end('Sign-in could not be completed. You can close this tab.');
```
hjjackyang, line 75: Nit: Is there a way to replicate this locally and attach a screencast to demo this fix in action?

hjjackyang (approved):

src/common/loopback-server.ts
kevineger: Yep. I'll add to PR desc.

